PR #3, opened 2017-07-26
Modify tooltips to trim length of pet/mount names and titles, as well…

Connor-Macleod:
… as owner's names on pet tooltips.


Ellypse:
👌

Issue #49, opened 2018-01-02
Add /trp3 command to request and open a player profile


Ellypse:
Add a slash command to request and open a specific player profile, like `/trp3 open Ellypse-ArgentDawn`
The profile would open once the profile has been completely received.

PR #56, opened 2018-01-02
Issue #49 : Added "/trp3 open" command to open profile

Solanya:
Usages :
- **/trp3 open** to open your target's profile.
- **/trp3 open CharacterName-RealmName** to open that character's page.
- **/trp3 open CharacterName** will append automatically the player's realm.

It will open instantly if the profile is in the directory before sending the query, else it will wait until the first tab of the profile is loaded before opening.

Might be worth adding a warning in chat when it has to wait for the exchange ?


Solanya:
Stupid edge case : 
- Requesting profile from someone offline, only answer is "No character named "..." currently exists".
- The player comes back online, hovering over him will trigger a request for update.
- Profile opens automatically because the addon still has the name in its variable.

I need to know if the requested character doesn't exist.

Ellypse:
Used C_Timer.After to empty the variable after 1 minute, to abort opening the profile and make sure we get back to an empty state at some point.

Solanya:
If you're using the command on a wrong target (no response) then run the command 59 seconds later on a target that needs downloading, the first timer might abort the opening of the second profile before the download is complete.

Ellypse:
Ooops. We should use `[C_Timer.NewTimer](https://wow.gamepedia.com/API_C_Timer.NewTimer)` instead then, so we can store a reference to the timer and cancel it when running a new command.

Solanya:
One fun issue remains : If capitalization is wrong on the player's name/realm, it tells the player it's requesting the profile, requests it, but doesn't actually open the profile since the string in the register won't match the command string.
Is there a way to get the proper capitalization ? (Or compare strings with no regards to capitalization ?)

totalRP3/modules/register/main/register_exchange.lua
```diff
@@ -49,13 +49,14 @@ local displayMessage = TRP3_API.utils.message.displayMessage;
 -- WoW imports
 local UnitName, UnitIsPlayer, UnitFactionGroup, CheckInteractDistance, UnitFullName = UnitName, UnitIsPlayer, UnitFactionGroup, CheckInteractDistance, UnitFullName;
 local tinsert, time, type, pairs, tonumber = tinsert, GetTime, type, pairs, tonumber;
-local after = C_Timer.After;
+local after = C_Timer.NewTimer;
```
Ellypse: It would be better to have our local reference be newTimer so that it is clear in that we are not actually using after here. We may need to use after in this file at some point.

totalRP3/modules/register/main/register_exchange.lua
```diff
@@ -476,6 +477,10 @@ TRP3_API.slash.registerCommand({
 		local args = {...};
 		local realmName;
```
Ellypse: This variable is never used 🤔

Issue #60, opened 2018-01-11
Add patch 7.3.5 images to the images browser


Ellypse:
- [x] Question frame images (screenshot of each zone)

Ellypse:
Images were done by @Solanya 
Now we also need to update the list of icons.

Solanya:
I can do that tonight.
Tried to look for a listfile for musics but the most updated I found doesn't have the 7.3.5 musics yet. :(

PR #64, opened 2018-01-15
Added 7.3.5 quest & allied races images

Solanya:
why can't i hold all those images ?!


Solanya:
Oh yeah #60 hi issues tracker

PR #86, opened 2018-02-28
Add image alignment feature

### LINKED ISSUE #31: Allow image tag to indicate alignment
Currently, TRP3 images are hardcoded as centered in the HTML version rendered. It could be interesting to add a parameter in the image tag to change the alignment attribute of the

Meorawr:
This allows specifying an (optional) alignment in an `{img}` tag replacement, so the full form now is `{img:src:width:height[:align]}`. Fixes #31.

If the alignment is not given, or is otherwise invalid, it will default to using center to match existing behaviour. When inserting an image, the alignment isn't specified due to some problems outlined below.

Unfortunately it seems the WoW has some issues; images aligned either left or right don't seem to stack correctly in the same way centered ones do. You can kind of see it in the test output, but it's more noticeable if you for example have two large right-aligned images right after each other in the markup - they'll start overlapping.

There's also a lack of any text wrapping with regards to paragraphs and images defined before them, but I'm not sure what can be done about that; if you've got a right-aligned image followed by a wordy paragraph in the markup the text will be rendered across the page and on top of the image.

I think the feature could have some limited use even with the known issues above, but that's up to you.

Test cases:

```
Backwards compatibility test. Standard image without an alignment - expect centered:
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32}

New functionality test. Expect centered, left, and right aligned in-order:
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:c}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:l}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:r}

The below two should be centered - test invalid alignments like "aaa" and "!":
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:aaa}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:!}

Only the first of the below should show (centered); test if there's a colon indicating an alignment but no actual value, and what happens if you have lots of colons:
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32::}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32::::}

If for some reason we have high unicode (Cyrillic) characters in the alignment, don't fall over - all three should show as centered:
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:П}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:Ф}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:ФФФ}

Expect the first three here to fail due to missing parameters, the last two should show centered:
{img:::}
{img:Interface\QUESTFRAME\UI-HorizontalBreak::}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32}
{img:Interface\QUESTFRAME\UI-HorizontalBreak:256:32:}
```

And the results:

trp-image-alignment-test

### DIFF
--- a/totalRP3/core/impl/utils.lua
+++ b/totalRP3/core/impl/utils.lua
@@ -42,7 +42,7 @@ local loc = TRP3_API.locale.getText;
 
 -- WOW imports
 local pcall, tostring, pairs, type, print, string, date, math, strconcat, wipe, tonumber = pcall, tostring, pairs, type, print, string, date, math, strconcat, wipe, tonumber;
-local strtrim = strtrim;
+local strsplit, strtrim = strsplit, strtrim;
 local tinsert, assert, _G, tremove, next = tinsert, assert, _G, tremove, next;
 local PlayMusic, StopMusic = PlayMusic, StopMusic;
 local UnitFullName = UnitFullName;
@@ -314,7 +314,7 @@ function Utils.str.getUnitID(unit)
 	return playerName .. "-" .. realm;
 end
 
-local strsplit, UnitGUID = strsplit, UnitGUID;
+local UnitGUID = UnitGUID;
 
 function Utils.str.getUnitDataFromGUIDDirect(GUID)
 	local unitType, _, _, _, _, npcID = strsplit("-", GUID or "");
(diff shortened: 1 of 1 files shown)


Ellypse:
Hi. Thank you so much for this amazing pull request. Great job here.

I think the text cannot wrap around the images because the image tag has to be outside of a <P> tag, but the text will be inside the next <P> tag, causing line breaks are all the things for paragraphs.

I had heard of the image stacking issue before, and I have seen it happen with inline icons too. I'm not sure there is anything we can do about it, unfortunately.

I won't have time to review everything today, but I will look into approving the request in the coming days. I'm so glad you did the regex part, this is one of my weaknesses and I was dreading having to do it, you are helping me big time here.

Thank you again!

Meorawr:
You're more than welcome, I love the addon too much to pass up an opportunity to poke my head in and fix a few low hanging fruits! 😄

PR #87, opened 2018-02-28
Disable map scanning button while initialising

### LINKED ISSUE #85: Disable the map scans button when broadcast channel has not been joined yet
### What steps will reproduce the problem?

When logging into the game, the broadcast channel is only joined after a while. In the meantime, the user can try to run a scan on the map and will get an empty result since they have not joined the channel yet.

### What is the expected output? What do you see instead?

The map scans button should be disabled (with a tooltip indicating the thing is still firing up) until the broadcast has been joined.

### What version of Total RP 3 are you using?

1.3.1

Meorawr:
Fixes #85.

Adds in a few extra events for the lifecycle of the broadcast process (`BROADCAST_CHANNEL_CONNECTING`, `BROADCAST_CHANNEL_READY`, `BROADCAST_CHANNEL_OFFLINE`) that the map module listens to and adjusts the state of the button appropriately.

Default state of the button remains the same, but we always fire a broadcast related event once `WORKFLOW_ON_LOADED` is called so it's unlikely you'd be able to log in and open your map fast enough to press the button before we get that handled.

Tested the obvious case of broadcast initialising and verified the button is disabled and its tooltip is set appropriately, and also the case where broadcast is configured to be disabled (the button will remain disabled and the tooltip will say you've set it up such).

Haven't checked the case of what happens if the channel is password protected or if you've hit the channel limit, but it all looks like it'll work. Can verify later.


totalRP3/core/impl/communication_protocol_broadcast.lua
```diff
@@ -279,8 +290,12 @@ Comm.broadcast.init = function()
 	-- When you are already in 10 channel
 	Utils.event.registerHandler("CHAT_MSG_SYSTEM", function(message)
 		if config_UseBroadcast() and message == ERR_TOO_MANY_CHAT_CHANNELS and not helloWorlded then
-			Utils.message.displayMessage(loc("BROADCAST_10"));
+			local message = loc("BROADCAST_10");
+			Utils.message.displayMessage();
```
Ellypse: You forgot to pass the message to the displayMessage() function here ^^

PR #89, opened 2018-03-01
Implement reordering of additional characteristics

### LINKED ISSUE #51: Re-organize additional info slots
>jessicaohwell, comment #798
>Hey! I love TRP, but something that bugs me is the inability to move around the Additional Information slots. I like to keep things organized, so sometimes if I add a new slot and I want to move it up by "nickname" or "alignment" I end up having to delete every slot I have below them in order to have the new one up higher. I hope that makes sense, but having some way to re-order the additional info slots would be REALLY helpful.

Meorawr:
Fixes #51.

The list can be reordered while editing the profile by dragging and dropping the icon (we're using that as the handle). The values correctly persist when saved.

The drag/drop operation is immediate, so when you move an item over another it'll take its place rather than waiting for you to release the mouse button. This is a bit simpler to implement, since otherwise we'd need an indicator texture to say where it's going to drop.

In addition, if the additional info fields are near the boundaries of the scroll frame, it'll scroll in the correct direction automatically if you go past them so you can see where you're dragging to.

d191aab507028c222dd56f227ad31b85 1

### DIFF
--- a/totalRP3/modules/register/characters/register_characteristics.lua
+++ b/totalRP3/modules/register/characters/register_characteristics.lua
@@ -791,6 +791,16 @@ local function onMiscInfoDragStart(handle)
 	local ticker = C_Timer.NewTicker(MISC_INFO_DRAG_UPDATE_PERIOD, onMiscInfoDragUpdate);
 	ticker.handle = handle;
 	handle.miscInfoTicker = ticker;
+
+	-- Stick the icon of the item in question onto the cursor for some feedback.
+	-- Also throw in some sound cues a-la spellbook drag/drop.
+	local node = handle.node;
+	SetCursor(node.Icon.Icon:GetTexture());
+
+	-- For some reason there's no constant for the pickup sound effect
+	-- used by ability icons. Logically this'd be present as
+	-- IG_ABILITY_ICON_PICKUP.
+	PlaySound(837);
 end
 
 --- onMiscInfoDragStop is called when a handle is no longer being dragged.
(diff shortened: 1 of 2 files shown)


Meorawr:
If anyone feels that we should have a separate handle to the left of the icon for drag/drop I'd be fine with implementing that. Implementation is abstract enough to make it a non-issue.

Personally I'm a bit torn on making the icon the handle; from a UI perspective it'd feel more logical that if I dragged and dropped an icon that I'd be moving the icon itself from one item to another, not reordering the list items.

Ellypse:
I agree that it is okay to use the icon as a handle and not have an actual dedicated handle. The game's UI is usually using icons to drag and drop. But we probably need some kind of visual feedback for the user to confirm that the drag is happening, either by changing the cursor or attaching the icon to the cursor. I have a module in my Ellyb library to attach icons to the cursor easily https://github.com/Ellypse/Ellyb/wiki/Cursor#seticoncursortexture-x-y

Meorawr:
One issue with that function is the icon doesn't seem to overlay the window - it's only visible when I take my mouse out of it.

e30058881b55e8346e38b8aff5e5adb6 1

Ellypse:
Oh, yeah, I should probably increase the texture strata 🤔 SetCursor is good for now, since we are not moving the cursor outside of the frame (which would remove the texture)

totalRP3/modules/register/characters/register_characteristics.lua
```diff
@@ -791,6 +791,16 @@ local function onMiscInfoDragStart(handle)
...
+
+	-- Stick the icon of the item in question onto the cursor for some feedback.
+	-- Also throw in some sound cues a-la spellbook drag/drop.
+	local node = handle.node;
+	SetCursor(node.Icon.Icon:GetTexture());
+
+	-- For some reason there's no constant for the pickup sound effect
+	-- used by ability icons. Logically this'd be present as
+	-- IG_ABILITY_ICON_PICKUP.
+	PlaySound(837);
```
Ellypse: I **really love** the idea of sound feedback!
But then you should use the `TRP3_API.ui.misc.playUISound(soundID)` function. The other dev who worked with me on Total RP 3 insisted that if we have UI sounds we have an option to disable them in the settings, so using that function will follow that setting :)

Ellypse:
Thanks again. Your work is stellar! Your help is truly appreciated 🙏

PR #90, opened 2018-03-01
Implement higher-resolution personality traits

### LINKED ISSUE #39: Convert personality traits to bars and change scale to 20
Some users want to be able to set their traits as 0% or 100% instead of being stuck with at least 1 dot in one trait.

Since having all the dots in the same color can be confusing to know which trait it is, we'll probably have to work on the UI to highlight the traits with the corresponding color?

Meorawr:
This'll fix #39, but it isn't quite done yet. Notably the UI aspect isn't implemented right now. Thought I'd open the PR early since this could be a big one and I don't want to be the person who broke everyones profile 😄.

Current implementation as of this commit includes:

* Profile migration/patch to schema version 6 which handles upscaling the `PS.VA` field and storing it in a new one (`PS.V2`).
* Necessary changes to `register_characteristics.lua` to make it read from the `PS.V2` field.
* Disabled UI until I put in the slider bar.
* Changes to `PS.V2` are backported to the `PS.VA` field too, so older clients can still receive updated values.

My main question right now is would there be anywhere else that the `PS.VA` field gets referenced? The only place I can think of is `register_msp.lua` which (at a glance) seems to correspond to the network comms protocol. However `CHARACTERISTICS_FIELDS` doesn't reference `PS` at all, so either I'm wrong or it's supposed to be in that table but is missing.


Ellypse:
The personality traits are not in the Mary Sue Protocol, the field you found in `register_msp.lua` is for a different thing.

Meorawr:
Figured as much wrt. MSP protocol in the end. I did a test ingame with a friend to see if my 0-20 range values were downgrading correctly for older clients (they do, it seems!) and had logging in the MSP bits and came to the conclusion that it indeed was an entirely other beast shortly after.

UI bits are ~60% there now. Notably missing correct colours and the stack count on icons.

bdce06eb97add9e58db06f3125403ceb 1

(The GIF loops awkwardly at the end, in case it looks like they didn't save correctly).

Meorawr:
Pretty happy with it as it is now, needs a good set of testing though.

530d94134191a4112a37219f0168fa44

Ascathor:
I would suggest an initial doubling of the attributes to reflect the new changes properly. Opinions ?

Solanya:
What do you mean by doubling ? The attributes are already changed proportionnally from 0-6 to 0-20.

Ascathor:
Do we change existing values ? As far as I saw, it's left untouched (at least that's how code and comment describe it). If someone has a profile already and is checking it, they'd see values corresponding to the current implementation - not the new one. At least, as far as I saw.

Solanya:
totalRP3/modules/flyway/flyway_patches.lua lines 55 to 96, updating all stored profiles (player and register) the first time the old version is encountered.

Ascathor:
Ah, comments were indicated otherwise.

I stand corrected, all good then \o Looks great, good job

Meorawr:
Just to clarify, _player_ profiles are updated with the values increased proportionally.

The last commits made it so that the register profiles aren't touched directly during the migration, but there **is** a [function](https://github.com/Ellypse/Total-RP-3/pull/90/commits/d50f7d583c5f9c945e607693c632a63676c106d1#diff-285080bd402ae0ea66d60b4a021858b3R175) call handled on the UI side of things that translates and increases the `VA` attribute if there's no `V2` present on the profile.

So yeah, net result is they're all properly inflated.

I did it this way in the end since I'd otherwise also need to hook into the data exchange code to also apply the `VA` to `V2` change on new profiles that come in. Ultimately it's a lot easier to just translate the old values at display time than to have to sprinkle migration code in lots of different places.

Ellypse:
You have done an amazing work, this is better than what I would have done. Thank you again for your help and all the good work!

I just have two small requests, to preserve consistency across the add-on:
- Instead of right-clicking the icon to change the color, use a [TRP3_ColorPickerButton](https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/core/ui/widgets.xml#L569) like the ones we have for the class and eye color (you can look at how it is implemented by searching for the `TRP3_RegisterCharact_Edit_ClassButton` for example). Since it already implements the behavior to switch between the different color pickers and to reset the color, it's less code duplicated and it will be easier to maintain in the future.
- Use [Ellyb's Color](https://github.com/Ellypse/Ellyb/wiki/Color) class to handle colors, instead of using the hexadecimal code (parsing string is expensive from what I understand). The constructor itself accepts many formats, so you can create your default colors using the hexadecimal code if you want (it's done once at launch, it's okay), and you can retrieve the RGB values (and maybe store the value as an RGB table?). In the long run, I want that class to be used as much as possible for colors to have consistency in the code and to have one place to maintain. I will be slowly migrating existing code to this in the future.

totalRP3/modules/register/characters/register_characteristics.lua
```diff
@@ -448,13 +446,25 @@ local function saveInDraft()
...
 			-- If not a preset
 			psychoStructure.LT = stEtN(psychoLine.CustomLeftField:GetText()) or loc("REG_PLAYER_LEFTTRAIT");
 			psychoStructure.RT = stEtN(psychoLine.CustomRightField:GetText()) or loc("REG_PLAYER_RIGHTTRAIT");
 
-			psychoStructure.LC = stEtN(psychoLine.LC);
-			psychoStructure.RC = stEtN(psychoLine.RC);
+
+			local lc = psychoLine.LC;
+			if lc then
+				psychoStructure.LC = { r = lc:GetRed(), g = lc:GetGreen(), b = lc:GetBlue() };
```
Ellypse: I'm gonna need to make a color:GetRGBTable() method to help with this kind of usage 🤔

Meorawr: Yeah, was thinking that it's not the prettiest. I can store it as hex codes again - I'm probably the only one storing with the new format right now after all!

Ellypse: I have added the `color:GetRGBTable()` and `color:GetRGBATable()` methods to Ellyb and I have update the submodules in Total RP 3 to include the latest changes. You can merge the development branch into your branch to get this and use it :)

totalRP3/modules/register/characters/register_ui_characteristics.xml
```diff
@@ -237,6 +237,7 @@
 					<Anchor point="TOPLEFT" relativeKey="$parent.Bar" x="-16"/>
 					<Anchor point="BOTTOMRIGHT" relativeKey="$parent.Bar" x="16"/>
 				</Anchors>
+				<HitRectInsets left="12" right="12" top="0" bottom="0" />
```
Meorawr: Just to clarify, we don't use the 16px offset for the inset here because otherwise the extreme edges of the slider wouldn't be clickable to set the bar to its min/max value. 12px seems to be a good compromise between not cutting either widget off too much.

Meorawr:
d2b0daeaadbe389b94b392e98d0a0806

The UI is a little bit squeezed for space in its default size with the swatches which was why I'd used a dropdown approach at first. Of course as you can resize the UI it's not too big a deal, and it came out better than I'd expected.

Meorawr:
One minor irritation is when you add/delete a trait all the colour swatches play the shine animation as if they've just been changed.

Ellypse:
I think the current position at the edges is better to avoid confusion as you said.
I'm gonna look into either removing the animation (it's getting old, I'm not sure I like it still. It's kind of Powerpoint-ish) or making the animation optional when setting the value.

Meorawr:
At the very least I'd think setColor should only play it if the r/g/b have changed. It'd also be nice if setColor could accept an instance of `Ellyb.Color` (perhaps just `type(r) == "table"` and assume it is one?).

Even nicer still would be spitting out a Color instance as a fourth (fifth?) argument in the `onSelection` call. I say fifth because I'm not sure if you plan on putting alpha support in - if so it'd be a good idea to set the fourth to nil for now I guess.

And while I'm writing a wishlist - perhaps check if there's a non-nil `callbackObject` field present on the button itself, and if so pass that as the first argument to `onSelection`. That way you can set `onSelection` to be a method on any arbitrary table without needing to create forwarding closures.

Ellypse:
I agree with the idea of only animating if the color is different?

I was actually in the process of moving the color picker button to a widget using mixins. Using clearly defined widgets and mixins is a big part of the modernization I want to do for TRP. I had the untested code on a local branch, I just pushed it to [feature/color_picker_button_mixin](https://github.com/Ellypse/Total-RP-3/tree/feature/color_picker_button_mixin) if you want to give it a look. It probably has some issues since I did that today while at work (it probably needs to be nil-safe in some places).
The tooltip functions will have to be changed in the future to use Ellyb's Tooltips (still in WIP) and automatically provide the "Click: Set color\nRight-click: Reset" part of the tooltip for all instances of the button while offering to add more text dependant on each instance.

Meorawr:
I think the current implementation is okay enough and would rather not block this PR to wait on that, however I'll sprinkle some FIXME comments in to say that when it lands this should be migrated over if that's acceptable.

That said I _love, love, love_ the approach of using mixins for these types of things, and the changes on that branch look great so far 👍 .

Issue #63, opened 2018-01-15
Add copy option for some text fields


Ellypse:
It would be interesting to allow people to copy some of the information in someone's profile or in some other parts of the UI (kind of like you could copy someone's info from a web page).

How I see it is that we would create a widget (so the code can be easily abstracted and re-used everywhere) that when you *Ctrl + Click (?)* on the text, the widget turns into an edit box that is autofocused with the text preselected so they can Ctrl + C. When the field loses focus, the edit box goes away and the text element is put back in.

It could be used on the profile page (player, companions), in the directory, and more.
It could probably get tied in with issue #37 where Ctrl + Clicking an at-first-glance slot would open a custom copy frame.

Meorawr:
One catch of course is FontStrings don't support script handlers. We'd need to upgrade them to Frames and add forwarding calls for SetText and co. Not too much of a pain.

Easy solution would be to add a Mixin that has the methods for popping up a small dialog with the text itself in, similar to most add-ons let you copy links in chat for example.

I'd be wary about making the field itself become an edit box. It might visually come off as a way of making the field itself look editable in-place.

Ellypse:
Hmm, I haven't thought of that, good call.

This ticket was really an idea thrown at the wall to see what could be made of it. It hasn't been requested (not that way at least). It's really a "what if" thing.

Meorawr:
My feeling on the usefulness is that it's not too great; off the top of my head most fields are quite short with the exception of the about page. If that were copyable it'd need a tiny bit of custom logic since, well, HTML.

I'll put this one in the very furthest back reaches of my mind for now.

PR #92, opened 2018-03-02
Implement copy/pasting of glance traits

### LINKED ISSUE #37: Add a way to copy-paste someone's at-first-glance data
It could be interesting to offer a way to copy someone else's at first glance to then paste it into one of our own slot.

Meorawr:
Fixes #37. Tested with my own profile and companions, as well as copying from profiles in the register. Unfortunately trying to find someone ingame right now with a companion profile is proving impossible, so haven't tested copying another persons companion traits.

This is provided as part of the right-click menu. While another ticket (#63) does suggest the use of Ctrl-Click for the copy operation, there's the question of how do you then paste it. A button could be added to the editor instead, but this approach feels fine (I mean, _everything_ lets you right-click for copy and paste normally, right?).

A lot of this would benefit from a better dropdown system, as the current implementation relies on the glance buttons having global names to perform the copy at least.


Meorawr:
Did some additional testing; confirmed that presets aren't broken with the dropdown changes and verified that I can copy another user's companion glances onto my own just fine.

PR #93, opened 2018-03-03
Map scanner blip colors

Meorawr:
No actual ticket for this feature, I came up with the idea last night though and so far I'm loving the potential usefulness of it.

24332464694926e3ea7f46b1f0e35195

The current implementation colours blips according to the relationship status of a characters' profile (so hated = red, friend = green, etc.) as well as colouring based upon whether or not that character is on your friends list.

The priority right now is friends > relationships > others. I keep flip-flopping this in my mind, so it could be made into an option I guess.

The blips for players without relationships are unchanged and remain the same colour and texture. Ones that do get changed use the vertex-colourable party member minimap blip and get coded accordingly. In particular, WoW friends get a purple-ish colour and Battle.net friends get the standard BNet whisper colour.

There's a few other changes I want to make - in particular a filtering system. Clicking the scan button would have some additional options like "Show only friends", "Show characters with relationships", and so on.

Finally I want to work a bit on the tooltip display, particularly for large player clusters. I'm thinking if the number of entries is over some arbitrary lower bound, we display the tooltip in a two-column format. If it's over a higher bound, we cut it off and just have a line at the end saying "<x> additional characters in this area.".

I also want to then group the tooltip a bit, so you'd have blips that represent characters with relationships or friends in separate sections at the top.

Thoughts? Opened the PR early in case anyone thinks this is a terrible idea 😄 .

### DIFF
--- a/totalRP3/core/impl/globals.lua
+++ b/totalRP3/core/impl/globals.lua
@@ -89,7 +89,31 @@ TRP3_API.globals = {
 	PSYCHO_MAX_VALUE_V2 = 20,
 	PSYCHO_DEFAULT_LEFT_COLOR = Ellyb.Color(255, 140, 26):Freeze(),
 	PSYCHO_DEFAULT_RIGHT_COLOR = Ellyb.Color(32, 208, 249):Freeze(),
-}
+};
+
+--- RELATIONS is a list of (backwards-compatible) relationship IDs.
+local RELATIONS = {
+	UNFRIENDLY = "UNFRIENDLY",
+	NONE = "NONE",
+	NEUTRAL = "NEUTRAL",
+	BUSINESS = "BUSINESS",
+	FRIEND = "FRIEND",
+	LOVE = "LOVE",
+	FAMILY = "FAMILY",
+};
+TRP3_API.globals.RELATIONS = RELATIONS;
+
+--- RELATIONS_ORDER defines a logical ordering for relations.
+local RELATIONS_ORDER = {
+	[RELATIONS.NONE] = 0,
+	[RELATIONS.UNFRIENDLY] = 1,
+	[RELATIONS.NEUTRAL] = 2,
+	[RELATIONS.BUSINESS] = 3,
+	[RELATIONS.FRIEND] = 4,
+	[RELATIONS.LOVE] = 5,
+	[RELATIONS.FAMILY] = 6,
+};
+TRP3_API.globals.RELATIONS_ORDER = RELATIONS_ORDER;
 
 local emptyMeta = {
 	__newindex = function(_, _, _) end
--- a/totalRP3/modules/map/map_markers.lua
+++ b/totalRP3/modules/map/map_markers.lua
@@ -19,6 +19,7 @@
 
 TRP3_API.map = {};
 
+local Ellyb = TRP3_API.Ellyb;
 local Utils, Events, Globals = TRP3_API.utils, TRP3_API.events, TRP3_API.globals;
 local Comm = TRP3_API.communication;
 local setupIconButton = TRP3_API.ui.frame.setupIconButton;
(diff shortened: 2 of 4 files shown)


Ellypse:
I like the idea of coloring the dots according to the relationship, but I’m not sure about the friends and BN friends part. It’s a non RP information and it’s never used elsewhere in the addon.
Also I don’t want to fill the scanning menu with many options for individual cases. I would rather keep one scanning option, but have people make use of the colors to differentiate people with relationships from others.

Meorawr:
Removing the friends stuff would simplify it greatly at least, so I'm all for killing that. I'm sure there's other addons around that put blips on the map for friend locations anyway.

Meorawr:
Very much a WiP (non-localised group titles, bit messy internals), but how does this look?

b63a3a0ba2bf08fd1bdf49c2af2f72a6

Idea is profiles that have relationships set would be grouped up on the right-side of the tooltip together, away from the other characters that you haven't configured them for. That way in crowded spots they'll always be visible near the top of the tip.

Solanya:
Just my personal opinion before Ellypse gives his feedback : it looks crowded. Having a small character colored with the relationship color next to the name would make them stand out in the list without taking a lot of space.

Ellypse:
I don't like that you have two informations on the same line, it can be confusing. Is FRIEND related to Wrago or is it a title above Liam? It's not obvious.

I think it would be better to keep everything aligned to the left, with the categories (in yellow, like other dropdown/tootltip categories) like so:
> **Characters**
> Wrago
> Dallgen
> **FRIENDS**
> Liam Qhuin
> **LOVE**
> Nicina Emberwind

Meorawr:
Agreed, I wasn't a huge fan of the multi-column stuff and it was really horrible to implement in the end.

Meorawr:
Layout 1 - Displaying a small icon next to characters names. Coloring here is temporary (fixed to red), it'd match the actual colors in use for the relationship levels if chosen. This is my favourite as it's very compact, however the implementation isn't the nicest due to limitations with texture embeds in fontstrings.

dc939a3f8d2d993dd833769bae7c2f28

Layout 2 - Single column with category headers. Same with the coloring. Looks okay-ish?

5d3c0a43a8c2c6024cf5e0f3a0badb3b

Layout 3 - Dual-column with the relationship level on the right side of the tooltip. Colors again temporary. This might look weird if you have 5 "Friend" characters in one spot.

0feca6076d5ad875c49cfd4cba821f18

Layout 4 - Not actually implemented, but would be a hybrid of 1 and 3 whereby the colored icon would be displayed in a right-column.

Ellypse:
From experience with the OOC red indicator, using only icons can be confusing for some people, I think we need the name of the relationship (especially if we ever do custom relationships, with custom colors).
As you said, I don't think layout 3 would look good if you have a bunch of people with the same relationship in the tooltip.

I think layout 2 is the best. The relationship name should be using normal case, we never used all-uppercase text in the add-on. Also, it might be better to use a separator texture (like we can see in some dropdowns) instead of blank lines to separate categories (consistency, and it's explicit that it is a separation, as we might have bugs where the name is blank (or really dark and not visible) and create confusion)

Meorawr:
Capital letters were just another temporary artifact that I forgot to mention - was displaying the raw ID of the relationship in the screenshots rather than plugging into `loc()`.

Will iterate some more on layout two tonight.

Meorawr:
Single-column display, categorised with correct colouring and localisation strings, as well as the same separator used in dropdown menus. I think it looks great.

7496eb778cedf57664cb9ab906567377

Meorawr:
I've rebased all the changes into a single commit to remove all the wip cruft from the various layouts and scrapped decisions.

Ellypse:
That's perfect! As always, stellar work, the code is A+! Thank you so much for helping :)

(I won't judge you for loving random people, you do you 😜)

PR #98, opened 2018-03-06
Convert width and height to numbers in img tags

### LINKED ISSUE #97: Unable to view or edit profile, error on About page
### What steps will reproduce the problem?
In a previously established companion pet profile, I had used a picture of the Pet Battle Arcane Storm header... thing. 
### What is the expected output? What do you see instead?
I added " c" (with the space) to the end of the {bit} in an attempt to center the picture. Now I everything appears blank and I am unable to edit the profile. I also tried deleting the profile, but it kept reappearing in the list. I think I must have forgotten how to get it centered and typed the wrong thing, as I feel like I've done so before on different pictures. Though maybe not in a Pet profile.
Here is a screenshot. 

### Did you try having Total RP 3 as the only enabled addon and everything else (especially something like ElvUI) disabled?
No, the bug report seems to only mention TRP.
### What version of Total RP 3 are you using?
Latest I think. 1.3.1
### Was it working in a previous version? If yes, which was the last good one?
Uncertain. Maybe?
### Do you have an error log or screenshot of what happened?
Yes. https://puu.sh/zBUZj/bb42e51611.png
ss 2018-03-06 at 01 06 33
> If you don't see any errors, make sure that error reporting is enabled (`/console scriptErrors 1`) or install the [BugSack](https://mods.curse.com/addons/wow/BugSack) add-on.

### Please provide any additional information below.
Sorry I dun bork it 😢 
### If your issue is related to profile data, please provide your data file.
When I try to drag it in here as instructed, it says the file type isn't supported and only takes images, DOCX, GZ, LOG, PDF, PPTX, TXT, XLSX or ZIP. I am certain I am grabbing the .lua and not the .lua.bak. I'll put it in a ZIP instead. 
[totalRP3.zip](https://github.com/Ellypse/Total-RP-3/files/1786198/totalRP3.zip)

> It is the `totalRP3.lua` file (not the .lua.bak files!) located in your `World of Warcraft/WTF/Account/ACCOUNT_NAME/SavedVariables` folder (replace `ACCOUNT_NAME` with your actual account name, visible on the Battle.net launcher, usually WoW1 for accounts created after the introduction of Battle.net). You can just drag and drop it here, no need for a third party service.

Meorawr:
Fixes #97.

Not the greatest solution but we don't error out completely now, and hopefully the user will notice the image isn't displaying (the patch defaults them to zero in the worst case). In the future we could have some validation code I guess?

### DIFF
--- a/totalRP3/core/impl/utils.lua
+++ b/totalRP3/core/impl/utils.lua
@@ -870,6 +870,11 @@ Utils.str.toHTML = function(text, noColor)
 			-- you'll get the old default of center.
 			align = alignmentAttributes[align] or "center";
 
+			-- Don't blow up on non-numeric inputs. They won't display properly
+			-- but that's a separate issue.
+			width = tonumber(width) or 128;
+			height = tonumber(height) or 128;
+
 			-- Width and height should be absolute.
 			-- The tag accepts negative value but people used that to fuck up their profiles
 			return string.format(IMAGE_TAG, img, math.abs(width), math.abs(height), align);


Ellypse:
Oops, that comment was from my phone, I had the wrong account in. Still, that's me ^^"

Meorawr:
Yeah, I didn't like the zero. Let's make it 128...

Meorawr:
Rebased, 128px now.

PR #100, opened 2018-03-06
Dashboard Revamp

### LINKED ISSUE #99: Make contributors list a variable in About screen
For now, every time we have a new contributor joining the project we have to update the list in the About screen of the Dashboard to add them, then make sure the changes are populated for all the localizations.

We would want to have a table (similar to how I handle Patreon supporters in the same text) of { contributorName = role }  then output that list in the text.

While we're at it, also add Contributor to Solanya.

Meorawr:
Fixes #99.

This requires manual changes to the non-English locales that are outlined below.

Credits are defined in an ADDON_CREDITS table in the globals file. This table has a subtable per-role which contains a flat pre-ordered list for display in the appropriate sections.

```lua
local ADDON_CREDITS = {
	[ADDON_CREDITS_ROLES.AUTHORS] = {
		[[Renaud "{twitter*EllypseCelwe*Ellypse}" Parize]],
		[[Sylvain "{twitter*Telkostrasz*Telkostrasz}" Cossement]],
	},
        ...
};
```

The ADDON_CREDITS_ROLES table defines the available roles.

```lua
local ADDON_CREDITS_ROLES = {
	AUTHORS = "AUTHORS",
	CONTRIBUTORS = "CONTRIBUTORS",
	COMMUNITY_MANAGERS = "COMMUNITY_MANAGERS",
	TESTERS = "TESTERS",
	GUILD_MEMBERS = "GUILD_MEMBERS",
};
```

The ADDON_CREDITS_ROLE_GROUPS table forms a logical mapping between the individual roles and the sections in the About text. The current about text has four sections - the contributors and community managers are merged - so we use this logical mapping to do exactly that.

```lua
local ADDON_CREDITS_ROLE_GROUPS = {
        -- When %{CONTRIBUTORS} is replaced in the About text format string, both of the roles inside
        -- this group will be inserted into that section in-order.
	[ADDON_CREDITS_ROLES.CONTRIBUTORS] = {
		ADDON_CREDITS_ROLES.CONTRIBUTORS,
		ADDON_CREDITS_ROLES.COMMUNITY_MANAGERS,
	},
        ...
};
```

This design allows us in the future to add more roles if needed and assign them to an existing logical group without needing to then update each localisation string; you can just add the new roles to a pre-existing group.

This is backwards compatible with existing locales however they won't display the new data unless they include a set of **named** replacements in the form `%{ROLE_GROUP}`. For example, `%{AUTHORS}` would be replaced with all of the members of the `AUTHORS` group and its associated role(s).

This decision was made because replacements like `%s` are positional; if one day we were to add another logical group to display before the QA Testers section, we'd need to update the `string.format()` call to account for the argument order change as well as every localisation to include the new `%s`. Named replacements let us do these sorts of things progressively by not tying the presentation of the text to the actual implementation of the formatting code.


Meorawr:
Obligatory screenshot to prove I'm not a total madman with these changes.

e8971acca16b90e4cb9a4763e304091a

Ellypse:
You are madman 😛
In return for this madness, I'm gonna do extensive code review 😝

Solanya:
I'm extensively still not a contributor.

totalRP3/modules/dashboard/dashboard.lua
```diff
@@ -146,6 +146,98 @@ local ShowingHelm, ShowingCloak, ShowCloak, ShowHelm = ShowingHelm, ShowingCloak
...
+			creditsWithRole[i] = loc("THANK_YOU_WITH_ROLE"):format(name, localisedRole);
+		end
+
+		credits = creditsWithRole;
+	end
+
+	-- Net result here is we'll insert a newline and a "- " between each
+	-- item in the list. We'll prefix it with a "- " in the format call so
+	-- the first item is properly indented too.
+	local formattedMembers = ("- %s"):format(table.concat(credits, "\n- "));
```
Ellypse: There is a local import of `table.concat` declared at the top as `tconcat`

totalRP3/modules/dashboard/dashboard.lua
```diff
@@ -146,6 +146,98 @@ local ShowingHelm, ShowingCloak, ShowCloak, ShowHelm = ShowingHelm, ShowingCloak
...
+
+	-- You've got a few roles, that's fine. Bit more effort.
+	local groupMembers = Utils.table.getTempTable();
+	for i = 1, #group do
+		local role = group[i];
+		local members = Globals.ADDON_CREDITS[role];
+
+		-- We want to include the role information when there's a single
+		-- logical group with multiple roles attached.
+		table.insert(groupMembers, getFormattedAboutCreditsMembers(members, role));
```
Ellypse: Import `table.insert` at the top of the file, to get a local reference and spare a few global lookups. Especially here since we are in a loop of unknown size.

totalRP3/core/impl/globals.lua
```diff
@@ -142,3 +142,82 @@ TRP3_API.globals.build = function()
...
+		[[Sylvain "{twitter*Telkostrasz*Telkostrasz}" Cossement]],
+	},
+
+	[ADDON_CREDITS_ROLES.CONTRIBUTORS] = {
+		[[Connor "{twitter*Saelorable*Sælorable}" Macleod]],
+		[[Daniel "Meorawr" Yates]],
+	},
+
+	[ADDON_CREDITS_ROLES.COMMUNITY_MANAGERS] = {
+		[[{twitter*Solanya_*Solanya}]],
```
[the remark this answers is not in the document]
Meorawr: Hmm, multiple roles for a single person... Yeah that's a good idea.

totalRP3/modules/dashboard/dashboard.lua
Meorawr: I might also reformat the imports to match the convention you were using in the color picker mixing branch if you're happy with that?

Ellypse: Sure. I'm trying to organize the imports in the new files to make it easier to look for stuff when we need that, but I did not have the motivation to go through the old files. I'm updating the stuff I change on the go ^^

Ellypse: Oh, small side note. The IDE I am using (IntelliJ IDEA) cannot tell unused imports when there is more than one import on the same line. Those were usually made by Kaji back then. I also find them not easy to read when they are quite long. So if I was to refactor the imports I would make them all one import per line, even if it takes more space at the top of the file, for the benefit of it being easier to read and the IDE highlighting unused imports.

Meorawr: Re: 1 line per import, that's more than fine with me. Considerably easier to read.

renaudparize:
Please keep your classes!
We want to have a carousel to spotlight features at some point. What you did here will probably make this a whole lot easier!

renaudparize:
Dammit, wrong account again. Well you know who it is.

Ellypse:
But the copyrights are wrong. You created the files, not me 😛

Meorawr:
The carousel is actually kind of why I made the design like this, funnily enough. Would need a proper view insertion API like (eg. `TRP3_API.Dashboard.RegisterTab(viewClass)`) I think but this is a good enough start for that.

totalRP3/modules/dashboard/DashboardAboutTabView.lua
```diff
@@ -0,0 +1,340 @@
...
+
+	-- If this person is multi-talented and can perform not one, but MULTIPLE
+	-- roles then we need to join them up. Now, in all the supported locales
+	-- there is GUILD_MEMBER_TEMPLATE which exactly matches "%s, %s",
+	-- however we'll use our own GEN_LIST_DELIMITER just in case that one
+	-- ever changes purpose in some way.
+
+	-- Grab the first entry and then start the loop from the second index.
+	local output = loc("THANK_YOU_ROLE_" .. self.roles[1]);
+	for i = 2, roleCount do
```
Ellypse: Couldn't this be simplified by adding all the localized roles in a table and doing a `table.concat(roles, ", ")`?

totalRP3/modules/dashboard/DashboardAboutTabView.lua
```diff
@@ -0,0 +1,340 @@
...
+
+	-- This monstrosity below is a foul and arcane incantation that basically
+	-- matches the subset of C's printf that Lua actually implements, as
+	-- well as providing support for Blizzard's positional extension syntax.
+	local MATCH_STRING = "(%%((%d?)[%w_]*)$?([-+ #0]*%d*%.?%d*[diuoxXfFeEgGcsq]))";
+	return strgsub(formatString, MATCH_STRING, self.callback);
+end
+
+--- Internal handler invoked by string.gsub when it finds a specifier match.
+function StringFormatter:replace(full, key, keyFirstChar, specifier)
```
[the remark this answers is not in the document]
Meorawr: Yeah, wasn't too sure on what convention we're using for protected methods any longer as most the existing code is `camelCase`'d and all that. Will fix that one later.

totalRP3/modules/dashboard/DashboardTabFrame.lua
```diff
@@ -0,0 +1,330 @@
...
+end
+
+--- Unregisters all registered URL handler functions.
+function TRP3_DashboardHTMLContentMixin:UnregisterAllHyperlinks()
+	twipe(self.urlHandlers);
+end
+
+--- Replacement handler for SetText() which records the text for retrieval
+--  via GetText().
+function TRP3_DashboardHTMLContentMixin:setTextHooked(text)
```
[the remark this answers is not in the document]
Meorawr: Wouldn't hooking it in this way cause the `SaveText` to be invoked for every `SimpleHTML:SetText()` call, including those outside our widget, since the SetText method is itself stored on the shared metatable for a frame?

totalRP3/modules/dashboard/DashboardAboutTabView.lua
Meorawr: I'd actually made a change here last minute and didn't think of the simple solution of `table.concat(roles, LIST_DELIMITER)`, good catch.

`LIST_DELIMITER` is defined in the global strings as `%s, %s` but seems to get changed somewhere along the line to being `, `. I'd prefer using that localised constant for the delimiter just in case tomorrow the Queen decrees that henceforth all list items should be separated by banana emoji.

totalRP3/modules/dashboard/DashboardAboutTabView.lua
```diff
@@ -0,0 +1,340 @@
...
+function AboutTabView.static.GetTabTitle()
+	return loc("DB_ABOUT");
+end
+
+function AboutTabView.static.GetTabWidth()
+	return 175;
+end
+
+function AboutTabView:OnShow()
+	DashboardTabView.OnShow(self);
```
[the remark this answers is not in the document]
Meorawr: `self.super.OnShow(self)`* 😄

I'll make that consistent across the various view subclasses.

totalRP3/modules/dashboard/DashboardAboutTabView.lua
```diff
@@ -0,0 +1,340 @@
...
+local StringFormatter = Class("StringFormatter");
+
+--- Allocation pool for reusable instances.
+StringFormatter.static.pool = setmetatable({}, { __mode = "k" });
+
+--- Original allocator that we haven't hooked.
+StringFormatter.static.defaultAllocator = StringFormatter.static.allocate;
+
+--- Custom allocator that pools instances of this class for reuse.
+function StringFormatter.static:allocate()
```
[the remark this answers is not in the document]
Meorawr: I'd suggest a mixin rather than an inheritance structure for such a pool, but yes, was going to PR such a thing at some point.

totalRP3/modules/dashboard/DashboardAboutTabView.lua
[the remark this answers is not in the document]
Ellypse: It does let you access static fields from the instance using `__index` ^^ I do that a lot.
Although I actually think having `static` explicitly used is better for readability in most cases.

totalRP3/modules/dashboard/DashboardTabFrame.lua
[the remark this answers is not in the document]
Ellypse: No, as you hook the `SetText` of `self`, first argument. It will only hook the method for that specific instance. 
I am already using this in the `ChatLinkActionButton` to update the width of the button when we change its text, and I can confirm it doesn't hook all `SetText` of all Buttons ^^
You can see that usage here https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/modules/ChatLinks/ChatLinkActionButton.lua#L73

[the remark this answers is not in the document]
Meorawr: Perfect then, cleans that bit up nicely 😄 .

Meorawr:
I see you met my madderness and now I present you with my final form. Let's see you review THAT sunshine!

Fixed all the casing issues with regards to static functions and instance methods. I've nastily decided to hack the string formatting (and new `Funcs` utilities) into Ellyb if they're not present already.

This means when we get around to dragging those over and the library is updated here, we don't need to hunt around for references to some temporary code that was going to hopefully get moved in the next couple of days anyway. For the purposes of review pretend I'm not extending Ellyb and we'll get a PR sorted for that stuff tomorrow? 😇 

Other than that all the classes you pointed out have been split up and I've done some optimisations elsewhere to reduce the amount of needless string formatting and generation (primarily through `Funcs.memoize`).

The `TabView` API changed a bit; methods are `Show` and `Hide` because really they're supposed to be actions, not script handlers. In addition they no longer register themselves to a table; you're required to pass a `TabClasses` field containing the views you want on the instantiated `TabFrame` instance prior to calling `OnLoad` (you'll see what I mean in `TabFrame.xml`.

I also dropped the `Dashboard` prefix on any locals or filenames; it's the dashboard module after all - we don't really need the whole `Dashboard.DashboardXXX` stutter going on.

Meorawr:
Improved the memoizer to support n arguments and return values. Haven't fully tested it but it hasn't broken in terms of the dashboard usage requirements.

Meorawr:
Decision made to cull the memoizer (can't say I'll miss it :smile:). I'll get that done today.

Meorawr:
Rebased to pick up the latest development changes and squash it all down. RIP Cthulhu-esque memoizer.

Meorawr:
I've changed the Ellyb submodule to point to my `feature/trp3-dashboard-merge` branch, so I can remove the temporary extension hack for Strings.interpolate and pull in the latest changes without needing to copy/paste.

This should obviously be reverted back to your master branch before accepting the PR.

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "totalRP3/libs/Ellyb"]
 	path = totalRP3/libs/Ellyb
-	url = https://github.com/Ellypse/Ellyb
+	url = https://github.com/Meorawr/Ellyb
```
Ellypse: We can revert that now 🙂

Meorawr:
Rebased to remove the submodule change and squash the intermediate drycode fixes down. All works well.

totalRP3/tools/localization.lua
```diff
@@ -54,6 +54,7 @@ TRP3_API.loc = {
 
 	GEN_WELCOME_MESSAGE = "Thank you for using Total RP 3 (v %s) ! Have fun !",
 	GEN_VERSION = "Version: %s (Build %s)",
+	GEN_LIST_DELIMITER = "%s, %s",
```
Meorawr: That could actually be removed, come to think of it.

Meorawr:
I remember why I didn't use `self.class.super` for method calls now; it breaks in the case of inheritance.

Given a class structure of `A`, `B(A)`, and `C(B)`:

```lua
function A:Work()
    return 1
end

function B:Work()
    return 1 + self.class.super.Work(self)
end

local c = C()
c:Work()
-- Result: stack overflow
```

The `c:Work()` call would overflow the stack; `self.class` would point to `C` (and thus `self.class.super => B`) so we'd effectively be doing...

```lua
function B:Work()
    return 1 + B.Work(self)
end
```

Explicitly naming the class implementing the method is the only way around that.

```lua
function A:Work()
    return 1
end

function B:Work()
    return 1 + B.super.Work(self)
end

local c = C()
c:Work()
-- Result: 2
```

Ellypse:
That's super weird 🤔

PR #113, opened 2018-04-21
Fix for stopSoundID

Solanya:
Maybe the function should be renamed, but here's what's been done : 
- soundID as "0" works the same way as nil, as I seemingly can't broadcast nil.
- If the handle is on the Music channel, checking the handlerID to make sure we're using the right function for the job.
- The "stopped" boolean is there to handle an edge case where a player starts a music, replaced by another player's music. The first player would be able to stop the second player's music with his "Stop local music".

### DIFF
--- a/totalRP3/core/impl/utils.lua
+++ b/totalRP3/core/impl/utils.lua
@@ -1004,7 +1004,7 @@ function Utils.music.playSoundID(soundID, channel, source)
 	assert(soundID, "soundID can't be nil.")
 	local willPlay, handlerID = PlaySound(soundID, channel, false);
 	if willPlay then
-		tinsert(soundHandlers, {channel = channel, id = soundID, handlerID = handlerID, source = source, date = date("%H:%M:%S")});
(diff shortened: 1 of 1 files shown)


totalRP3/core/impl/utils.lua
```diff
@@ -1018,8 +1018,13 @@ end
 
 function Utils.music.stopSoundID(soundID, channel, source)
 	for index, handler in pairs(soundHandlers) do
-		if (not soundID or handler.id == soundID) and (not channel or handler.channel == channel) and (not source or handler.source == source) then
-			Utils.music.stopSound(handler.handlerID);
+		if (not handler.stopped) and (not soundID or soundID == "0" or handler.id == soundID) and (not channel or handler.channel == channel) and (not source or handler.source == source) then
+			if (handler.channel == "Music" and handler.handlerID == 0) then
+				StopMusic();
```
Ellypse: Shouldn't this be `Utils.music.stopMusic()`?  `StopMusic()` would only stop musics played using a path, while `Utils.music.stopMusic()`stops musics both played with a path or an ID.

Solanya: It is correct. The check makes sure that the music we're trying to stop at this moment is a path music, and only stops that.
Also `Utils.music.stopMusic()` calls back `stopSoundID()` so it would go on an infinite loop.

PR #141, opened 2018-06-30
Map changes for 8.0

Solanya:
Changes I've made for 8.0 : 

- From what I understand, GetCurrentMapAreaID meant either the map where the player is, or the map opened at the moment. I replaced the usages accordingly.
- GetPlayerMapPosition is in C_Map and requires a mapID.
- I will try to take a look into Map data providers later. In case I can't figure it out in time, the scan button on the map has been disabled with an appropriate error message.

### DIFF
--- a/totalRP3/modules/map/map.xml
+++ b/totalRP3/modules/map/map.xml
@@ -25,7 +25,7 @@
 		</Frames>
 		<Scripts>
 			<OnEnter>
-					WorldMapPOIFrame.allowBlobTooltip = false;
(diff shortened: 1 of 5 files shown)


Solanya:
Additional fix : 
- Setting residence in an instance only sets the text but doesn't set a map reference. Button to look at the residence was checking that the RC table had 4 arguments so it wasn't causing issues (since setting in instances would not give X/Y) but it seems cleaner to just not set that table.

PR #148, opened 2018-07-08
Fixed some alignment issues on the target frame

Solanya:
- Target frame caption + glance bar now have their position centered relative to the target frame. Unfortunately, to avoid weird positioning after the update, the best course of action was to reset the glance bar position in a flyway patch. (I could technically do math to avoid this, but the glance bar wasn't perfectly centered to begin with as far as I've seen...)

- When too many buttons on the target frame require to expand it, it will now always expand from the center.
**Technical shizzle-whizzle :** when moving the target frame around, it chooses the nearest point of the UIParent as the reference anchor and the same point on the target frame to save the relative position (for instance, on the left of the screen, it'll position the "LEFT" point of the target frame relative to the "LEFT" point of UIParent).
Previously, expanding would only change the width, which means if the target frame was left, it would expand right, and vice-versa. Top, bottom & center would expand from the center. Based on this, after changing the width, I offset the position of the target frame if it is left or right to compensate.
Yes maybe it would have been smarter to change the target frame so it would always use the "CENTER" point as the anchor. But I honestly had no clue how to do it so I went with the math.

- I saw that copy-paste job.

### DIFF
--- a/totalRP3/modules/flyway/flyway_patches.lua
+++ b/totalRP3/modules/flyway/flyway_patches.lua
@@ -103,3 +103,10 @@ TRP3_API.flyway.patches["8"] = function()
 		end
 	end
 end
+
+-- Reset glance bar coordinates
+TRP3_API.flyway.patches["9"] = function()
+	if TRP3_API.register.resetGlanceBar then
+		TRP3_API.register.resetGlanceBar();
+	end
+end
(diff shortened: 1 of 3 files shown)


Ellypse:
FYI: This is gonna way for post 8.0. I wanna test the impact of modifying the anchor for existing users. I will probably want the flyway patch to go further than just resetting the bar, but rather mutate the current location.

PR #151, opened 2018-07-12
Added event for opening a user's profile.

mukunda-:
Added two events, `PAGE_OPENED(pageId, pageContext)` and `REGISTER_PROFILE_OPENED(pageContext)`. The latter is a mere convenience that filters only listens for the page "player_main" from `PAGE_OPENED`.

Added a few fields to the register page context to preserve the unit ID when opening profiles with methods that support it.

I was going to have `REGISTER_PROFILE_OPENED` get its arguments from the context values, but passing the whole context structure seems more future compatible. The `source` field I added should also help with future compatibility.

### DIFF
--- a/totalRP3/core/impl/events.lua
+++ b/totalRP3/core/impl/events.lua
@@ -75,21 +75,6 @@ local Events = {
 	-- Called when as "About" page is shown.
 	-- This is used by the tooltip and the target bar to be refreshed
 	REGISTER_ABOUT_READ = "REGISTER_ABOUT_READ",
-
-	-- Called when the user opens a profile in the viewer.
-	-- Arg1 : Profile ID - may be nil when viewing the player's profile.
-	-- Arg2 : Context. This comes in two formats:
-	--         Player's profile format: (triggered when opening from the Character menu)
-	--           source   = "player"
-	--           isPlayer = true
-	--           profile  = player's profile data
-	--         Directory profile format: (triggered when opening someone else's profile)
-	--           source            = "directory".
-	--           profile,profileID = Profile data and ID.
-	--           openingWithUnitID = true if opening the page with a fresh unitID (name).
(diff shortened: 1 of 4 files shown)


mukunda-:
On second thought, maybe it's a good idea to make the REGISTER_PROFILE_OPENED more user friendly: don't pass the raw context, pass friendly arguments, and don't pass player-related events (i.e. keep it only to when you're viewing someone else, as is in the name "register" opened).

Thoughts?

Ellypse:
I actually don't think "REGISTER_PROFILE_OPENED

totalRP3/core/impl/events.lua
```diff
@@ -71,6 +76,21 @@ local Events = {
...
+	--           source   = "player"
+	--           isPlayer = true
+	--           profile  = player's profile data
+	--         Directory profile format: (triggered when opening someone else's profile)
+	--           source            = "directory".
+	--           profile,profileID = Profile data and ID.
+	--           openingWithUnitID = true if opening the page with a fresh unitID (name).
+	--           unitID            = unitID (name) used to open the page; may be nil or 
+	--                               the previously set value if openingWithUnitID=false.
+	REGISTER_PROFILE_OPENED = "REGISTER_PROFILE_OPENED",
```
Ellypse: Use only one event, `PAGE_OPENED` and use its parameters in the listener instead.

mukunda-:
The reason that I opted for the friendlier event, is because then in the listener you have clearly defined values passed in as arguments, and you don't have to access the `context` which could be considered an internal value that isn't defined super well. The system as a whole could use a *lot* more events that expose things in a user-friendly way, especially if you want to hide internals. I don't mind, but it's something to keep in mind.

Ellypse:
Oh crap, GitHub forgot to tell me you updated it. Sorry, had I knew I would have had included it in the 1.4.0 release.

Issue #160, opened 2018-07-18
Directory multi-selection doesn't deselect properly


Meorawr:
Open the directory and select a few profiles using those checkboxes on the left. Click the Actions button near the top and it'll confirm you've got _n_ profiles selected ("Action on _n_ selected profiles" appears).

Now uncheck them and click Actions again. [The UI still thinks they're checked](https://gyazo.com/ba86d9f8908af7d99f46633c9362331d).

Of further interest, if you select profiles and then delete them via this method they remain selected when you re-open the Actions menu; if you try to re-delete them you'll run into an unknown profile ID error in the process.

Only way to properly clear the selection is to open another page entirely.

PR #161, opened 2018-07-18
Fixed directory selection bugs (#160)

Solanya:
Selected profiles are detected by setting the value indexed to the profile ID to `true` in the array (which starts empty) when selecting it, and iterating on the array elements to make an action.

The first issue comes from the deselection setting the value to `false` instead of `nil`. The array still has a value for it, so considers it selected still (and the array size still takes it into account).

The second issue came from the fact the deletion of multiple selected profiles wasn't firing an event for every profile deleted, but only a general one which wouldn't update the selected profiles array.

**I think there might still be a potential issue if you select profiles then purge.** I'm not sure if purges should trigger the event per profile purged, although the event seems to be clearing some array in MSP if the profile was received through it. If not, at the very least the selected profiles array should be wiped after a purge. Let me know what to do about that.

### DIFF
--- a/totalRP3/modules/register/characters/register_ignore.lua
+++ b/totalRP3/modules/register/characters/register_ignore.lua
@@ -155,7 +155,7 @@ end
 
 function TRP3_API.register.unignoreID(unitID)
 	blackList[unitID] = nil;
-	Events.fireEvent(Events.REGISTER_DATA_UPDATED, unitID, hasProfile(unitID), nil);
+	Events.fireEvent(Events.REGISTER_DATA_UPDATED, unitID, TRP3_API.register.isUnitIDKnown(unitID) and hasProfile(unitID) or nil, nil);
 end
 
 function TRP3_API.register.getIgnoredList()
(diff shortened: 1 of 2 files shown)


Ellypse:
No! The single event being fired is intentional and was part of a fix back in the days! Firing an event for every single profile delete can crash the game when too many profiles are selected, as it causes the UI to be refreshed to reflect the deletion (sorting and rendering the list again, for every single deleted file).

The actual solution is to have the things that need to be refreshed listen to that one single event and do their stuff by looking at which profiles are left.

Solanya:
Hi I'm back with a vengeance.

Firing the `REGISTER_PROFILE_DELETED` without an argument now deselects everything just to be sure. The other option would be to send him the list of profiles deleted so he only deselects those, but it's unlikely you'll select profiles, purge profiles, then decide you wanted to delete/ignore the remaining selected profiles.

While I did that, I figured the "Purge ignored profiles" didn't have the events, so I added them. Also this purge was removing the character from the characters array, which in turn threw an error when trying to remove the character from the blacklist (because the directory forgot about that guy when asked for a list of linked profiles).

Issue #138, opened 2018-06-14
Add option to open RP profile when right-clicking on user name in chat


Ellypse:
We would like to add an "open profile" option in the menu when right-clicking a player name in the chat.

PR #172, opened 2018-07-22
[WIP] Open profile via right-click menu and/or ctrl-click in chat

Meorawr:
Very much an in-progress implementation of #138 that I'll fix up some more, think of this as an initial pass for some review I guess.

A few notes:

* As discussed on Discord there's some concern around taint if we modify the menus, so we'd need to remove bad entries like target and report AFK. The module itself right now is disabled by default.
* The ctrl-click stuff I lumped in with the chat links module because I ctrl-f'd `ChatFrame_OnShowHyperlink` and was like "aww heck yeah, it can go with that!" and then when writing the commit message I realised the chatframe module would be a better place.
* I've got some stub entries for Battle.net entry integration such that you could right-click a battle.net friend and get their current-character profile open too. Not implemented yet, so commented out.
* The opening of the profiles goes through the same code as the `/trp3 open` command, which I had to split out into a function. I shoved that into the `TRP3_API.slash` namespace for now because the roll stuff is there, so why not.

Not sure if we want ctrl-click to be an optional thing, nor if we actually want/need both of these solutions implemented. Ctrl-click probably should be behind an option flag if only because some chat addon somewhere is going to have implemented that to puke rainbows over someones' screen or something.


totalRP3/tools/Locale.lua
```diff
@@ -1062,6 +1062,16 @@ Putting your companion name in [brackets] will allow color and icon customizatio
 
 Importing content coming from a different version may cause issues in case of incompatibilities. Do you want to proceed anyway?]],
 
+	--*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
```
Ellypse: New locales should now be placed after the
```
--- PLACE LOCALIZATION NOT ALREADY UPLOADED TO CURSEFORGE HERE
```
message at [the end of the locale](https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/tools/Locale.lua#L1255), so I know what needs to be uploaded to Curse.

totalRP3/modules/register/main/register_exchange.lua
```diff
@@ -484,63 +484,66 @@ end
...
-TRP3_API.slash.registerCommand({
-	id = "open",
-	helpLine = " " .. loc.PR_SLASH_OPEN_HELP,
-	handler = function(...)
-		local args = {...};
 
-		if commandOpeningTimerHandle then
-			commandOpeningTimerHandle:Cancel();
-		end
+function TRP3_API.slash.openProfile(...)
```
Ellypse: As you said, this function should indeed be moved, maybe to the register_main file. Split the actual fetching and opening of the profile and indicate in the function's name that a request will be sent, so name it something like `TRP3_API.register.getAndOpenProfile(charID)` and only have the actual slash command code in this file.

Ellypse:
Pull request was reverted, file `OpenProfileAction` was not included and caused errors.

PR #187, opened 2018-07-30
Linebreak filtering in tooltip currently and OOC fields

Solanya:
Character limit does a good work but it is still possible to spam line breaks to make the tooltip overly high. This adds a filter with customizable limit on the number of max lines to allow in currently & OOC fields (5 by default).

Meorawr did all the stuff but I'm stealing the glory anyway.

April Fools code has been removed due to reaching the upvalue limit in the tooltip builder, which is an issue that will need to be resolved later.

### DIFF
--- a/totalRP3/tools/Locale.lua
+++ b/totalRP3/tools/Locale.lua
@@ -1266,7 +1266,7 @@ Please keep in mind that changing those settings might alter your experience wit
 	--- THEN MOVE IT UP ONCE IMPORTED
 	------------------------------------------------------------------------------------------------
 
-	CO_TOOLTIP_CURRENT_LINES = "Max \"current\" lines count",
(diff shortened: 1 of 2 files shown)


totalRP3/tools/Locale.lua
```diff
@@ -1266,6 +1266,7 @@ Please keep in mind that changing those settings might alter your experience wit
 	--- THEN MOVE IT UP ONCE IMPORTED
 	------------------------------------------------------------------------------------------------
 
+	CO_TOOLTIP_CURRENT_LINES = "Max \"current\" lines count",
```
Ellypse: The meaning is kind of obscure. Does it means it limits the number of lines in a long text that wraps over multiple lines?
It should probably be "Max \"current\" line breaks".

PR #189, opened 2018-07-30
Added new map module system

Ellypse:
A new map module system allows for creating map scanners (see the [PlayerMapScanner](https://github.com/Ellypse/Total-RP-3/tree/feature/new_map_module/totalRP3/modules/register/characters/PlayerMapScan) for implementation).

Map scanners must provide a `dataProviderTemplate` that is a Frame inheriting `BaseHighlightableMapPoiPinTemplate` and a custom Mixin (created using `AddOn_TotalRP3.MapPOIMixins.createPinTemplate([optionalTemplates])`. `MapPOIMixins` offers templates for automatically handling coalesced tooltips (displaying content for multiple points on top of each other inside on tooltip) and automatically animating the pin in a "boucing" effect.

- The pin Mixin is responsible for displaying itself using the data and providing tooltip information.
- The scanner is only responsible for sending, answering and receving scan data.


totalRP3/modules/map/MapDataProvider.lua
```diff
@@ -0,0 +1,62 @@
...
+function TRP3_MapDataProvider:RemoveAllData()
+	self:GetMap():RemoveAllPinsByTemplate(self.pinTemplate);
+end
+
+function TRP3_MapDataProvider:RefreshAllData(fromOnShow, ...)
+	self:RemoveAllData();
+
+	if not self.data then return end
+
+	for _, POIInfo in pairs(self.data) do
```
Meorawr: Casing-wise this should be "poiInfo"?

totalRP3/modules/map/MapScannersManager.lua
```diff
@@ -0,0 +1,101 @@
...
+--region Total RP 3 imports
+local Events = TRP3_API.Events;
+--endregion
+
+Events.MAP_SCAN_STARTED = "MAP_SCAN_STARTED"
+Events.MAP_SCAN_ENDED = "MAP_SCAN_ENDED"
+
+
+---@param scan MapScanner
+function MapScannersManager.registerScan(scan)
```
Meorawr: This'll be a contentious one; drop the superfluous "scan" suffix from the functions?

`MapScannersManager.register`, `MapScannersManager.getByID` and `MapScannersManager.launch`, etc. avoids the stutter of having "scan" mentioned twice in the name; the submodule itself already has a single responsibility so the stutter adds little clarity.

I'd go a step further and drop "Manager" too but there's existing code that follows that rule.

totalRP3/modules/map/WorldMapButton.lua
```diff
@@ -0,0 +1,165 @@
...
+	 :OnShow(function(tooltip)
+	tooltip:AddTempLine(WorldMapButton.subtitle)
+end)
+--endregion
+
+WorldMapButton:SetScript("OnClick", function(self)
+	local structure = {};
+	for scanID, scan in pairs(TRP3_API.MapScannersManager.getAllScans()) do
+		if scan:CanScan() then
+			insert(structure, { icon(scan.scanIcon, 20) .. " " .. (scan.scanOptionText or scanID), scanID});
```
Meorawr: Space after the `{` but none before the following `}`. The next occurrence has no surrounding spaces at all. This triggers me immensely.

totalRP3/modules/map/MapScannersManager.lua
```diff
@@ -0,0 +1,101 @@
...
+	end
+
+	promise:Success(function()
+		-- If the displayed map changed
+		if displayedMapID ~= AddOn_TotalRP3.Map.getDisplayedMapID() then
+			return
+		end
+		scan:OnScanCompleted();
+		TRP3_API.MapDataProvider:OnScan(scan:GetData(), scan:GetDataProviderTemplate())
+		TRP3_API.ui.misc.playSoundKit(43493);
```
Meorawr: Use the string name for the sound kit (Spell_Vehicle_Speedup?) for a bit more clarity perhaps? Both options seem pretty bad though. Also protects against the (hopefully never) case that sound kit IDs get moved.

totalRP3/modules/map/MapPOIMixins.lua
```diff
@@ -0,0 +1,184 @@
...
+	local nameA = a.sortName or "";
+	local nameB = b.sortName or "";
+
+	return (categoryA > categoryB)
+		or (categoryA == categoryB and nameA < nameB);
+end
+
+function GroupedCoalescedMapPinMixin:OnMouseEnter()
+
+	local tooltip = Ellyb.Tooltips.getTooltip(self);
```
Meorawr: `Ellyb.Tables` and `Ellyb.Tooltips` are referenced a bunch in this file and could probably be upvalued?

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -0,0 +1,110 @@
...
+-- Expose template name, so the scan can use it for the MapDataProvider
+TRP3_PlayerMapPinMixin.TEMPLATE_NAME = "TRP3_PlayerMapPinTemplate";
+
+--- This is called when the data provider acquire a pin, to transform poiInfo received from the scan
+--- into display info to be used to decorate the pin.
+function TRP3_PlayerMapPinMixin:GetDisplayDataFromPOIInfo(poiInfo)
+	local characterID = poiInfo.sender;
+	local displayData = {};
+
+	if TRP3_API.register.isUnitIDKnown(characterID) and TRP3_API.register.hasProfile(characterID) then
```
Meorawr: Invert this check and bring the `else` block up here and do an early return inside of it, then 90% of the function won't need to be indented.

totalRP3/modules/map/MapDataProvider.lua
Ellypse: Hmm, I guess, but I don't like it. It's debatable, but POI being an acronym I like it better as uppercase.

totalRP3/modules/map/MapScannersManager.lua
Ellypse: I agree with you and I like that change :)

totalRP3/modules/map/MapScannersManager.lua
Ellypse: That's bad copy pasta, it should use `SOUNDKIT` instead.

totalRP3/modules/map/MapDataProvider.lua
Ellypse: After more reflexion, I think it's better to go with your suggestion as it follows the naming convention of Blizzard, which is something I try to do whenever possible.

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -0,0 +1,121 @@
...
+		profile = TRP3_API.profile.getPlayerCurrentProfile().player;
+		displayData.iconAtlas = "PlayerPartyBlip";
+		displayData.iconColor = Ellyb.ColorManager.CYAN;
+		displayData.categoryName = loc.REG_RELATION .. ": " .. Ellyb.ColorManager.CYAN("SELF");
+		displayData.categoryPriority = -huge;
+	else
+		profile = TRP3_API.register.getUnitIDCurrentProfile(characterID);
+
+		--region Player relationship
+		local relation = TRP3_API.register.relation.getRelation(TRP3_API.register.getUnitIDProfileID(characterID));
```
Solanya: Should probably be `profile` inside `getRelation`

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -0,0 +1,121 @@
...
+		if relation ~= TRP3_API.register.relation.NONE then
+			local relationShipColor = TRP3_API.register.relation.getColor(relation);
```
Solanya: Nitpicking but why the uppercase S ? 🤔

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -0,0 +1,121 @@
...
+
+			-- Store the relationship on the marker itself as the category.
+			displayData.categoryName = loc.REG_RELATION .. ": " .. relationShipColor(loc:GetText("REG_RELATION_".. relation));
+			displayData.categoryPriority = TRP3_API.globals.RELATIONS_ORDER[relation] or -huge;
+		end
+		--endregion
+	end
+
+	--region Player name
+	displayData.playerName = TRP3_API.register.getCompleteName(profile.characteristics, characterID, true);
```
Solanya: You shouldn't send characterID without removing realm from it. If profile.characteristics or profile.characteristics.FN is nil, you'll have the realm in displayData.playerName

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapScanner.lua
```diff
@@ -0,0 +1,126 @@
...
+local loc = TRP3_API.loc;
+local broadcast = AddOn_TotalRP3.Communications.broadcast;
+local registerConfigKey = TRP3_API.configuration.registerConfigKey;
+local getConfigValue = TRP3_API.configuration.getValue;
+local get = TRP3_API.profile.getData;
+local Map = AddOn_TotalRP3.Map;
+--endregion
+
+local CONFIG_ENABLE_MAP_LOCATION = "register_map_location";
+local CONFIG_DISABLE_MAP_LOCATION_ON_OOC = "register_map_location_ooc";
```
Solanya: Where did `CONFIG_DISABLE_MAP_LOCATION_ON_PVP` go ? 🤔

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapScanner.lua
```diff
@@ -0,0 +1,126 @@
...
+	end
+	return false;
+end
+--endregion
+
+--region Broadcast commands
+broadcast.registerCommand(SCAN_COMMAND, function(sender, mapID)
+	if Map.playerCanSeeTarget(sender) then
+		mapID = tonumber(mapID);
+		if shouldAnswerToLocationRequest() and Map.playerCanSeeTarget(sender) then
```
Solanya: That's the second time you test `Map.playerCanSeeTarget(sender)`

totalRP3/modules/register/characters/register_main.lua
```diff
@@ -57,6 +57,8 @@ local tcopy = tcopy;
 local type = type;
 local showAlertPopup = TRP3_API.popup.showAlertPopup;
 
+local Map = AddOn_TotalRP3.Map;
```
Solanya: Unused import ?

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
Ellypse: 🤗 🚢

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapScanner.lua
```diff
@@ -0,0 +1,126 @@
...
+--region Scan behavior
+function playerMapScanner:Scan()
+	broadcast.broadcast(SCAN_COMMAND, Map.getDisplayedMapID());
+end
+
+-- Players can only scan for other players in zones where it is possible to retrieve player coordinates.
+function playerMapScanner:CanScan()
+	if getConfigValue(CONFIG_ENABLE_MAP_LOCATION) then
+		local x, y = Map.getPlayerCoordinates()
+		if x and y then
```
Solanya: That condition was in the previous version of the scan and I'm not sure it's actually warranted. There's no reason you can't scan Stormwind while you're in a raid or such, you never send your own coordinates while scanning.

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -0,0 +1,121 @@
...
+			displayData.playerName = TRP3_API.utils.str.icon(profile.characteristics.IC, 15) .. " " .. displayData.playerName;
+		end
+	end
+	--endregion
+
+	return displayData
+end
+
+--- This is called by the data provider to decorate the pin after the base pin mixin has done its job.
+---@param poiInfo {position:Vector2DMixin, sender:string}
```
Solanya: Parameter is displayData, not poiInfo, the content of it is probably incorrect as well.

Solanya: That param is actually the one from the previous function `GetDisplayDataFromPoiInfo`, I would guess you separated this part of the function later and didn't move the doc part.

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapPin.lua
Ellypse: I'm not sure about this comment. `TRP3_API.register.relation.getRelation(profileID)` expect a profileID to check for the current player relation against that profile.

See: https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/modules/register/characters/register_ignore.lua#L67

totalRP3/modules/register/characters/PlayerMapScan/PlayerMapScanner.lua
Ellypse: Yes. We check on the sender and receiver side, just to be extra sure. Checking before sending avoid answering uselessly, and checking when receiving is to make sure we don't have invalid data received.

Ellypse: Wait no, I'm dumb. Sorry.

PR #194, opened 2018-08-02
MSP to TRP profile fixes

Solanya:
- Main reason for this PR : The AE field received from a MSP profile was stored in the AE field in TRP's characteristics page, except the eye color field is EC in TRP, so that was never displayed.

- For the second commit, I've peered into the future and MRP will send name, class and eye color with color tags sometimes. While the name color was escaped and stored into the proper field, class and eye color weren't. I suggest adding that preemptively. As of now, MRP class color will work the same way as TRP (same as name).


Ellypse:
This PR is no longer relevant following Itarater's work

PR #206, opened 2018-08-20
sendObject function now handles both Channel type and numbered channels

Connor-Macleod:
This allows DM add-ons to confirm that the user is not using the wrong modifier when the result is within the expected range
This will be ignored in all other cases without impact beyond message size

### DIFF
--- a/totalRP3/core/impl/CommunicationProtocol.lua
+++ b/totalRP3/core/impl/CommunicationProtocol.lua
@@ -60,6 +60,9 @@ local PRIORITIES = {
 	HIGH = "HIGH",
 }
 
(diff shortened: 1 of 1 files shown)


totalRP3/core/impl/CommunicationProtocol.lua
```diff
@@ -59,6 +60,9 @@ local PRIORITIES = {
 	HIGH = "HIGH",
 }
 
+local validChannels = {"PARTY", "RAID", "GUILD", "BATTLEGROUND", "WHISPER", "CHANNEL"};
```
Ellypse: "Constants" should be declared using UPPERCASE_SNAKE_CASE so we kind of know we should touch them

PR #207, opened 2018-08-22
Made sendObject method accepts table of params

Ellypse:


Ellypse:
**NOTE:** After some discussion on Discord, it has been decided that instead of Frankensteining the existing function with some type checking and if statements, we should provide a separate function in the Communications module that will handle broadcasting object separately (with some shared code being refactored).
We will have to go through existing usages to update to the new function.

Issue #124, opened 2018-05-16
Slash commands system should resolve macro modifiers


Ellypse:
For macro scripting, the slash command system in Total RP 3 should resolve modifiers like `[mounted]` when executing a command. This would make the following command possible:
```
/trp3 [mounted] profile Flappy bird; [mod:alt] profile Happy cat; [btn:2] profile Moonmoon; profile Belyelle
```

Ellypse:
Depending on the regex skills of the person doing this ticket, it might be better to resolve the string constructed and then pass the command, allowing the modifiers to be used for arguments too (have the conditional part after the `profile` command here).

PR #208, opened 2018-08-22
Macro conditionals for /trp commands

Connor-Macleod:
All vanilla conditonals are supported by SecureCmdOptionParse
Additional commands are supported and can be added by the use of TRP3_API.slash.registerConditional
Added commands also get the "no" inversion of regular macros.
Additional comands included by default are:
[ic][ooc] to test for IC/OOC status
[location:locationName][loc:locationName] to test if the player is in a specific location (as reported by the minimap)
[profile:profileName] to test for a specific selected profile. Profile names are sanitised of all punctuation and converted into camelcase. As such the profile "bob's profile" and "bobs profile" would both pass the same check of "bobsProfile"

Addresses ticket https://github.com/Ellypse/Total-RP-3/issues/124

### DIFF
--- a/totalRP3/core/impl/slash.lua
+++ b/totalRP3/core/impl/slash.lua
@@ -100,30 +100,29 @@ local function parseCommandTable(args)
 end
 
 local function testConditions(conditionals)
-	local conditionalsPassed = false;
-	for _, condition in pairs(conditionals) do
-		--cleanup the square brackets to give us a comma seperated lsit of tests
-		local passed = true;
-		for test in string.gmatch(condition, "[^, ]+") do
-			if test ~= "" then
-				local args = {strsplit(":", test)};
-				if CONDITIONALS[args[1]] then
-					if not CONDITIONALS[args[1]].test(unpack(args)) then
+	for _, condition in pairs({strsplit("[]", conditionals)}) do
+		if condition ~= "" then
+			local passed = true;
+			for test in string.gmatch(condition, "[^,]+") do
+				if test ~= "" then
+					local args = {strsplit(":", test)};
+					if CONDITIONALS[args[1]] then
+						if not CONDITIONALS[args[1]].test(unpack(args)) then
+							passed = false;
+							break;
+						end
+					elseif SecureCmdOptionParse("["..test.."] true; false")=="false" then
 						passed = false;
 						break;
 					end
-				elseif SecureCmdOptionParse("["..test.."] true; false")=="false" then
-					passed = false;
-					break;
 				end
 			end
-		end
-		if passed then
-			conditionalsPassed = true
-			break;
+			if passed then
+				return true;
+			end
 		end
 	end
-	return conditionalsPassed;
+	return false;
 end
 
 function SlashCmdList.TOTALRP3(msg, editbox)
@@ -133,10 +132,8 @@ function SlashCmdList.TOTALRP3(msg, editbox)
 		return parseCommandTable(args);
 	end
 
-	for segment in string.gmatch(msg, "(%b[] ?[^%;%[%]]*)") do
-		local segmentConditionals = {string.match(segment, "%[([^%;%[%]]*)%] ?([^%;%[%]]*)")};
-		local segmentCommand = segmentConditionals[#segmentConditionals];
-		table.remove(segmentConditionals);
+	for segment in string.gmatch(msg, "(%[[^;]*%] ?[^%;%[%]]*)") do
+		local segmentConditionals, segmentCommand = string.match(segment, "(%[[^%;]*%]) ?([^%;%[%]]*)");
 		if testConditions(segmentConditionals) then
 			return parseCommandTable({strsplit(" ", segmentCommand)})
 		end
--- a/totalRP3/core/impl/slash.lua
+++ b/totalRP3/core/impl/slash.lua
@@ -275,6 +275,7 @@ TRP3_API.events.listenToEvent(TRP3_API.events.WORKFLOW_ON_LOADED, function()
 			profileName = string.lower(profileName);--we lowercase it all
 			profileName = string.gsub(profileName, "[^%s%a]", "");--remove anything that's not a letter or space)
 			profileName = string.gsub(profileName,"(%s)(%a)", function(a,b) return string.upper(b) end) -- and turn it into a camelCaseValue
+			profileName = string.gsub(profileName, "%s", ""); --and clean up any remaining spaces
 			return profileName==targetProfile; --then compare it to the given string
 		end
 	})
--- a/totalRP3/core/impl/slash.lua
+++ b/totalRP3/core/impl/slash.lua
@@ -100,7 +100,6 @@ local function parseCommandTable(args)
 end
 
 local function testConditions(conditionals)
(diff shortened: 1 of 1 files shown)


totalRP3/core/impl/slash.lua
```diff
@@ -34,6 +34,7 @@ TRP3_API.slash = {}
 --*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
 local COMMANDS = {};
+local CONDITIONALS = {};
```
Meorawr: It has `CAPITAL_CASING` as if it's a constant, but with functions like `registerConditional` this isn't really a constant.

totalRP3/core/impl/slash.lua
```diff
@@ -46,15 +47,35 @@ function TRP3_API.slash.unregisterCommand(commandID)
 	COMMANDS[commandID] = nil;
 end
 
+
+function TRP3_API.slash.registerConditional(conditionalStructure)
```
Meorawr: I imagine the idea going forward is to actually dump API stuff in the AddOn_TotalRP3 table somewhere. Unless this is supposed to be internal-API only, though if it is I'd do away with register/unregister and actually make the conditionals table constant.

totalRP3/core/impl/slash.lua
```diff
@@ -46,15 +47,35 @@ function TRP3_API.slash.unregisterCommand(commandID)
...
+	--we automatically create the inverse as well
+	CONDITIONALS["no"..conditionalStructure.id] = {
+		id = "no"..conditionalStructure.id,
+		test = function(...)
+			return not conditionalStructure.test(...)
+		end
+	}
+end
+
+function TRP3_API.slash.unregisterConditional(conditionalID)
```
Meorawr: This'll be nitpicky and probably can be ignored, but I'm not seeing much of a use case for where someone would want to unregister a conditional.

From a user's perspective it'd be really weird if their macro suddenly stopped working because some addon unregistered itself.

totalRP3/core/impl/slash.lua
Meorawr: Also, on the one hand the parameter being a table matches the API convention used by `registerCommand` but I'm wondering if it could be clearer if instead there was a `MacroConditional` class (using Ellyb's class stuff), such that API usage would be more akin to:

```
local condition = AddOn_TotalRP3.slash.MacroConditional("mounted", function(...) --[[ test ]] end);
AddOn_TotalRP3.slash.registerConditional(condition);
```

The aim being to make a bit clearer for users of the API; if they know it takes a specific class since they can look up the possible arguments for the constructor vs. looking at the implementation of this function to determine what magic keys are available.

totalRP3/core/impl/slash.lua
```diff
@@ -78,6 +99,58 @@ function SlashCmdList.TOTALRP3(msg, editbox)
 	end
 end
 
+local function testConditions(conditionals)
+	logger:Info(Ellyb.Tables.toString(conditionals));
+	local conditionalsPassed = false;
+	for _, condition in pairs(conditionals) do
+		--cleanup the square brackets to give us a comma seperated lsit of tests
```
Meorawr: "lsit"? :pedantic:

totalRP3/core/impl/slash.lua
```diff
@@ -167,6 +240,48 @@ end
...
+		locationString = string.lower(locationString);--we lowercase it all
+		locationString = string.gsub(locationString, "[^%s%a]", "");--remove any apostrophies (or anything else that's not a letter)
+		locationString = string.gsub(locationString,"(%s)(%a)", function(a,b) return string.upper(b) end) -- and turn it into a camelCaseValue
+		return targetLocation==locationString; --then compare it to the given string
+	end
+	TRP3_API.slash.registerConditional({
+		id = "loc",
+		test = location
+	})
+	TRP3_API.slash.registerConditional({
```
Meorawr: I'd favour having one condition rather than two for location, ignoring the fact that `[stance]`, `[form]`, etc. are a mess.

Personally I'd favour `[location:goldshire]` over `loc` for a condition. Stands out more when you read it.

totalRP3/core/impl/slash.lua
```diff
@@ -78,6 +99,58 @@ function SlashCmdList.TOTALRP3(msg, editbox)
...
+					end
+				elseif SecureCmdOptionParse("["..test.."] true; false")=="false" then
+					passed = false;
+					break;
+				end
+			end
+		end
+		if passed then
+			conditionalsPassed = true
+			break;
```
Meorawr: At this point you might as well just return rather than breaking-then-returning? Also means you can kill `conditionalsPassed` entirely and just `return true` here and `return false` at the tail of the function.

totalRP3/core/impl/slash.lua
```diff
@@ -78,6 +99,58 @@ function SlashCmdList.TOTALRP3(msg, editbox)
 	end
 end
 
+local function testConditions(conditionals)
+	logger:Info(Ellyb.Tables.toString(conditionals));
+	local conditionalsPassed = false;
+	for _, condition in pairs(conditionals) do
```
Meorawr: Is `testConditions` called with an _ordered_ table of conditionals? If I'm reading rightly due to `pairs` here things wouldn't work so well if I had a bunch of conditionals in a macro since they'd potentially get iterated in an indeterminate order.

Meorawr:
Ugh, expand the ones that said they're "outdated". I swear I reviewed against the latest but apparently not.

totalRP3/core/impl/slash.lua
Connor-Macleod: This is consistent with other similar variables.

totalRP3/core/impl/slash.lua
Connor-Macleod: This was added under the logic of:
1. it's three lines of code
2. if you can remove slash commands you should be able to remove slash conditionals.

totalRP3/core/impl/slash.lua
Connor-Macleod: Location is the prefered form. However, thanks to the character limit in conditionals, it's preferable to have a short form of long conditions. I.e. location can be shortened to loc. Blizzard works under the same theory with things like modifier -> mod

totalRP3/core/impl/slash.lua
Ellypse: I'm okay with it, and as was said on Discord (I think) it might be used in advanced use cases in Extended.

totalRP3/core/impl/slash.lua
Ellypse: I'm not sure about this, but for future public APIs going in `AddOn_TotalRP3` a documentation of the table would be provided so that people don't have to go look at the implementation.

In general, I've been going with OOP for the new system I create for more clarity and extensibility, but I'm okay with have things that are piggybacking on existing systems to not go with classes. If Connor feels fancy he can make that a proper class, but I won't require it here.

Ellypse:
If no changes are made to this pull request by the end of the month I'm closing it 🙂

Ellypse:
Closing for lack of activity, as mentioned before.

PR #209, opened 2018-08-23
MSP rewrite fixes

Itarater:
These commits fix a few lingering issues and corner cases from the MSP rewrites merged in #196.

### DIFF
--- a/totalRP3/modules/register/msp/register_msp.lua
+++ b/totalRP3/modules/register/msp/register_msp.lua
@@ -310,8 +310,8 @@ local function onStart()
 						end
 						-- Remove title from full name, if present
 						if field == "NA" and value and data.PX then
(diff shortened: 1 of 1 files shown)


totalRP3/modules/register/msp/register_msp.lua
```diff
@@ -150,7 +150,7 @@ local function onStart()
 		for i = 1, 5 do
 			local peek = dataTab.PE[tostring(i)];
 			if peek and peek.AC then
-				if i > 1 then
+				if #peeks > 1 then
```
Solanya: Should be > 0

Itarater: Rebased with this fixed. I actually made that error multiple times, including when we were talking about it in Discord.

PR #215, opened 2018-09-02
Fixed issue with glance frame reset + removed useless setting

Solanya:
So, some weird stuff going on. It turns out `TRP3_API.register.resetGlanceBar` was only updating the config settings and not the actual frame, so the right position wouldn't be seen until reload.
The button to reset was calling the local `resetPosition` instead, which would update the frame.
And in the config settings, there was a button to use Total RP 3 style positions, just below the reset button, doing exactly the same thing as the reset button, not even using `resetPosition` but a local function with the exact same code.

I added the frame update to `TRP3_API.register.resetGlanceBar`, removed `resetPosition` and replaced the function called by the reset button, and removed the button to use TRP3 style position. As an added bonus, just to be sure, I wait for the loading workflow to finish before resetting the glance bar. Not sure it's needed but doesn't hurt.

### DIFF
--- a/totalRP3/modules/flyway/flyway_patches.lua
+++ b/totalRP3/modules/flyway/flyway_patches.lua
@@ -106,7 +106,9 @@ end
 
 -- Reset glance bar coordinates
 TRP3_API.flyway.patches["9"] = function()
-	if TRP3_API.register.resetGlanceBar then
-		TRP3_API.register.resetGlanceBar();
-	end
+	TRP3_API.events.listenToEvent(TRP3_API.events.WORKFLOW_ON_FINISH, function()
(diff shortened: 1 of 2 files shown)


Ellypse:
What's the logic behind removing the TRP3 style? The at-first-glance bar can be placed anywhere in the UI by the user, not just the TRP3 target frame. This behaviour should be preserved.

Solanya:
It's not the TRP2 style, it's the TRP3 style. It just happens that the text on the button is the same because it's something like "Use". Everything else is marked TRP3.

Ellypse:
Alright, my bad. Sounds good, make sense.

PR #261, opened 2018-12-17
#259 - Removed references to guild chat support

### LINKED ISSUE #259: Remove all references and usage of guild chat
Guild chat customizations are gone, we should remove them.

Ellypse:
Fixes #259


Solanya:
You can pull back those changes, Blizzard has hotfixed guild channels to enable customization once more (tested for both chat links and emote/OOC detection)

PR #267, opened 2019-02-01
Project organization improvements

Ellypse:
- Added stricter Luacheck rules to the project
- Cleaned existing issues raised by Luacheck
- Updated Ellyb to the newest version and updated usage of all assertions for the new system.
- Updated copyright notice in all files to be more consistent.
- Removed all documents, localizator and zTest folders.
- Updated and enforce line ending in the git repository.
- Travis CI will now send alerts on Discord if a commit doesn't pass Luacheck.


Meorawr:
How about doing away with the per-file copyright notices entirely?

While the Apache license says you _should_ include the preamble in each source file but it's not a requirement.

http://www.apache.org/dev/apply-license.html#new

> Each original source document (code and documentation, but excluding the LICENSE and NOTICE files) SHOULD include a short license header at the top. 

"SHOULD" links to their [headers page](http://www.apache.org/legal/src-headers.html#headers) which then has [the following FAQ entry](http://www.apache.org/legal/src-headers.html#faq-whyheader).

> Why is a licensing header necessary?
> 
> License headers allow someone examining the file to know the terms for the work, even when it is distributed without the rest of the distribution. Without a licensing notice, it must be assumed that the author has reserved all rights, including the right to copy, modify, and redistribute.

So by not having the header present the worst case is the source file becomes more protected (and if really desired you could have a small one-line comment at the root of each file saying "This file is covered by the license terms present in the LICENSE file at the root of the project.").

Benefits:

1. I bloody hate long-ass preambles at the top of each file that serve no real purpose.
2. Everyone else sane in the world also does, right?
3. You don't need to update that copyright year each time in the future.
4. No one needs to open up another file and copy/paste the preamble into each new file because dear god that nonsense is irritating.

Cons:

1. Literally none.

Ellypse:
I hesitated, I was about to remove the headers, but it says "SHOULD" in all uppercase, so I decided to comply 🙂

Issue #270, opened 2019-02-08
Send profile music theme via LibMSP


Ellypse:
As per described in @Katoriehooves proposal https://github.com/Ellypse/LibMSP/issues/20, we should expose the profile music theme of the player via LibMSP in the `MU field`.

**Given** the user selected a music theme for their profile
**When** the profile is sent via LibMSP
**Then** Total RP 3 should provide a new `MU` field with the path of the music without the `Sound\Music\` part of the path.

> **Technical notes:** Total RP 3 might move to using file IDs to pick musics in the future. We would then need to use the [`GetFileIDFromPath`](https://wow.gamepedia.com/API_GetFileIDFromPath) API function to make sure we send a file path to LibMSP, or coordinate with other developers to also support file IDs.

Issue #269, opened 2019-02-08
Add relationship status to Total RP 3 profiles


Ellypse:
As per described in @Katoriehooves' proposal https://github.com/Ellypse/LibMSP/issues/18, MyRolePlay is now exposing relationship status for the player via LibMSP.

We should add this new fiel to Total RP 3

**Given** a Total PR 3 user
**When** they are editing their profile
**Then** a dropdown should be displayed in their profile, in the characteristics section, with the main informations fields, to let them pick their relationship status.

**Given** a relationship status has been selected by the user
**When** the user profile is sent via Total RP 3 communications
**Then** the relationship status should be sent as a numeric value alongside the miscellaneous informations of the profile.

**Given** a relationship status has been selected by the user
**When** the user profile is sent via LibMSP
**Then** the relationship status should be included in the `RS` as a numeric value.

**Given** the user is opening a profile that has relationship informations
**When** the miscellaneous tab is being displayed
**Then** the relationship status should be displayed with the other flags as a localised string value that corresponds to the numeric flag stored in the profile.

Here are the values as per the LibMSP proposal:
- `0` - Unknown / Prefer not to say
- `1` - Single
- `2` - Taken
- `3` - Married
- `4` - Divorced
- `5` - Widowed

Issue #268, opened 2019-02-08
Add trial account information to MSP fields


Ellypse:
As per described in Katorie's proposal https://github.com/Ellypse/LibMSP/issues/19, we should expose the trial account flag for the player in the `TR` MSP field.

**Given** the game is giving valid data for the status of the player account
**When** the profile is sent via LibMSP
**Then** Total RP 3 should provide a `TR` field with the values:
- `0` for a regular account
- `1` for a trial account
- `2` for a veteran account

PR #272, opened 2019-02-10
MSP fields updates

Solanya:
Issues #268 #269 #270 (why did you do this to me)

- Trial global changed to an integer (0, 1 or 2) and flag appears on tooltip when global is not 0. Global sent through MSP
- Music sent through MSP
- Addition of a titled dropdown template
- Addition of relationship status in the characteristics tab, sent through MSP
- Fixed a typo for unknown profiles


totalRP3/modules/register/characters/register_characteristics.lua
```diff
@@ -312,7 +316,12 @@ local function setConsultDisplay(context)
 	local shownCharacteristics = {};
 	local shownValues = {};
 	for _, attribute in pairs(CHAR_KEYS) do
-		if strtrim(dataTab[attribute] or ""):len() > 0 then
+		if attribute == "RS" then
+			if dataTab[attribute] and dataTab[attribute] > 0 then
+				tinsert(shownCharacteristics, attribute);
+				shownValues[attribute] = RELATIONSHIP_STATUS_DROPDOWN[dataTab[attribute] + 1][1];
```
Ellypse: This line is quite confusing. If it's possible to break it down in a more readable code it would make it easier to maintain in the future.
Maybe sprinkle some checks in case the value of `dataTab[attribute]` is not listed in `RELATIONSHIP_STATUS_DROPDOWN` (profiles coming from MSP), as you might be trying to index a `nil` value.

totalRP3/modules/register/characters/register_characteristics.lua
```diff
@@ -1523,6 +1548,8 @@ function TRP3_API.register.inits.characteristicsInit()
 	TRP3_RegisterCharact_Edit_ResidenceFieldText:SetText(loc.REG_PLAYER_RESIDENCE);
 	TRP3_RegisterCharact_Edit_BirthplaceFieldText:SetText(loc.REG_PLAYER_BIRTHPLACE);
 
+	setupListBox(TRP3_RegisterCharact_Dropdown_RelationshipField, RELATIONSHIP_STATUS_DROPDOWN, onRelationshipStatusSelection, loc.REG_PLAYER_RELATIONSHIP_STATUS_UNKNOWN, 200, false);
+	TRP3_RegisterCharact_Dropdown_RelationshipFieldTitle:SetText(loc.REG_PLAYER_RELATIONSHIP_STATUS);
```
Ellypse: It's minor, but the template you have implemented in XML has some code for displaying a tooltip. If you have any idea of a tooltip text to explain to the user how to use this field you could add it here (either using the old TRP3 way or Ellyb's Tooltips module).

You could also go for a description of each of the relationship statuses by passing a third field to your `RELATIONSHIP_STATUS_DROPDOWN` table above that would be displayed in a tooltip when the user goes to select an option, if you feel fancy.

Solanya: I don't feel fancy.

PR #277, opened 2019-03-10
Export Psychological Traits via MSP

Meorawr:
Purpose of this field is to allow her to add in some pretty displays for traits on MRP's end. While we do have that psychological section in template 3 present, I don't think it's reasonable for us to ever come up with a human-readable text format for traits that remains extensible, especially given the features available to custom traits.

As it's a new field I designed it around the limitation that it'd be machine-readable, but it's also somewhat debuggable by having a rather simple `[trait key="value" key2="value2"]` format.

Additionally I took the opportunity to try and make MSP fields testable by adding a basic framework for registering/serializing/deserializing them. I didn't really want to add another ~100 lines to the middle of the existing MSP code to further confuse things there.

A few things need to be resolved:

1. Kat needs to implement her end of things :)
2. We need to decide on a mapping of ID numbers for the "built-in" or "standard" traits (like Lustful/Chaste). Could use TRPs ID numbers but a better idea might come along (string IDs? Easier to recognise but longer to send).
3. The `key="value"` pairs at the moment need `]` and `"` characters stripped from the values. Current impl does the latter only because I forgot about the square brackets. Adding support for escaping quotes would be a lot of work on the reading end, so I'm against that - and it's not like anyone is really going to use them in a trait.
4. More protection is needed when receiving a trait that has no icon/color information possibly.
5. If there's no color/icon/custom name/ID field we might send some weird stuff out. Should probably protect against that somehow.

### DIFF
--- a/totalRP3/modules/register/main/register_exchange.lua
+++ b/totalRP3/modules/register/main/register_exchange.lua
@@ -514,8 +514,7 @@ TRP3_API.slash.registerCommand({
 		end
 
 		sendQuery(characterToOpen);
-		local request = {"TT", "HH", "AG", "AE", "HB", "DE", "HI", "AH", "AW", "MO", "NH", "IC", "CO"};
-		msp:Request(characterToOpen, request);
+		msp:Request(characterToOpen, AddOn_TotalRP3.MSP.REQUEST_FIELDS);
 		-- If we already have a profile for that user in the registry, we open it and reset the name (so it doesn't try to open again afterwards)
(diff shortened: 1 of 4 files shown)


Meorawr:
Ah crap, I based it off the template 3 HTML branch... I'll rebase it at some point.

Meorawr:
While I'm not particularly enamored with the current implementation, it works as a MVP and delivers the feature. I'd rather move the stuff to a common library at some point rather than faff around with a TRP-specific implementation.

Consider it a "good enough for now" implementation that we'll split out later.

Meorawr:
Force pushed to rebase on development as I mentioned earlier (had previously based off the HTML branch by accident).

PR #278, opened 2019-03-10
Swap to a HTML display for template 3 descriptions

Meorawr:
Draft PR until we get confirmation that making template 3 display in a HTML frame isn't going to murder the baby jesus.

Swapping things was simple until I learned that resizing the frame invoked the wrath of satan.


Meorawr:
One potential issue with a HTML view is that ASCII art might be broken in profiles.

Someone had a lovely heart that displayed like a heart in MRP, but in TRP it was a bit deflated.

We might want to call an ambulance for their heart.

Meorawr:
It was decided that there's no reason to not implement this. Support won't be extended to T2 profiles because they're not used for cross-addon interop.

Solanya:
I see no reason not to add HTML to Template 2 though. Might as well do it for every template, so people don't ask for it down the line.

PR #280, opened 2019-03-11
Fix map scans

Meorawr:
Your linting god can't save you now.

A few fully qualified field names never had upvalues created so tooltips for map scans were utterly buggered.

Rather than upvalue them I've just made the references explicit, because excessive upvalues are an over-optimisation in this case.

### DIFF
--- a/totalRP3/modules/map/MapPoiMixins.lua
+++ b/totalRP3/modules/map/MapPoiMixins.lua
@@ -60,9 +60,9 @@ local function sortMarkerEntries(a, b)
 	local categoryA = a.categoryPriority or math.huge;
 	local categoryB = b.categoryPriority or math.huge;
 
-	if categoryA == huge and type(categoryB) == "string" then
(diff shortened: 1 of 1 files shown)


totalRP3/modules/map/MapPoiMixins.lua
```diff
@@ -57,8 +57,8 @@ local TOOLTIP_CATEGORY_SEPARATOR = [[|TInterface\Common\UI-TooltipDivider-Transp
 ---  in order of their category priority (descending), or if equal, their
 ---  sortable name equivalent (ascending).
 local function sortMarkerEntries(a, b)
-	local categoryA = a.categoryPriority or huge;
-	local categoryB = b.categoryPriority or huge;
+	local categoryA = a.categoryPriority or math.huge;
+	local categoryB = b.categoryPriority or math.huge;
 
 	if categoryA == huge and type(categoryB) == "string" then
```
Solanya: I imagine it should be `math.huge` here and two lines below as well.

PR #287, opened 2019-03-17
Don't use localized section headers in DE field

Meorawr:
Other addons rely on these being in plain English so they can operate on them for changing how they get displayed on their end.

This PR is brought to you by Katorie. She screamed.


Katoriehooves:
I saw Spanish. ;_;

Meorawr:
We all saw Spanish. /pat

PR #288, opened 2019-03-18
Fix passing of nil events through the dispatcher

Meorawr:
Dunno what causes it but let's fix it and figure things out later. ¯\\_(ツ)\_/¯

### DIFF
--- a/totalRP3/core/impl/CommunicationProtocol.lua
+++ b/totalRP3/core/impl/CommunicationProtocol.lua
@@ -181,7 +181,14 @@ local function onIncrementalMessageReceived(_, data, _, sender, _, _, _, _, _, _
 		local messageToken = extractMessageTokenFromData(data);
 		internalMessageIDToChompSessionIDMatching[sessionID] = messageToken;
 	end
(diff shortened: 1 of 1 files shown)


Solanya:
So, just as a heads-up, we discussed this on Github and I'm actually not sure this fix should happen unless we're sure this issue doesn't come from TRP/MSP/Chomp. The `internalMessageIDToChompSessionIDMatching` is never cleared and its value comes from a string.sub, so the few reasons we could think of causing a nil value : 
- User did a /reload while receiving progress messages (doesn't seem to be the case in the issue report)
- Another bug happened and messed stuff up (that's bad)
- First msgID received wasn't 1 (that's bad)
- sessionID changed (that's bad)
If we don't know where the problem comes from, it might not be a good thing to sweep it under the rug

Meorawr:
Note of course that any issues with Chomp would very likely have been seen before in the wild. We'll need to get some more details from the person who reported the bug, since they seemed to indicate it only affected one character (?), and was rather persistent.

Katoriehooves:
Note a friend of mine has been getting this error frequently, just started happening. Sent screenshot of error:

https://i.imgur.com/CSZgIFe.png

PR #293, opened 2019-04-15
Private notes

Solanya:
Added the private notes feature to the main interface (as well as cleaned up old April Fools code). The notes tab replaces the RP.IO one-day tab and can also be directly accessed through a target frame button (when targeting a player character with a profile).

The notes tab consists of 2 text areas (except for your own profile, which only has the "top" one taking the whole tab). The top area is profile-specific and changes based on your current profile, the content being saved in the profile array - making it exportable with the rest of the profile. The bottom area is account-wide, and saved as a new SavedVariables array, `TRP3_Notes`. Both of them are also tied to the profile you're currently viewing.

The content is saved as the text is edited, similar to the Currently and OOC fields, letting people take notes on the fly.

This also includes the removal of text areas highlighting on focus. It is prone to users forgetting to click a second time to clear the highlight and losing the area content, while its usefulness is not guaranteed, considering CTRL-A is a valid shortcut to select everything.


Solanya:
Added the check on profiles during auto-purge. As a side-note, I moved the notes table to be on the same level as relationships (calling a different function to get the current profile, that difference in behaviour is a problem for another time)

Issue #298, opened 2019-06-01
Player names in world map tooltip are not following the increased color contrast setting


Ellypse:
**Describe the bug**
Player names that are shown in tooltips when looking at player locations on the world map are not using the increased color contrast setting, like in other places.

**To Reproduce**
Steps to reproduce the behaviour:
1. Have the option to dynamically increase color contrast on names enabled
2. Use the maps scan to show other TRP3 users on the map
3. Find a void elf with a dark purple name

**Expected behaviour**
The name of the edgy void elf should have been made ligther to be more readable on the dark tooltip background

**Version of the add-on:**
Total RP 3 1.6.0

**Did you try having Total RP 3 as the only enabled addon and everything else (especially something like ElvUI) disabled?**
Yes

Meorawr:
Fixed in 1.6.1.

PR #299, opened 2019-06-01
#298: Improved usage of player custom color

Ellypse:
- Removed individual settings for color contrast, consolidated into one setting on the General settings
- Added `:GetCustomColorForDisplay()` to the Player model, to get color that follows the global setting for increased color contrast. `:GetCustomColoredRoleplayingNamePrefixedWithIcon()` uses this one instead of the raw `:GetCustomColor()` now
- Added more ways to creates a `Player` to accomodate chat frames, such as `Player.static.CreateFromGUID(guid)` or `Player.static.CreateFromNameAndRealm(name[, realm])`
- Added PUBLIC `AddOn_TotalRP3.Configuration.shouldDisplayIncreasedColorContrast()` for third parties

### DIFF
--- a/totalRP3/core/Models/Player.lua
+++ b/totalRP3/core/Models/Player.lua
@@ -98,12 +98,24 @@ function Player:GetRelationshipWithPlayer()
 	return TRP3_API.register.relation.getRelation(TRP3_API.register.getUnitIDProfileID(self:GetCharacterID()));
 end
 
----@return Color|nil customColor
+---@return Ellyb_Color|nil customColor
 function Player:GetCustomColor()
 	local characteristics = self:GetCharacteristics();
 	if characteristics and characteristics.CH then
 		return Ellyb.Color.CreateFromHexa(characteristics.CH);
 	end
+end
+
+
+--- Returns a suitable for display custom color for the player.
+--- If the current user has the option to increase color contrast on darker names, the color will be affected.
+--- @ return Ellyb_Color|nil
+function Player:GetCustomColorForDisplay()
+	local color = self:GetCustomColor()
+	if color and AddOn_TotalRP3.Configuration.shouldDisplayIncreasedColorContrast() then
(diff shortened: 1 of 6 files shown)


Meorawr:
> Removed individual settings for color contrast, consolidated into one setting on the General settings

Should there be migration stuff added to make it so that if you had *any* of the individual settings checked, the (new?) setting should be enabled?

Ellypse:
> > Removed individual settings for color contrast, consolidated into one setting on the General settings
> 
> Should there be migration stuff added to make it so that if you had _any_ of the individual settings checked, the (new?) setting should be enabled?

That is a valid thing that I thought about briefly before forgetting. It seems trivial to check if one of them was enabled and switch the new setting on in that case. Good call.

Meorawr:
> Added PUBLIC AddOn_TotalRP3.Configuration.shouldDisplayIncreasedColorContrast() for third parties

I'm a bit on the fence about this change; if we add functions for *every* setting as part of the public API, suddenly removing a setting later on (like we're doing here!) either means you're left with deprecated functions that do nothing, or cleaning them up becomes an API breakage.

I think exposing config keys and just having a pair of `IsValidConfigKey(key)` and `GetConfigValue(key)` type functions is best for public use.

Ellypse:
> I'm a bit on the fence about this change; if we add functions for _every_ setting as part of the public API, suddenly removing a setting later on (like we're doing here!) either means you're left with deprecated functions that do nothing, or cleaning them up becomes an API breakage.
> 
> I think exposing config keys and just having a pair of `IsValidConfigKey(key)` and `GetConfigValue(key)` type functions (with better names) is best for **public** use.

That is indeed the future I want, but that felt out of scope for now, so I went with a quick and easy way 🤷‍♂

Meorawr:
If it's the future you want, perhaps document the function as for internal use only for now, at least until we sort out the config API?

PR #311, opened 2019-06-26
Fix MSP MU field integration

Meorawr:
If a TRP profile contains no MU field data, we're sending `"nil"` through as a string via MSP. Other addons cope with this fine, but it's still clearly incorrect.

More problematically, if we get an empty MU field back from MSP (if for example someone sets music and then later unsets it), the `emptyToNil` function is applied to `value` which makes it nil, and we were then passing that to LibRPMedia's conversion function which would explode in your face.


Meorawr:
image

_more internal screaming_

PR #314, opened 2019-06-26
Use upstream packager

Meorawr:
Use a deploy block with a restriction to only deploy tagged builds, this preserves the custom behaviour from our old packager fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,13 @@ git:
 
 install: luarocks install --local luacheck
 
-before_script: /home/travis/.luarocks/bin/luacheck .
+script: /home/travis/.luarocks/bin/luacheck .
 
-script: curl -s https://raw.githubusercontent.com/Ellypse/packager/master/release.sh | bash -s -- -p 75973 -w 24113 -g 8.2.0 -o
+deploy:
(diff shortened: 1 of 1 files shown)


Meorawr:
To confirm, the alpha/beta tag crap is already upstream with the exact same code.

```sh
		# When packaging is triggered on your repository, the generated file’s release type will
		# automatically be set based on two factors:
		#   1) If configured to package all commits, the latest untagged commit will be packaged
		#      and will be marked as an alpha.
		#   2) Otherwise, when a tagged commit is pushed, it will be flagged as either alpha, beta,
		#      or release depending on the tag itself:
		#        - If the tag contains the word "alpha", it will be marked as an alpha file.
		#        - If instead the tag contains the word "beta", it will be marked as a beta file.
		# https://authors.curseforge.com/docs/packaging
		file_type="alpha"
		if [ -n "$tag" ]; then
			if [[ "${tag,,}" == *"alpha"* ]]; then
				file_type="alpha"
			elif [[ "${tag,,}" == *"beta"* ]]; then
				file_type="beta"
			else
				file_type="release"
			fi
		fi
```

PR #326, opened 2019-07-09
Nameplate module

Meorawr:
This is the result of me going insane overnight when people started talking about nameplates.

Implements customization of the built-in Blizzard name plates, with support for showing custom player/pet names, titles, icons, colours, and basically mirrors all the options (minus hide non-RP) present in the KNP module.

I've not tested how this works in instances yet, but there's checks in place to ensure we don't touch forbidden frames (and we shouldn't be getting them in the system anyway). I'm not seeing any errors arising from anchor issues; while we do create custom widgets for the title/icon I've taken care to ensure we reset points and stuff appropriately.

As for how it looks...

image

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -33,6 +33,9 @@ globals = {
 	"SlashCmdList",
 	"StaticPopupDialogs",
 
+	-- AddOn integrations
+	"ElvUI", -- Adding oUF tags.
+
 	-- Saved variables
 	"TRP3_Presets",
 	"TRP3_Register",
@@ -70,8 +73,8 @@ read_globals = {
 	"TipTac",
 	"LibStub",
 	"TinyTooltip",
-	"ElvUI",
 	"KuiNameplates",
+	"KuiNameplatesCore",
 
 	-- FrameXML misc
 	"bit",
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -3,27 +3,27 @@ max_line_length = false
 exclude_files = {
 	"totalRP3_Localizator",
 	"totalRP3/libs",
-    "totalRP3_zTest"
+	"totalRP3_zTest"
 }
 ignore = {
-    -- Ignore undefined globals for now TODO cleanup global access
-    "112",
-    "113",
-    -- Ignore unused self. This would popup for Mixins and Objects
-    "212/self",
+	-- Ignore undefined globals for now TODO cleanup global access
+	"112",
+	"113",
+	-- Ignore unused self. This would popup for Mixins and Objects
+	"212/self",
 }
 globals = {
 	"Ellyb",
-    "TRP3_API",
-    "AddOn_TotalRP3",
-    "_G",
+	"TRP3_API",
+	"AddOn_TotalRP3",
+	"_G",
 
-    -- Globals
-    "BINDING_HEADER_TRP3",
-    "BINDING_NAME_TRP3_TOOLBAR_TOGGLE",
-    "BINDING_NAME_TRP3_TOGGLE",
-    "SLASH_TOTALRP32",
-    "SLASH_TOTALRP31",
+	-- Globals
+	"BINDING_HEADER_TRP3",
+	"BINDING_NAME_TRP3_TOOLBAR_TOGGLE",
+	"BINDING_NAME_TRP3_TOGGLE",
+	"SLASH_TOTALRP32",
+	"SLASH_TOTALRP31",
 	"msp_RPAddOn",
 	"msp",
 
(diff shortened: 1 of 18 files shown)


Ellypse:
The spelling should be "nameplate" just like how Blizzard writes it in their UI

Solanya:
If we go ahead with this change, I think the Kui module needs to be merged into TRP, similarly to tooltip/chat modules. At the very least it needs to use the same settings.
We'd also need to look at a later date into other nameplates addons, including customizing ElvUI nameplates...

.luacheckrc
```diff
@@ -5,6 +5,12 @@ exclude_files = {
 	"totalRP3/libs",
     "totalRP3_zTest"
 }
+
+-- Allow unused arguments. Naming arguments even if they're unused is
+-- perfectly fine since you can't exactly tell what "_" is for when it isn't
+-- explicitly named.
+unused_args = false
```
Ellypse: This rule is intended to be enabled on the project.

totalRP3/modules/NamePlates/NamePlates.lua
```diff
@@ -0,0 +1,1124 @@
...
+	return profile.data;
+end
+
+-- Returns an appropriate default OOC indicator style for configuration.
+local function GetDefaultOOCIndicator()
+	local defaultOOCIndicator = "TEXT";
+
+	-- The default OOC icon should inherit from the tooltip setting if
+	-- available, but without adding a module dependency please.
+	if TRP3_Configuration then
```
Ellypse: It could be a good thing to move the OOC indicator setting outside of the tooltip settings so it can be shared across modules, like what was done with the increased color contrast option.
I would even make it [OCC] by default and move it to advanced settings. Only Lenwë uses the red dot.

totalRP3/modules/NamePlates/NamePlates.lua
```diff
@@ -0,0 +1,1124 @@
...
+	TRP3_Config.registerConfigKey(CONFIG_NAMEPLATES_OOC_INDICATOR, GetDefaultOOCIndicator());
+
+	-- Prevent the module from actually being *enabled* if there's any of the
+	-- conflicting addons installed.
+	for _, addonName in ipairs(CONFLICTING_ADDONS) do
+		local state = GetAddOnEnableState(TRP3_Globals.player, addonName);
+		if state == 2 then
+			return false, format(L.NAMEPLATES_ERR_ADDON_CONFLICT, addonName);
+		end
+	end
```
Ellypse: Could this be moved higher up, so we don't register config keys for nothing?

totalRP3/modules/NamePlates/NamePlates.lua
```diff
@@ -0,0 +1,1124 @@
...
+		C_Timer.NewTicker(1, function()
+			local currentUser = Player.GetCurrentUser();
+			local isInCharacter = currentUser:IsInCharacter();
+
+			if previousState ~= isInCharacter then
+				previousState = isInCharacter;
+				self:OnRoleplayStatusChanged();
+			end
+		end);
+	end)();
```
Ellypse: This is probably okay for now, but it probably makes sense to add the missing event.

totalRP3/modules/NamePlates/NamePlates.lua
```diff
@@ -0,0 +1,1124 @@
...
+
+		return nameColor;
+	elseif UnitIsCombatPet(unitToken) then
+		-- Combat pets use companion pet profiles.
+		local profile = GetCombatPetProfile(unitToken);
+		if not profile then
+			return nil;
+		end
+
+		return profile.NH and Color.CreateFromHexa(profile.NH) or nil;
```
Ellypse: This will not provide a color suitable for display.
```suggestion
		local petColor = profile.NH and Color.CreateFromHexa(profile.NH);
		if petColor and AddOn_TotalRP3.Configuration.shouldDisplayIncreasedColorContrast() then
			petColor:LightenColorUntilItIsReadableOnDarkBackgrounds();
		end
		return petColor;
```

totalRP3/modules/NamePlates/NamePlates.lua
```diff
@@ -0,0 +1,1124 @@
...
+end
+
+-- Module registration.
+TRP3_API.module.registerModule({
+	name        = L.NAMEPLATES_MODULE_NAME,
+	description = L.NAMEPLATES_MODULE_DESCRIPTION,
+	version     = 1,
+	id          = "TRP3_NamePlates",
+	onInit      = function(...) return TRP3_NamePlates:OnInitialize(...); end,
+	onStart     = function(...) return TRP3_NamePlates:OnEnable(...); end,
```
Ellypse: I don't like this. This not how all other modules are declared (we have local functions `onInit`) and this looks like just trying to retrofit a AceAddon-ish based system with `AddOn:OnEnable()` method with a self bound to the scope for no actual usage (your calls to self could use `TRP3_NamePlates` directly).

If you take the entire add-on and existing modules into consideration I think this is an anti pattern

totalRP3/modules/NamePlates/NamePlates.lua
Meorawr: Moving tooltip settings as described is out of the scope of this feature. If another PR comes along to do that, I'll merge in the appropriate changes.

totalRP3/modules/NamePlates/NamePlates.lua
Meorawr: The module itself is stateful, hence the preference for `self` since I can easily get at it from a global context and debug the whole module via `/dump` and `/tinspect`.

There was definitely a borrowing of AceAddon terminology, but that's mostly because my initial design would have had the global "enable customizations" toggle also call the appropriate "OnDisable" function to fully turn off events/hooks.

totalRP3/modules/NamePlates/NamePlates.lua
Meorawr: Is there a reason to _not_ register config keys? No UI is added unless the module actually starts up, but one of the annoyances as noted with the tooltip settings is the fact that config keys will error if they're queried but not yet registered.

I'd rather have the keys registered all the time at a global level somewhere else honestly, for _all_ the settings. That way we don't need to paste them around the whole shop and run into issues where module A is disabled but it registers a key that module B might be interested in.

Ellypse: Okay, that makes sense

totalRP3/totalRP3.lua
```diff
@@ -104,6 +104,29 @@ local function loadingSequence()
...
+		if unitID ~= TRP3_API.globals.player_id then
+			return;
+		end
+
+		local currentUser = AddOn_TotalRP3.Player.GetCurrentUser();
+		local isInCharacter = currentUser:IsInCharacter();
+
+		if currentRoleplayStatus ~= isInCharacter then
+			currentRoleplayStatus = isInCharacter;
+			TRP3_API.events.triggerEvent(TRP3_API.events.ROLEPLAY_STATUS_CHANGED);
```
Ellypse: Could you pass the new roleplay status as a parameter to the event since you already have it here, so future listeners don't have to also go look for it?

Meorawr: I thought about it, but I had reservations. The event signals that the status has changed, but the only value I can meaningfully pass through at the moment is a boolean (true for IC, false for OOC). 

Passing a boolean doesn't feel particularly obvious from the standpoint of using that event if you think it out loud; "okay, the roleplay status has changed and my new state is... `true`?".

If a payload were to be added I'd prefer it to send a token like "in_character" or "out_of_character" that's a bit more descriptive of the actual state.

Ellypse: You could add something to our enums and pass that value, so it can be used on the other ends for comparing. Numbers are probably lighter than strings for that kind of values.

Ellypse: Or even use booleans for the values, to make it "compatible" with all existing functions

Meorawr:
In terms of known issues/what needs testing:

* The Blizzard nameplates don't work too well with the added title line if not in name-only mode. I think the solution here is hide the title line if the healthbar is showing.
* Kui nameplates might still have issues with name colouring; I took a sledgehammer and used that to "solve" the problem but I'm not sure if this breaks stuff in Kui (namely, health colouring in the name). It works in a minimal sense, but needs more effort still.
* I've not tested how well these work with regards to units whose "friendly" state might change (eg. what happen if you duel someone?).
* Obviously the lack of a "hide non-RP setting".
* Speaking of settings, there's a chance that not all settings can be suitably applied to all nameplate addons - or that some settings might require special consideration. How would we communicate this to users? The one thing I'm thinking of in particular is Kui and the title text option; we piggy back off the guild text line which (currently) requires that setting be enabled.

Meorawr:
Of those previous notes:

* Resolved default nameplates not working too well outside of name-only mode.
* Added tooltip stuff to config settings to outline certain requirements.
* Improvements to Kui name colouring issues, including the OOC indicator prefix. Not fully tested, but seems to work.

Meorawr:
Closing PR for now due to inactivity; I'll probably pick this up again in the future and re-open it then.

PR #327, opened 2019-07-10
Clear All Points on target_frame anchor update

Solanya:
nice joke of a system, blizz


Ellypse:
Why was this merged without the requested change done?

Meorawr:
Because as noted the context that function is called in only occurs once on initialization.

Ellypse:
Oh okay. I didn't see that here. Move along then.

Meorawr:
image

Context for future reference :)

PR #336, opened 2019-08-16
Classic support

Solanya:
The big one. Icon library is missing but this should be a separate issue, even if it means a temporary new file.

This wasn't tested on a live environment and as such might have lingering bugs, but as far as I've been able to check :
- All the missing icons have been replaced on the Classic version.
- The scan button has been fixed (was sneaking under the map)
- Support for the old way of playing musics (because Blizzard)
- "Disable while in war mode" will disable while in PvP on Classic
- Image list for Classic (and icon list, to be replaced)
- TEMP icon has been replaced by the question mark **on all versions**, made sense to me but open to feedback on reverting it to be Classic-only.

Remaining issues for launch :
- Icons in the credits are missing on Classic. Given that we are going for duplicate localisation, this will be fixed there.

Remaining issues after launch : 
- The mount/pet API calls have been disabled until we can make proper tests in-game.
- The main tutorial button has been disabled until we can replace the missing template.


totalRP3/libs/libs.xml
```diff
@@ -39,6 +39,6 @@
 	<Include file="LibDeflate\lib.xml" />
 	
 	<!-- LibRPMedia https://github.com/wow-rp-addons/LibRPMedia -->
-	<Include file="LibRPMedia\LibRPMedia-Retail-1.0.xml" />
+	<Include file="LibRPMedia\LibRPMedia-1.0.xml" />
```
Meorawr: Maybe change this to a packager comment like you've done in the TOC (`<!--@retail@-->`/`<!--@non-retail@-->`) to load the retail/classic specific files.

totalRP3/resources/imageList.lua
```diff
@@ -17,6 +17,11 @@
 --- limitations under the License.
 ----------------------------------------------------------------------------------
 
+local version = select(4, GetBuildInfo());
+if version < 80100 then
```
Meorawr: This should probably test the WOW_PROJECT_ID constant to be consistent with other checks.

totalRP3/totalRP3.lua
```diff
@@ -47,62 +47,78 @@ end
...
+		TRP3_API.Ellyb.Popups:OpenURL("https://www.curseforge.com/wow/addons/total-rp-3", "You are trying to use |cffff0000Total RP 3: Classic|r. Please install the retail version of Total RP 3 instead.");
+		clientMismatch = true;
+	elseif not isClassicBuild and WOW_PROJECT_ID == WOW_PROJECT_CLASSIC then
+		TRP3_API.Ellyb.Popups:OpenURL("https://www.curseforge.com/wow/addons/total-rp-3-classic", "You are trying to use the |cffff0000retail|r version of Total RP 3. Please install Total RP 3: Classic instead.");
+		clientMismatch = true;
+	end
+	--@end-non-debug@]===]
+	--endregion
+
+	if not clientMismatch then
```
Meorawr: This block encompasses the entire remaining function (minus that one log statement at the end). I'd suggest just getting rid of `clientMismatch` entirely and making the checks above return early.

PR #342, opened 2019-08-27
Rename War Mode related options to PvP for Classic

Ellypse:
Because War Mode doesn't make sense on Classic.

Since these are two separate projects on Curse I'm not sure how you are going to handle localization 🤔


Meorawr:
The packager would pull translations from the CF project ID used in the build, so locale stuff would need uploading to both I'd think.

PR #345, opened 2019-08-30
Language Field

Meorawr:
This field displays an icon in the tooltip for a user indicating their preferred roleplay language. On an upgrade, this will initialize to the addon locale if set or will fall back to the client locale.

Wow_2019-08-30_01-29-05

The icon will only display in tooltips if it differs from the users currently selected locale; so you won't see enUS flags if you're running TRP with the enUS locale selected since the info isn't all that useful in that case.

The setting can be changed from the dashboard as a new dropdown with the RP status/RP XP options.

Wow_2019-08-30_01-38-25

*Note: Flags aren't displayed in that dropdown because the Chinese locales artifically scale up the flags, making them look unaligned and horrible.*

If the user changes the addon locale, we'll run through all the profiles and if the language set for any profile matches the locale we're changing *from*, we'll update it automatically. I'd have preferred to prompt for this, but we're already popping up a prompt to do a reload UI in this case so having two back to back prompts feels nasty. This behaviour feels sensible as a default.

We only support the locale codes present in our localization system (which mirror those returned by GetLocale); for accuracy in EU/RU clients we'll transform enUS and ptBR to enGB and ptPT respectively. This pulls in LibRealmInfo to do so on Retail, but will fall back to `GetCurrentRealmName()` if that data is unavailable.

The field is exposed via LibMSP in the `LC` field as described in wow-rp-addons/LibMSP#26.


Meorawr:
Just to confirm the flag image licensing stuff: https://github.com/tkrotoff/famfamfam_flags

> As stated on http://famfamfam.com/lab/icons/flags/: "These flag icons are available for free use for any purpose with no requirement for attribution."

totalRP3/core/Models/Player.lua
```diff
@@ -146,7 +149,19 @@ function Player:IsCurrentUser()
...
+function Player:GetRoleplayLanguage()
+	return self:GetInfo("character/LC");
+end
+
+function Player:GetRoleplayExperience()
+	return self:GetInfo("character/XP");
+end
+
+function Player:GetRoleplayStatus()
+	return self:GetInfo("character/RP");
```
Ellypse: Nice, you get bonus points for making this model evolve 👍

totalRP3/modules/dashboard/dashboard.lua
```diff
@@ -318,3 +294,343 @@ TRP3_API.events.listenToEvent(TRP3_API.events.WORKFLOW_ON_LOADED, function()
 		TRP3_API.toolbar.toolbarAddButton(Button_RPStatus);
 	end
 end);
+
+--- Mixin for dropdown menus on the dashboard status frame. These widgets
+--  come with an attached label for the control and a dropdown as a subframe.
+local DashboardStatusDropDownMixin = {};
```
Ellypse: Could I suggest moving this to a new file? The dashboard file itself is quite big, it could use some splitting

PR #346, opened 2019-08-30
Add IC/OOC slash commands

### LINKED ISSUE #329: Slash command to switch character status
We could add a slash command to be able to switch between IC and OOC similar to the behaviour of the toolbar button. We need to make sure the toolbar button still gets properly switched as well.

Meorawr:
Closes #329. Execute `/trp3 ic` to go IC, `/trp3 ooc` to go OOC. There's no toggle slash command, but one could be added if a name can be thought up.

This'll likely have merge conflicts with #345. The slash command is in the dashboard file since that's where the toolbar button for this is already registered, so makes sense to keep it all together (even if it is a bit weird).


Meorawr:
Changed to `/trp3 status <ic|ooc|toggle>`.

PR #348, opened 2019-08-30
Push default log output to its own logger

Meorawr:
This just moves the logging output through the Ellyb.Logger class API, meaning by default log messages will (attempt) to go to a separate "Logs" tab unless there's warnings/severe level messages.

I'm sick of the spam in the default chat frame. Has no impact outside of debug builds.


Ellypse:
💜

PR #352, opened 2019-09-01
Fix errors loading when target bar is disabled

Meorawr:
Disabling the target bar module breaks the addon quite horrifically, but you *can* still get to the settings page to re-enable it.

This fixes the errors by adding a few key `if TRP3_API.target` checks in areas where it was initially failing.

One issue that I still need to look at after this fix is with the module disabled, it seems the glance portion doesn't quite take the hint:

image


Meorawr:
Nevermind on that secondary glance bar issue as it seems to be intended; the target frame just adopts the glance bar and re-positions it from what I can tell.

PR #362, opened 2019-09-05
Drycode fix for #361

### LINKED ISSUE #361: Changing icon size or max icons w/ cape/helmet switch enabled
**Describe the bug**
While having either (or both) of the new cape switch or helmet switch options enabled, the error linked below occurs

**To Reproduce**
Steps to reproduce the behaviour:
1. Log in with TRP3: Classic 1.6.6 enabled
2. Enable the new Cape switch and/or the Helmet switch option in Frames settings
3. Change the Icons size and/or the Max icons per line slider
4. See error (as well as the Toolbar becoming empty and not changing size/formation)

**Expected behaviour**
No error. Duh..

**Screenshots or error message**
[https://www.wowace.com/paste/998cf4fa](https://www.wowace.com/paste/998cf4fa)

**Version of the add-on:**
Total RP 3: Classic 1.6.6

**Did you try having Total RP 3 as the only enabled addon and everything else (especially something like ElvUI) disabled?**
No. The error implies this is caused by some of the changes between 1.6.5 and 1.6.6

Meorawr:
Fixes #361 (pending testing). This should prevent us assigning Ellyb.Icon instances to the button structure.

### DIFF
--- a/totalRP3/core/impl/utils.lua
+++ b/totalRP3/core/impl/utils.lua
@@ -369,7 +369,21 @@ end
 -- Return an texture text tag based on the given icon url and size. Nil safe.
 function Utils.str.icon(iconPath, iconSize)
 	iconPath = iconPath or Globals.icons.default;
-	return Utils.str.texture("Interface\\ICONS\\" .. iconPath, iconSize);
+	return Utils.str.texture(Utils.getIconTexture(iconPath), iconSize);
+end
+
+--- Gives the full texture path of an individual icon.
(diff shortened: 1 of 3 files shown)


Ellypse:
I made other parts of the toolbar accept Ellyb.Icon (see https://github.com/Total-RP/Total-RP-3/blob/development/totalRP3/modules/toolbar/toolbar.lua#L209 and https://github.com/Total-RP/Total-RP-3/blob/development/totalRP3/modules/toolbar/toolbar.lua#L245), the proper fix would be to fix where it doesn't handle it correctly.

Issue #74, opened 2018-01-28
Add option to color quoted text in emotes into /say messages


Ellypse:
I change my mind. I want it now.

Ellypse:
Scratch that. I want my TRP2 "cut messages" feature back actually.
Typing `/say Hi! *waves* How are you today? *smiles*` would output:  
```
Ellypse says: Hi!
Ellypse waves.
Ellypse says: How are you today?
Ellypse smiles.
```

Ellypse:
And obviously typing `/me shrug. "Who knows?" Pat his cat` would output:
```
Ellypse shrug.
Ellypse says: Who knows?
Ellypse pat his cat.
```

PR #365, opened 2019-09-22
Speech detection

Solanya:
I know, #74 asked to split messages. It'll stay open, but it would require a bit more work + the whole UCM/EmoteSplitter feature. That took me less than an hour to do and at least we'll have a small something to give everyone with the 8.2.5 update (especially Classic players who would otherwise get a useless TOC update)

### DIFF
--- a/totalRP3/tools/Locale.lua
+++ b/totalRP3/tools/Locale.lua
@@ -576,6 +576,7 @@ Possible status:
 (When this option is enabled, you can ALT-SHIFT-Click on a name when you want the default behavior and insert the character name instead of the full RP name.)]],
 	CO_CHAT_MAIN_COLOR = "Use custom colors for names",
 	CO_CHAT_INCREASE_CONTRAST = "Increase color contrast",
+	CO_CHAT_SHOW_OOC = "Show OOC indicator",
 	CO_CHAT_USE_ICONS = "Show player icons",
 	CO_CHAT_USE = "Used chat channels",
 	CO_CHAT_USE_SAY = "Say channel",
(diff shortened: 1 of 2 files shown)


totalRP3/tools/Locale.lua
```diff
@@ -592,6 +592,8 @@ Possible status:
 	CO_CHAT_MAIN_OOC_COLOR = "OOC color",
 	CO_CHAT_MAIN_EMOTE_YELL = "No yelled emote",
 	CO_CHAT_MAIN_EMOTE_YELL_TT = "Do not show *emote* or <emote> in yelling.",
+	CO_CHAT_MAIN_SPEECH = "Speech detection",
+	CO_CHAT_MAIN_SPEECH_USE = "Use speech detection",
```
Meorawr: Can we add a tooltip text that says what "speech detection" is? Namely, that it'll colour quoted text in white.

Otherwise we'll get some *valued user* who thinks we've implemented some sort of TTS feature or something daft.

totalRP3/tools/Locale.lua
```diff
@@ -1298,6 +1295,15 @@ Changes your character status to the specified option:
...
 ]],
 
+	------------------------------------------------------------------------------------------------
+	--- PLACE LOCALIZATION NOT ALREADY UPLOADED TO CURSEFORGE HERE
+	--- THEN MOVE IT UP ONCE IMPORTED
+	------------------------------------------------------------------------------------------------
+
+	CO_CHAT_MAIN_SPEECH = "Speech detection",
+	CO_CHAT_MAIN_SPEECH_USE = "Use speech detection",
+	CO_CHAT_MAIN_SPEECH_USE_TT = "Text surrounded by quotation marks will be coloured as if written in /say.",
```
Meorawr: "coloured" what is this british english

PR #375, opened 2019-12-12
Speech detections tweaks

Solanya:
- Speech detection will only apply to emotes
- This applies to NPC emotes (and NPC speech) which will also get other detections (the content of NPC emotes being stored in the name, it wasn't passing through the detection function).

### DIFF
--- a/totalRP3/modules/chatframe/chatframe.lua
+++ b/totalRP3/modules/chatframe/chatframe.lua
@@ -354,7 +354,7 @@ TRP3_API.utils.getCharacterInfoTab = getCharacterInfoTab;
 
 ---@param message string
 ---@param NPCEmoteChatColor Color
-local function detectEmoteAndOOC(message, NPCEmoteChatColor)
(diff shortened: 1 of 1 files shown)


totalRP3/modules/chatframe/chatframe.lua
```diff
@@ -354,7 +354,7 @@ TRP3_API.utils.getCharacterInfoTab = getCharacterInfoTab;
 
 ---@param message string
 ---@param NPCEmoteChatColor Color
-local function detectEmoteAndOOC(message, NPCEmoteChatColor)
+local function detectEmoteAndOOC(message, isEmote, NPCEmoteChatColor)
```
Ellypse: I would usually suggest adding new function parameters at the end of the list. It's a local function so I assume you caught all usages of the function so it's probably fine here.

Solanya: Yeah I was split on that. I knew there was a single usage of the function and I was thinking about the chat color possibly being nil. But given that isEmote can also be nil, I'll swap them around

PR #377, opened 2019-12-22
Added apostrophe to the icon pattern matching

Solanya:
Opening a PR right now but won't merge until next week.

Some icons break formatting because they have an apostrophe (achievement_boss_kael'thassunstrider_01, achievement_dungeon_drak'tharon_heroic) which was not currently included as a possible character in the icon pattern.

**It should be considered to switch the pattern with [^:]+ or [^%:]+ to avoid issues if more unexpected characters were to be found, unless there's a specific reason not to do so. Will need to test to be sure.**


Solanya:
Test was successful.

Issue #35, opened 2018-01-02
Add a search box to the list profiles


Ellypse:
>Zwixx, comment #941
>I use many Profiles and change the character to this Profile. If I see correctly, it is sorted by Name of the Profile. Is it possible to make a filter box on top of the character profile page? Or to group profiles with categories I can create?

Use a search box to look for a specific profile without having to scroll a long list of profiles.

PR #415, opened 2020-07-25
Added search bar in own profiles lists ( #35 )

Solanya:
A search bar was added in both character and companion profiles lists. Not much to say, it behaves the same way as the directory search bar does (search on enter, safe match on the lower strings). Added an empty text for character profiles list and adapted the companion one depending on whether a search is active or not.

No reliance on new code and no breaking changes so dev targeted.

### DIFF
--- a/totalRP3/core/impl/profiles.lua
+++ b/totalRP3/core/impl/profiles.lua
@@ -29,6 +29,7 @@ TRP3_API.profile = {};
 local Globals, Events, Utils = TRP3_API.globals, TRP3_API.events, TRP3_API.utils;
 local loc = TRP3_API.loc;
 local unitIDToInfo = Utils.str.unitIDToInfo;
+local safeMatch = Utils.str.safeMatch;
 local strsplit, tinsert, pairs, type, assert, _G, table, tostring, error, wipe = strsplit, tinsert, pairs, type, assert, _G, table, tostring, error, wipe;
 local displayDropDown = TRP3_API.ui.listbox.displayDropDown;
 local handleMouseWheel = TRP3_API.ui.list.handleMouseWheel;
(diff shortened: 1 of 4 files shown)


totalRP3/core/impl/profiles.lua
```diff
@@ -223,10 +225,21 @@ end
...
 	wipe(profileListID);
+	local profileSearch = Utils.str.emptyToNil(TRP3_ProfileManagerSearch:GetText());
 	for profileID, _ in pairs(profiles) do
-		tinsert(profileListID, profileID);
+		if not profileSearch or safeMatch(profiles[profileID].profileName:lower(), profileSearch:lower()) then
+			tinsert(profileListID, profileID);
+		end
+	end
+
+	local size = tsize(profileListID);
```
Meorawr, line 235: `profileListID` is being used as an array-like table as evidenced by the `tinsert` call - so `tsize` could be changed to `#` and be logarithmic time instead of function-call-overhead-and-linear-time.

totalRP3/modules/register/companions/register_companions_profiles.lua
```diff
@@ -241,16 +242,26 @@ end
...
+	wipe(profileListID);
+	local profiles = getProfiles();
+	local profileSearch = Utils.str.emptyToNil(TRP3_CompanionsProfilesSearch:GetText());
+	for profileID, _ in pairs(profiles) do
+		if not profileSearch or safeMatch(profiles[profileID].profileName:lower(), profileSearch:lower()) then
+			tinsert(profileListID, profileID);
+		end
+	end
+
+	local size = tsize(profileListID);
```
Meorawr, line 254: As above.

totalRP3/tools/Locale.lua
```diff
@@ -1405,6 +1405,8 @@ We are aware of a current issue on Retail causing **quest item usage from the ob
 	--- THEN MOVE IT UP ONCE IMPORTED
 	------------------------------------------------------------------------------------------------
 
+	PR_PROFILEMANAGER_SEARCH_PROFILE = "Search profile",
+	PR_PROFILEMANAGER_EMPTY = "No matching profile",
```
Meorawr, line 1409: "No profiles found" might be more typical for a search result error.

Issue #179, opened 2018-07-24
Switch default character status to OOC


Solanya:
Potentially related to https://github.com/Ellypse/Total-RP-3/issues/47 : once a tutorial experience is implemented in TRP and there's a callout for first-time/unexperienced users on how to switch between IC and OOC, it makes sense to have the default status as OOC, as newly-created characters are rarely IC upon creation, and the flag is more often than not forgotten about rather than knowingly kept.

Issue #122, opened 2018-05-14
Option to use a specific profile as the "default" profile for new characters


Ellypse:
> **Verisimi on CurseForge:**
> Is there any way to turn off automatically making a profile every time I roll a character? I have a LOT of PvE alts on other servers that I don't need/want profiles for and it just makes it hard to find the profile I  DO want to switch to when there's such a huge list for characters I am never going to need a profile on. I'd rather have an 'opt in' to save a profile  when I roll characters. This is a great add on and I love it and you clearly put a lot of time and effort into it. Thanks for all you do, and no big deal if this isn't possible.

We would want a way for Total RP 3 to use only one profile by default, instead of creating new ones for each character. This option would be opt-in (to avoid confusion for new players switching characters) and would be reset if the profile is deleted.

Connor-Macleod:
What if instead of making a single profile by default, we prevent profiles from showing up in the list until they've been changed away from containing nothing but the toon name/class?

Ellypse:
I don't think hiding existing profile to users based on if they have valid content or not is a good idea. It can lead to confusion, you would see your new character using a profile that is not currently listed. Also, you will never be able to delete those profiles, so they will pile up silently.

Somewhere in the settings, we should have a dropdown to choose between creating a new profile or using a default one, something like:

```
New characters uses: [New profile        V]
					 [Ellypse-KirinTor    ]
					 [Marybell            ]
					 [Baa'l               ]
```

Meorawr:
> Instead of profile tabs, you would see a single message indicating that you need to create a new profile to be able to edit it.

As long as this single message has a button or link right there on the page that'd allow users to perform that action, instead of saying "go to the X page and do Y". Optionally this could prompt for a profile name and, if so, should default to the current profile naming scheme ("character - server").

PR #416, opened 2020-07-30
Default profile ( #122 )

Solanya:
As discussed in the ticket, unlinked characters will now be linked to a single default profile instead of having a new one created for every character. This profile is currently always shown first in the list (could be put in last position, I just feel like a fixed position is better than it being in the alphabetical sorting) and its basic data is automatically updated depending on the character.

Actions have been hidden for this profile, as well as tab buttons in the profile details window. In the latter, if the default profile is selected, a message warns to create a new profile (with a button for this directly underneath) or link to an existing one. Creating a new profile there will automatically select it (I'm unsure if this behaviour should also be replicated in the Profiles list window, I think it should but importing a profile cannot currently be done on the selected profile - could also be something to look into) and move to the details of the new profile.

As notes are currently a tab in the profile details, their dedicated button has been disabled in the player's target frame if the default profile is selected, but I'll think it over. Extended features (inventory, bags, quest log) will work as with any other profile.

Points still to consider: Do we want to show more on the details tab? Would we use this opportunity to look at #179 (This might be tricky with new characters all tied to the default one)?

Last point: Will have to take this PR into account for #415 so that the default profile is hidden during a search.

### DIFF
--- a/totalRP3/core/impl/profiles.lua
+++ b/totalRP3/core/impl/profiles.lua
@@ -242,27 +242,25 @@ end
 -- Refresh list display
 local function uiInitProfileList()
 	wipe(profileListID);
+	local defaultProfileID = getConfigValue("default_profile_id");
 	local profileSearch = Utils.str.emptyToNil(TRP3_ProfileManagerSearch:GetText());
 	for profileID, _ in pairs(profiles) do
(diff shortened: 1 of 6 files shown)


Solanya:
Ready for review.

Meorawr (approved):

PR #427, opened 2020-10-13
Auto-reset password

Solanya:
Change was waiting for 9.0 as to not spam older versions with warnings. Irrelevant to Classic since the channel changes.


Meorawr (approved):
We are a clever species.

IvanVilla:
Tonight the admin of xtooltip channel in my server updated the addon.
Everybody reading "the password changed", and as ressult, it was a bit embarrising to him.

Investigating, we imagined that there was something like a thread that autodelete the password to skip evil users plans. And reading the code we are almost sure of that.

Thank you for the feature!

Solanya:
Never update addons the day before a patch! :p

PR #429, opened 2020-10-14
Force dropdowns to avoid AddNewbieTip in 9.x

Meorawr:
This function was deprecated in BFA and has since been removed, but MSA hasn't fixed this in its own code yet. This can be tested by mousing over some dropdown buttons - for instance the template dropdown in the profile editor.

Forcing the tooltipOnButton field of the info to true in 9.x clients will force MSA to take a branch that avoids using the removed function for now.

### DIFF
--- a/totalRP3/core/impl/ui_tools.lua
+++ b/totalRP3/core/impl/ui_tools.lua
@@ -163,7 +163,8 @@ local function openDropDown(anchoredFrame, values, callback, space, addCancel)
 				else
 					info.text = text;
 					info.isTitle = false;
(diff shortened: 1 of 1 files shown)


Meorawr:
[Upstream ticket ref](https://www.curseforge.com/wow/addons/msa-dropdownmenu-10/issues/6) to get it fixed there as well. We should keep this change present however even if fixed overnight since the Blizzard approach is to no longer show tooltips if `tooltipOnButton` isn't set - it's possible this same behaviour might also be adopted by the library.

Solanya (approved):

PR #434, opened 2020-10-14
Side anchors on the default profile warning

Solanya:
Added anchors to prevent the default profile warning translated text from overflowing

Untitled


Meorawr (approved):
I did think about questioning this with the initial PR but given my tendency to make you do tiny nitpicky changes I figured I'd leave it alone...

I guess this is just a hint that I should double down on such demands for future PRs.

PR #440, opened 2020-10-15
Migrate from Travis to GitHub Actions

Meorawr:
This migrates our usage of Travis to GitHub actions. The downside to this is we can no longer blame Travis when things go wrong, and screaming "GitHub" is harder, so I'd suggest we start referring to GitHub as "Gary" when it starts failing our builds for that _human touch_.

I've split our workflows into two - one for PRs, and one for tagged releases. The PR workflow is there to run luacheck/editorconfig-lint as we did with our Travis builds, and will also additionally take a leaf from the WeakAuras book and run the packager in a dry-run configuration and upload the artifacts to GitHub's infinite artifact storage space, in case we ever have a need to pull a zip down from a PR or want to inspect the contents of it. The PR workflow retains the "if targeting master you're an idiot" logic as well.

For tags, the release workflow is executed. This skips the luacheck/editorconfig linting and will focus solely on packaging - so if we're in a pinch and we want to force a release without a care for some silly whitespace issue then we have that capability now.

In both cases we use the action stuff that was merged into the packager repository recently to dispatch that, as well as the new "classic" game version support so we don't need to constantly change the version numbers in our workflows.

At the moment the luacheck action is using a specific commit on the upstream repo that provides annotation support. Annotations are also present for our editorconfig-check step with our own composite action.

If you want to check how this all works, I made [a literal bombsite](https://github.com/Meorawr/Total-RP-3/pull/1) in a fork where you can see [every run](https://github.com/Meorawr/Total-RP-3/actions) of each workflow and [even the future of TRP](https://github.com/Meorawr/Total-RP-3/releases/tag/v3.0). It's pretty rad.

TLDR; Travis smells, long live Gary. Gary does everything Travis did, but better. Good Gary.


Ellypse:
While you're changing the `.luacheckrc` file, you could remove the lines excluding the totalRP3_Localizator and totalRP3_zTest folders. Those debug thingies are long gone.

Solanya (approved):

PR #444, opened 2020-10-16
Add LibChatAnims dependency

Meorawr:
This is a taint fix for Blizzard's FCF family of functions. TRP installs (several) chat message event filters and replaces the GetColoredName function, this causes us to taint execution inside the Blizzard ChatFrame message event handler which seeps into the UI's FCF functions to make the chat frame tabs glow.

The taint from making the tabs glow spreads practically everywhere - you'll see it in the talent frame when changing specializations, potentially in the spell book frame when a spell is learned, maybe even on your windows at home.

A quick test with this running and whispering myself, followed by recursively checking all global taint showed a significant improvement, so this should resolve the issues we were (rarely) seeing with "action blocked" errors pointing to the spellbook in weird and wonderful ways.

### DIFF
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -21,6 +21,7 @@ ignore:
 embedded-libraries:
   - ace3
   - callbackhandler
+  - libchatanims
   - libcompress
   - libdatabroker-1-1
   - libdbicon-1-0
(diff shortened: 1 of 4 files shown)


Solanya (approved):
If it doesn't fix everything I'm blaming Simpy

Issue #450, opened 2020-10-21
TRP rearranging my chat channel orders


EzraleonBloodstrike:
**Describe the bug**
I recently became a Guide in World of Warcraft and set my chat channels to the order i wanted them (specifically channel 3 as the newcomer chat). When i would log out and then back in the next day, the chat channels were completely reorganized to swap channels 2 and 3. I played around with my addons thinking it might be one of them and was able to determine the reorganization was coming from TRP3.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Chat settings
2. Click on Channels
3. Drag NewcomerChat to other order
4. Upon logout, the channels are reorganized for me.

**Expected behaviour**
I expected that the chat channels would stay in the positions i placed them in.

**Screenshots or error message**
![wowscreenshot](https://user-images.githubusercontent.com/73210979/96671599-cc77fd00-1327-11eb-94cd-8901208549e9.PNG)

**Version of the add-on:**
Total RP 3 2.0.3, build 90

**Did you try having Total RP 3 as the only enabled addon and everything else (especially something like ElvUI) disabled?**
Yes

Ellypse:
Hello. In the Advanced Settings there is an option to disable a workaround in place in Total RP 3 that keeps the communication channel it uses (xtensionxtooltip2) last. I think it might be this workaround that is messing things up here. Until this can be properly fixed I would suggest to try disabling this setting and see if you are still having the issue.

PR #460, opened 2020-11-10
Only move channels after step 2

Solanya:
I don't remember why we made that change. Maybe because of the guide channel issue, or because it bugged out at some point? 🤷 
Great change though 13/10 would do it again


Meorawr:
#450 I think.

Meorawr (approved):

PR #461, opened 2020-11-11
Tags adjustments

Solanya:
First change is needed for adjustments to the objective tracker in Extended, it's an option to omit brackets around links (lets us show the Campaign/Quest names without brackets around them.
Second change is to allow icons to be used as "text" for links. Ultimately better parsing would be nice, but that's talk for another day.


Meorawr (approved):
Don't much like adding another boolean parameter since it makes the toHTML call a pain at each call site (`toHTML("some text", true, false)` isn't very descriptive) but a proper solution can wait until we decide if we want to make changes to this with profile validation work, should that ever happen.

Solanya:
Don't need to fill those parameters if you don't need them (nil, nil will keep the link green between brackets), it's only when required. But yes, if we add too many parameters we could always pass a table with attributes instead or something.

PR #466, opened 2020-11-17
Prevent nil indexing of full title in tooltips

Meorawr:
User report:

```
1x ...\totalRP3\modules\register\main\register_tooltip.lua:505: attempt to index local 'fullTitle' (a nil value)
[string "@totalRP3\modules\register\main\register_tooltip.lua"]:505: in function <...\totalRP3\modules\register\main\register_tooltip.lua:406>
[string "@totalRP3\modules\register\main\register_tooltip.lua"]:1068: in function <...\totalRP3\modules\register\main\register_tooltip.lua:1017>
[string "@totalRP3\modules\register\main\register_tooltip.lua"]:1185: in function <...\totalRP3\modules\register\main\register_tooltip.lua:1184>
[string "=[C]"]: in function `xpcall'
[string "@totalRP3\libs\Ellyb\Tools\EventsDispatcher.lua"]:74: in function `TriggerEvent'
[string "@totalRP3\core\impl\events.lua"]:120: in function `fireEvent'
[string "@totalRP3\modules\register\main\register_tooltip.lua"]:1166: in function <...\totalRP3\modules\register\main\register_tooltip.lua:1162>
[string "=[C]"]: in function `SetUnit'
[string "@FrameXML\UnitFrame.lua"]:462: in function `UnitFrame_UpdateTooltip'
[string "@FrameXML\UnitFrame.lua"]:452: in function <FrameXML\UnitFrame.lua:451>
```

### DIFF
--- a/totalRP3/modules/register/main/register_tooltip.lua
+++ b/totalRP3/modules/register/main/register_tooltip.lua
@@ -524,10 +524,10 @@ local function writeTooltipForCharacter(targetID, _, targetType)
 		local lineRight;
 		local race = UnitRace(targetType);
 		local class = localizedClass;
-		if info.characteristics and info.characteristics.RA and info.characteristics.RA:len() > 0 then
+		if info.characteristics and info.characteristics.RA and info.characteristics.RA ~= "" then
 			race = info.characteristics.RA;
 		end
-		if info.characteristics and info.characteristics.CL and info.characteristics.CL:len() > 0 then
+		if info.characteristics and info.characteristics.CL and info.characteristics.CL ~= "" then
 			class = info.characteristics.CL;
 		end
 		if getConfigValue(CONFIG_CROP_TEXT) then
@@ -573,7 +573,7 @@ local function writeTooltipForCharacter(targetID, _, targetType)
 	-- CURRENTLY
 	--*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
-	if showCurrently() and info.character and (info.character.CU or ""):len() > 0 then
+	if showCurrently() and info.character and info.character.CU and info.character.CU ~= "" then
 		tooltipBuilder:AddLine(loc.REG_PLAYER_CURRENT, 1, 1, 1, getSubLineFontSize());
 
 		local text = strtrim(info.character.CU);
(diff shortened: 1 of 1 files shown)


totalRP3/modules/register/main/register_tooltip.lua
```diff
@@ -497,12 +497,12 @@ local function writeTooltipForCharacter(targetID, _, targetType)
 
 	if showFullTitle() then
 		local fullTitle = "";
-		if info.characteristics and info.characteristics.FT and info.characteristics.FT:len() > 0 then
+		if info.characteristics and info.characteristics.FT and info.characteristics.FT ~= "" then
 			fullTitle = info.characteristics.FT;
 		elseif UnitPVPName(targetType) ~= targetName then
```
Connor-Macleod, line 502: Would changing this to 
```lua
elseif UnitPVPName(targetType)  and UnitPVPName(targetType) ~= targetName then
```
not be more consistant?

Meorawr: It'd _work_ but it's an extra call to the same C function, and the end result is the same.

If we wanted to go down the rabbit hole of what's probably best, we'd not default the title to an empty string and instead _expect_ a nil value if the full title weren't available, noting that it can't ever be an empty string in either branch anyway (API bugs notwithstanding):

```lua
local fullTitle;

if info.characteristics and info.characteristics.FT and info.characteristics.FT ~= "" then
    fullTitle = info.characteristics.FT;
elseif UnitPVPName(targetType) ~= targetName then
    fullTitle = UnitPVPName(targetType);
end

if fullTitle then
```

Alas given the read-only nature of CurseForge in the next couple of weeks I'd rather not take the chance on a risky fix such as the above in the not-entirely-improbable case that UnitPVPName would return an empty string in some circumstance; especially as this isn't a regression in the most recent release but rather a previously unseen (and not likely easily reproduced) issue.

Connor-Macleod: counterpoint: Empty strings are not causing the error, nil values in `fullTitle` are. which admittedly, line 505 fixes.

I was mostly suggesting for the sake of consistency between the branches of the if. because at the moment we've got three outcome states where IMO we should be trying to have only two.
Either `""` and `"string"` or `nil` and `"string"`

Solanya (approved):

PR #479, opened 2020-12-27
Update links from Curse to CurseForge in README

DJScias:
Curse site doesn't work anymore, it's transitioned to Curseforge.

Self-explanatory commit. Nothing major, and feel free to close it and do it yourself as I suspect it was planned as-is but I just happened to be "first".


Solanya:
Good point, thanks for the fixes, check is borked but I'll ignore that ^^

PR #482, opened 2021-01-07
Secondary Pet Companions

### LINKED ISSUE #52: Additional pet support
>BrotherNero, comment #700
>Greetings. I play an Unholy Death Knight and we have a talent called "All Will Serve" which summons a skeleton that follows us around and shoots arrows at people we don't like.  Blizzard has named this guardian "Risen Skulker" when his proper name should be "Bob". When I try to assign a profile to him, it does not allow me to do so. When you get a chance, could you please fix this, if it is at all possible? Thanks for your time.

Meorawr:
Closes #52 as best as we can support it within the limits of the API.

This implements the required changes to support "secondary" pets as exposed through the Beast Mastery talent [Animal Companion](https://www.wowhead.com/spell=267116/animal-companion).

This required surprisingly little to actually get fundamentally working; the first roadblock was our handling of unit type detection filtering out units that weren't the players' primary pet - this was solved by just checking if the unit is controlled by a player, has a Pet GUID, and whether or not the unit is actually owned by us. This allowed local tooltips to work on both a primary and secondary pet.

For comms, an additional set of query indexes were added for secondary pets to send information on the players' first stable slot pet through, as the stable master APIs can be used anywhere in the game world. With these changes, I'm able to verify that a second account can receive and show the profiles of two pets from the same hunter simultaneously.

Finally to ease actually assigning a companion profile to a hunter pet a new browser was added that'll show only for Hunters to allow them to select a pet via the stable master APIs that a companion profile should be bound to.

image
image
image


Meorawr:
Made a few tweaks to the warning messages shown in the tooltips for an upcoming commit to add clarity and cover the case where a pet hasn't been renamed from the family default.

image
image

Meorawr:
Pretty much happy with it now.

Made a few tweaks in the last set of commits, so here's some updated screenshots showing the main ones:

* Pets assigned to a profile will show a lock above their icon - couldn't find a better texture that says "I'm in use" other than a warning symbol, but that looked pretty bad at the size of the buttons.
* Pets will no longer be shown at all if they're already bound to the same profile that the user is adding a pet to. This fixes that weird warning message saying "this pet is bound to profile X" when you're adding a new pet to the same profile.
* A bunch of strings were shortened.

image
image

totalRP3/modules/register/companions/register_companions_main.lua
```diff
@@ -268,7 +268,7 @@ function TRP3_API.companions.player.getCurrentBattlePetQueryLine()
 end
 
 
-function TRP3_API.companions.player.getCurrentPetQueryLine()
+function TRP3_API.companions.player.getCurrentPrimaryPetQueryLine()
```
Ellypse, line 271: Just in case, could we make an alias to the previous method name?

Ellypse, line 271: Just in case, could we make an alias to the previous method name?

totalRP3/core/impl/Enums.lua
```diff
@@ -50,3 +50,13 @@ AddOn_TotalRP3.Enums.ROLEPLAY_EXPERIENCE = {
...
 	VOLUNTEER = 3,
 };
+
+-- UNIT_TYPE describes the base type of a unit, such as whether it represents
+-- a character or pet. These replace the old TRP3_API.ui.misc.TYPE_* constants.
+AddOn_TotalRP3.Enums.UNIT_TYPE = {
+	CHARACTER = "CHARACTER",
+	PET = "PET",
+	BATTLE_PET = "BATTLE_PET",
+	MOUNT = "MOUNT", -- TODO: Do we want mount in here? Seems wrong.
```
Ellypse, line 60: I don't remember the logic behind it so I would suggest to not touch it 😅

Ellypse (approved):

Ellypse (approved):

Solanya (approved):

PR #487, opened 2021-01-26
Add Pronouns Preset to Additional Information

Meorawr:
This implements full support for the pronouns field as a new preset in the additional information section and has been tested with basic TRP and MSP comms. The preset is mapped to the new `PN` field in MSP.

Included is also a small set of bugfixes for invalid icons being used in Classic when receiving other info fields such as the house name.

I couldn't think of an icon any more perfect than this, quite frankly - the only other option was Yet Another Generic Scroll:

Behold His Majesty

Screenshot of the new preset in the dropdown:
New preset in editor dropdown

Screenshot of the new preset in action:
New preset in viewer

Screenshot showing receipt of the `PN` field via MSP on a separate account:
Receipt of PN field via MSP

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -224,6 +224,7 @@ stds.wow = {
 		"CreateTextureMarkup",
 		"CreateVector2D",
 		"FCF_GetCurrentChatFrame",
+		"FindInTableIf",
 		"GameTooltip_AddNormalLine",
 		"GameTooltip_SetDefaultAnchor",
 		"GameTooltip_SetTitle",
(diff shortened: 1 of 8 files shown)


Ellypse:
When we talked about this I suggested having the field shown in the tooltip on hover if there is a value. I think it makes sense for this information to be available ASAP in interactions and having it only when opening the full profile might be too hidden. But it might also crowd the tooltip even more than it already is. What do you think?

Ellypse (dismissed):

Meorawr:
Putting it at the bottom of the tooltip seems to make most sense - anywhere else doesn't really fit. Chose the orange colour as used with the realm/guild lines since I think it looks a bit more consistent as a single-line field to have it that way, compared to the IC/OOC which aren't. Text in the tooltip display of the field crops to 20 characters, which should be enough for any imaginable subject/object forms and allows room for the possessive too.

image

Ellypse:
That looks perfect to me, thanks 😃

Ellypse (approved):

PR #505, opened 2021-03-09
Companion profile support for Classic

### LINKED ISSUE #370: [Classic] Companion profile cannot be set on "non combat pets"
**Describe the bug**
A companion profile cannot be set for a non-combat pet (summoned from cages/eggs/etc.) in **Classic.**

**To Reproduce**
Steps to reproduce the behaviour:
1. Summon a pet, like [this cute thing](https://classic.wowhead.com/npc=9662/sprite-darter-hatchling)
2. Create a companion profile
3. Click the **`Binds`** button.
4. Click *Target*.
5. Error saying "Target is not a valid pet, mage elemental, minion, renamed battle pet" pops up and nothing happens.

**Expected behaviour**
Be able to set a story and fluff text for the companion.

**Screenshots or error message**
image

**Version of the add-on:**
Total RP 3 1.6.7.1 (build 79) **Classic WoW version**

**Did you try having Total RP 3 as the only enabled addon and everything else (especially something like ElvUI) disabled?**
Yes

Meorawr:
Closes #370.

This implements the bulk of the work for companion profiles (both as non-combat pets and mounts) for Classic, and should be extensible enough to also work on the upcoming TBC client by just adding new data to the companion tables.

The bulk of the data and shims for the PetJournal/MountJournal APIs are present in the new CompanionData.lua script which handles all the heavy lifting, and a few minor redirections were placed in the other files to either operate from the C_ APIs where available or from our static data where unavailable. This is mostly to avoid adding more "is_classic" tests since that's going to be annoying to deal with when TBC rolls around.

Additional mounts and companion pets can be added to the `TRP3_CompanionPetData` and `TRP3_CompanionMountData` tables, which will be filtered on client load to remove any references to spells or items that don't exist - so no effort should be needed for manually flagging things as belonging to individual expansions.

**Non-Combat Pets**

The data for non-combat pets consists of the species ID, a summon item ID, a summon spell ID, and the creature ID that's summoned by the spell. The species ID for each pet in our data matches the species ID present on retail as exposed by the API there.

To work around our custom name requirement we don't use the name of the summoned pet creature since that isn't something we can query from the API - instead we use the name of the item that summons the creature. This does mean that if the user has three [cat carriers](https://www.wowhead.com/item=8489/cat-carrier-white-kitten) of the same type they'll all be linked to the same pet, but there's no realistic way to work around that.

As the API doesn't expose any way whatsoever for knowing what pet is summoned, we hack that knowledge in by listening for successful spell casts from the player and mapping the spell IDs from those back to known companion pets. The species ID of the summoned pet is stored in a temporary custom CVar ("totalRP3_SummonedPetID") which allows us to persist this data across UI reloads and logouts, however it isn't needed for the logout case as companion pets aren't resummoned automatically upon logging back in. One caveat is we don't know when a pet is dismissed, but this should be fine since other players won't be able to see the unit anyway in such a case.

**Mounts**

These are a lot simpler because we don't need too much API magic to get these working. The only annoying bit is knowing which mount is summoned; to not kill the client each time the browser is opened (or each time we need to get the current mount line for the exchange) this information is cached and only requeried when `UNIT_AURA` fires for the player.

As with non-combat pets, the mount ID present in the data also maps back to the correct mount IDs present on the retail client.


Meorawr:
Note that the diff for this is a bit messy as I've included the changes from #506 as it requires some tweaks with this branch; keep that in mind when reviewing, or merge in #506 first and I'll rebase.

Meorawr:
External reminder for myself for when this gets merged in that it might be worth doing a pass over the CF localizations for the Classic project to rename any references to battle pets to "non-combat pet", and to fix the one localization string that says they need renaming as this isn't relevant there.

Meorawr:
Rebased now that #506 was merged to clean up the diff.

Solanya (approved):

Solanya (approved):

PR #506, opened 2021-03-11
Refactor all Classic checks for icons

Meorawr:
This refactors all our `is_classic` checks when dealing with icon names and file paths to instead use a global table (`TRP3_InterfaceIcons`) which, when queried at runtime, associates an ID with a valid file name for an icon that's supported by the currently loaded client. All usages of icon names/paths in the addon that I've been able to find have been moved to this table.

The icons are defined as a mapping of identifiers to an array of icon file names to be checked on startup. For the first valid icon name that returns a file ID (via `GetFileIDFromPath`), we associate that identifier with the name and use it for all future lookups against the table. This allows us to not hardcode any information whatsoever about which client supports what icons, and solves the longstanding problem of having to duplicate icon file names/paths in various areas of the code.

**Note:** This work is also present in the Classic companions PR (#505) since that changes some areas which conflict; for review purposes I've put the majority of the work in this PR separately however.

Overall this significantly reduces our `is_classic` check count and should make TBC support much easier in the future - we're down to about 19 references now which are all guarding actual blocks of code instead of data.

### DIFF
--- a/totalRP3/modules/register/characters/register_characteristics.lua
+++ b/totalRP3/modules/register/characters/register_characteristics.lua
@@ -599,37 +599,37 @@ local MISC_PRESET = {
 	{
 		NA = loc.REG_PLAYER_MSP_HOUSE,
 		VA = "",
-		IC = TRP3_InterfaceIcons.MiscInfo_House,
+		IC = TRP3_InterfaceIcons.MiscInfoHouse,
 	},
 	{
 		NA = loc.REG_PLAYER_MSP_NICK,
 		VA = "",
-		IC = TRP3_InterfaceIcons.MiscInfo_Nickname,
+		IC = TRP3_InterfaceIcons.MiscInfoNickname,
 	},
 	{
 		NA = loc.REG_PLAYER_MSP_MOTTO,
 		VA = "",
-		IC = TRP3_InterfaceIcons.MiscInfo_Motto,
+		IC = TRP3_InterfaceIcons.MiscInfoMotto,
 	},
 	{
 		NA = loc.REG_PLAYER_TRP2_TRAITS,
 		VA = "",
-		IC = TRP3_InterfaceIcons.MiscInfo_Traits,
+		IC = TRP3_InterfaceIcons.MiscInfoTraits,
 	},
 	{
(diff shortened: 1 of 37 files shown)


totalRP3/resources/InterfaceIcons.lua
```diff
@@ -0,0 +1,249 @@
...
+	MiscInfoPronouns  = { "vas_namechange", "inv_scroll_08" },
+	MiscInfoTattoos   = { "inv_inscription_inkblack01", "inv_potion_65" },
+
+	--
+	-- Personality Trait Icons
+	--
+
+	TraitAltruistic    = { "inv_misc_gift_02" },
+	TraitAscetic       = { "inv_misc_food_pinenut", "inv_misc_coin_05" },
+	TraitBonViviant    = { "inv_misc_food_99", "inv_misc_coin_02" },
```
Solanya, line 163:
```suggestion
	TraitBonVivant    = { "inv_misc_food_99", "inv_misc_coin_02" },
```

totalRP3/modules/register/characters/register_characteristics.lua
```diff
@@ -1359,76 +1358,76 @@ local function initStructures()
...
+			LI = TRP3_InterfaceIcons.TraitCautious,
+			RI = TRP3_InterfaceIcons.TraitImpulsive,
 		},
 		{
 			LT = loc.REG_PLAYER_PSYCHO_Acete,
 			RT = loc.REG_PLAYER_PSYCHO_Bonvivant,
-			LI = is_classic and "INV_Misc_Coin_05" or "INV_Misc_Food_PineNut",
-			RI = is_classic and "INV_Misc_Coin_02" or "INV_Misc_Food_99",
+			LI = TRP3_InterfaceIcons.TraitAscetic,
+			RI = TRP3_InterfaceIcons.TraitBonViviant,
```
Solanya, line 1424:
```suggestion
			RI = TRP3_InterfaceIcons.TraitBonVivant,
```

Solanya (changes requested):

Solanya (approved):

PR #507, opened 2021-03-12
CI Updates for Wago addons support

Meorawr:
This changeset does the following:

  - Adds the Wago project ID to our packager runs.
  - Separates the workflow for a release into two, where one handles prerelease versions. This solves us having to modify the packager to not have it upload tagged alphas/betas to WoWI.
  - Removes the renaming of the Classic .zip as this was only really relevant for one site, and causes pain if we have to go and package the addon three times.

With these changes and the necessary PRs merged into the packager, we're able to remove the need for our packager fork entirely and can use the upstream one unaltered.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,6 @@
 ---
 name: Pull Request
-on:   pull_request
+on: pull_request
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


Meorawr:
image

Today the forces of good have triumphed over the forces of evil, or something.

Meorawr:
Upstream packager has Wago support now.

Solanya (approved):

PR #510, opened 2021-03-14
Diet reimplementation of the nameplates module

### LINKED ISSUE #18: Default nameplates customizations
Since patch 7.3 is bringing a native "name only" mode for default friendly nameplates, it is time for Total RP 3 to customize the default nameplates. 

The customizations we would want :

* Replace character name with custom RP names
* Replace class color with custom class color
* Opt in option to show player icon in front of the name
* Opt in option to show at-first-glance slots on nameplates
* All those things should work on player nameplates and companion nameplates (hunter, warlock pets, since battle pets doesn't get nameplates)
* For player, opt on option to also show the name of the mount below the player, with all the additional features mentioned avives.

Meorawr:
Closes #18.

This cuts out a lot of the features that had accumulated into the previous implementation in favor of a leaner approach that we can build upon based upon user requests after this launches. Consider this to be more an MVP more than a feature-complete module.

In terms of what's implemented:

 - Blizzard nameplates
 - Kui nameplates

ElvUI support has been dropped for now since implementing it has complications and I'd rather get user feedback first before committing to any real work in that area.

For the addons that are supported, the nameplate customizations are roughly the same as before with the notable exception of glance bar support. This might be re-added at a future date if it's requested, but for now it adds a lot of code and I'd rather ship without it.

Blizzard nameplates with the default configuration of the module:

image

With everything enabled:
image

Available options:
image

At the moment there's nothing to hide nameplates on units that don't have profiles - I _may_ implement this quickly but I might also prefer keeping that for a future release since it seems a risky one to get supported uniformly on all nameplates.


Meorawr:
Quickly did the Kui stuff tonight 'cause why not.

image

Meorawr:
Largely ready for review and merge at this point, not planning any more major changes aside from bugfixes if they come up.

[x] Finished implementing the Kui customizations.
[x] Finished implementing the Blizzard customizations.
[x] Added the "hide non-roleplay units" setting; I'd consider this beta quality at best and it's disabled by default. It might cause unknown issues, but I figured having it was important enough in terms of retaining parity with the existing Kui module. This is also supported on the Blizzard nameplates.

Meorawr:
Rebased to resolve conflicts and save the planet.

Meorawr:
Rebased to fix locale conflicts with development.

Solanya (approved):

PR #520, opened 2021-03-19
Nameplate Tweaks 3

Meorawr:
YEP

- [x] Fixed an extremely rare potential issue with Kui's fader plugin if we were modifying the visibility of nameplates.


Meorawr:
Not aware of any bugs aside from the Kui window resize issue, so flagging as ready for review + merge into the final release.

Solanya (approved):

PR #522, opened 2021-03-21
Apply custom bar color to personality trait names

Solanya:
When an attribute has a custom color selected, the color will now be applied to the attribute name as well. Contrast is improved by default to make the trait names more readable on the characteristics page dark background.


Meorawr (approved):
Maybe update the changelog in this PR too.

Meorawr (approved):

PR #530, opened 2021-03-24
Burning Crusade Companion IDs

Meorawr:
Work in progress, need to go through and make sure this list is correct when addons are enabled and I have access.


Meorawr:
Tested through means best left unspoken; the updated list works fine on both Classic and TBC as-is.

Solanya (approved):

PR #539, opened 2021-03-30
Remove Roleplay Language (LC) field

Meorawr:
This field was initially conceived to solve the lack of language specific RP servers in Classic. This problem ended up later solving itself and there's been a few issues with this field since, including the limited set of languages (since we based it off client locale), as well as the display of the information in tooltips where flags were chosen to save space.

Solutions to these problems could all be found but at this point in time we feel the field no longer serves any purpose and could probably just be removed outright, allowing players to put this information into their OOC texts where desired.


Solanya (approved):

Solanya (approved):

Meorawr:
Note I've tagged this as a 2.4 milestone thing 'cause I think it'd be weird for this field to just disappear in 2.3.3, plus we can get some feedback about the change potentially before rushing it out the door.

Solanya (approved):

PR #544, opened 2021-04-06
Protect "[TRP3:...]" links from chat transforms

### LINKED ISSUE #540: Profile chat links mangled if profile name contains parentheses
Creating a profile chat link where the profile name contains parentheses results in a bit of a mess. My guess is this is the OOC text colouring screwing things up since we don't send those as links but rather as normal strings that get reconstituted back into links upon receipt.

image

Meorawr:
Fixes #540. Previously a chat link that contained any OOC, speech or emote pattern matches would have their inner text mangled and replaced with the appropriate color sequences, as it was possible for the message handler that processes these to be installed after the one used for text color processing.

This rewrites the message protection code; the old approach would first find the patterns it was interested in and replace them with a temporary pre/postfix marker, it'd then check the things it didn't want to mess up and remove those markers, and then replace the remaining markers with the intended color sequence.

Instead, this new approach changes the order of operations - we find things that should be protected (hyperlinks, declensions, and our own textual link format) and replace those with a uniquely generated escape sequences, borrowing from Blizzards' "|K" protected strings, and store the replacements in a table. Message processing then occurs as normal after that with the protected strings stashed away, and then they're restored as the last step of the chat handler. This approach ends up being safer and more maintainable, as well as more efficient as we need less closures and string replacements.

Proof that it all still works - I also manually verified that we don't screw up `|H` or declensions separately.

image

### DIFF
--- a/totalRP3/modules/chatframe/chatframe.lua
+++ b/totalRP3/modules/chatframe/chatframe.lua
@@ -369,7 +369,7 @@ local function ProtectMessageContents(message)
 
 	message = string.gsub(message, "|H.-%|h.-|h", ProtectString);
 	message = string.gsub(message, "|3%-%d+%b()", ProtectString);
-	message = string.gsub(message, "%[TRP3:[^%]]+%]", ProtectString);
+	message = string.gsub(message, "%b[]", ProtectString);
 
 	return message, replacements;
 end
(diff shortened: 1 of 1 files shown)


totalRP3/modules/chatframe/chatframe.lua
```diff
@@ -384,70 +396,34 @@ local function detectEmoteAndOOC(message, NPCEmoteChatColor, isEmote)
 		NPCEmoteChatString = NPCEmoteChatColor:GetColorCodeStartSequence();
 	end
```
Meorawr, line 397: Unrelated but might be worth cleaning up one day; with the `|r` change in 9.0.1 to pop colours individually rather than reset them I think this code as-is would lead to a situation where, for NPC talk, we'd be pushing the emote colour onto the colour stack anew each time we finish an OOC/Speech substitution and if there were enough of them we'd hit the limit and break further colours in the message.

Solanya (approved):

PR #547, opened 2021-04-14
9.1: Update power type global strings

Meorawr:
Patch 9.1 seems to be removing the `MAELSTROM` global string entirely which breaks us for that patch. There exists an equivalent in the form of `POWER_TYPE_MAELSTROM`, which exists on all clients so we'll use that instead.

For all other power types that have corresponding `POWER_TYPE_*` global string constants these have also been changed, in case Blizzard makes more changes of this kind in the future.


Meorawr:
Check failing seem to be a GH actions/luarocks issue.

Meorawr:
Changing this to 2.3.3 because it's a small change and, as it's breaking in 9.1, shipping it now means we don't have a broken addon on patch launch day for people who are slow to update.

Solanya (approved):

PR #549, opened 2021-04-24
Profile sanitization improvements

Meorawr:
This removes the option to disable sanitization of profiles and adds a small tweak to also strip protected string sequences (|K). Some more work will be done here to improve sanitization of existing fields as well, but that's pending for now.


Solanya (approved):

Meorawr:
It's been 13 ~~years~~ months

PR #562, opened 2021-05-08
Rewrite MSP integration

Meorawr:
This rewrites our MSP integration from scratch to clean up the code, fix a few bugs, and reduce the risk of us sending weird data via MSP in the face of unlawful profile editing.

The key changes are:

- Implemented the use of the "dataload" callback to preload data for MSP profiles, which may cut down on comms bandwidth for those users as we won't necessarily request their full profiles on each fresh	UI session.
- Swapped to using the "updated" callback when processing incoming data from MSP users; this triggers faster profile updates as we receive fields incrementally - so we don't need to wait for an MSP user to send us their "About" text before we're able to get their name.
- Added logic to not send potentially disallowed or invalid things in fields - these include color codes where not permitted among other things.
- Cleaned up the logic for the hacks that had since accumulated, primarily around deferring requests until the VA field has arrived.
- Initial experimentation for the "VI" field as a profile identifier string. For now this is of the form "\<addon>:\<profile id>".

I've tested these changes over the past couple of days by force disabling TRP's comms protocol locally and exclusively using the MSP ones and haven't seen any issues that occur in regards to sending a full player profile nor receiving a large number of them - all data received and displayed in the UI looks correct. I've also done a quick smoke test to confirm that sending a full profile with everything set doesn't break when displayed in MRP.


Meorawr:
Will close this PR and re-open it at some point during DF prep.

PR #598, opened 2022-02-22
Added StopSound fadeout duration

Solanya:
Doing it now while we're making a new version, added the fadeout duration parameter in the stopSound functions so I can use it later in Extended.

Wow.


Meorawr (approved):

Katoriehooves (commented):
Hello I have reviewed your changes and approve

PR #600, opened 2022-02-25
Improve TextArea scroll behavior when editing

Meorawr:
This modifies the TextAreaBase template to use Blizzards' ScrollingEdit APIs which behave slightly better than our existing code with respect to scrolling-down a scrollframe when the user is editing text toward the lower end of the view.

The main changes this results in is that:

- Scrolling is no longer dependant upon the caret being at the end of the input field, and will occur if - for example - you're inserting text before it that would cause the new caret position to be outside the viewable field.
- Scrolling will leave a buffer region below the caret, so you've got  a few visible lines for the mid-field edit case.

Any usages of the TRP3_TextAreaBase or TRP3_TextArea templates in extended will need to be checked to see if they're providing an OnLoad script handler, as one has been added to the template to handle resizing of the child editbox component.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -386,6 +386,9 @@ stds.wow = {
 		"RemoveChatWindowChannel",
 		"ResetCursor",
 		"Saturate",
+		"ScrollingEdit_OnCursorChanged",
+		"ScrollingEdit_OnLoad",
+		"ScrollingEdit_OnTextChanged",
 		"SecondsToClock",
 		"SendChatMessage",
 		"SetCursor",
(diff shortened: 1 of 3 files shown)


Meorawr:
Quick video to demonstrate the effects of the changes when editing the end and middle of text in an About page; I've also thrown in a quick test to verify that resizing the editbox doesn't break things unexpectedly.

https://user-images.githubusercontent.com/287102/155772219-5c74cbdd-88a8-4d1f-9948-fcb7edf960fc.mp4

Solanya (approved):

PR #601, opened 2022-03-26
Fix for target frame ElvUI skin

Solanya:
9.2 (or was it 9.1.5?) changed tooltips to use NineSlice, and ElvUI therefore changed their tooltip skinning function that we used to skin both the TRP tooltip and the target frame. This in turn broke the target frame skinning, so I restored the old skinning function to use for target frame skin.


Meorawr (approved):
I'm just going to assume this works.

PR #605, opened 2022-05-30
Removed April Fools 2022

Solanya:
They've ooked their last ook and it's now time to leave.


Solanya:
I've left the global serious_day because I add it back every year and it might as well stay around.
Removed the setting though because it could need changing every year.

Meorawr (approved):

PR #607, opened 2022-06-13
Tooltip text color customizations

### LINKED ISSUE #606: Color settings for tooltips
While the idea of letting everyone choose what colors they want their tooltips to be shown for everyone around is a bit much to consider at the moment, it could be nice to have the option to decide what colors tooltip fields use locally. We could consider offering to change separately:
- Full title
- Currently text
- OOC text
- Sections titles (and other default white fields)
This will likely come with the unfortunate end to the symbol abuse through sanitization. It lived long enough.

Meorawr:
Fixes #606. This adds a new options to the tooltip settings that gives users a defined way to customize the colors of texts in the tooltip. These are split into three categories - "Title" (full titles), "Main" (most text such as race/level, as well as any header-like texts) and "Secondary".

The default tooltip settings keep everything remaining (roughly) the same as shown below.

Default tooltip settings

Customized, we can make it look like this.

Customized tooltip settings


Meorawr:
Per #606 I'll probably add the full title and section headers options into this PR too.

Solanya (dismissed):

Solanya (approved):

PR #612, opened 2022-07-04
Bump TOC version for 9.2.7

Meorawr:
Whether or not the patch actually retains this interface version bump when launched we don't yet know; a similar thing happened in 8.3.7 where the TOC increased on the PTR but was reset prior to launch.


Ellypse (dismissed):

Solanya:
Piggybacking off this one to include Wrath TOC

Solanya (approved):

PR #621, opened 2022-08-23
Sanitization improvements

Solanya:
Breaking kneecaps all around by removing unwanted modifications in a few more fields.

### DIFF
--- a/totalRP3/core/impl/utils.lua
+++ b/totalRP3/core/impl/utils.lua
@@ -448,14 +448,13 @@ local escapes = {
 }
 function Utils.str.sanitize(text)
 	if not text then return end
-	local originalText = text;
-	for k, v in pairs(escapes) do
-		text = text:gsub(k, v);
-	end
-	-- Looping for nested tags
-	if originalText ~= text then
-		text = Utils.str.sanitize(text)
-	end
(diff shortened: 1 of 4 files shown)


totalRP3/core/impl/utils.lua
```diff
@@ -441,10 +448,15 @@ local escapes = {
...
+	local originalText = text;
 	for k, v in pairs(escapes) do
 		text = text:gsub(k, v);
 	end
-	return text;
+	-- Looping for nested tags
+	if originalText ~= text then
+		text = Utils.str.sanitize(text)
+	end
+	return text
```
Meorawr, line 459: I'd feel better if this were a loop rather than recursion due to stack overflow potential, even if exceptionally unlikely due to string length limitations.

```suggestion
function Utils.str.sanitize(text)
	if not text then return end
	-- Repeat until nested tags are eliminated.
	repeat
		local originalText = text;
		for k, v in pairs(escapes) do
			text = text:gsub(k, v);
		end
	until originalText == text;
	return text
```

Meorawr (approved):
Minor nitpick aside it looks good.

Meorawr (approved):

PR #640, opened 2022-10-22
Add "/trp3 set" command

Meorawr:
For the sake of padding the changelog this PR implements a new `/trp3 set <field> <value...>` command that can modify fields of the current profile directly from the chatframe or in macros. Macro conditionals can be used with the value segment to further customize the assigned values. Additionally a custom macro conditional framework is added with support for `[ooc]`, `[ic]`, `[rpstatus:<status>]`, `[location:<zone name>]`, and `[profile:<profile name>]` conditionals.

The following fields are supported:
  - class &lt;text...&gt;
  - classcolor &lt;#rrggbb&gt;
  - currently &lt;text...&gt;
  - firstname &lt;text...&gt;
  - fulltitle &lt;text...&gt;
  - icon &lt;icon name&gt;
  - lastname &lt;text...&gt;
  - oocinfo &lt;text...&gt;
  - race &lt;text...&gt;
  - title &lt;text...&gt;

These fields were chosen as they're basically the "main" ones on most tooltips and are probably a good starting point for this feature. For macro conditionals, the conditions come after the field name and will select the value. These are only processed if the first character of the value is an "[" so as to avoid breaking strings on semicolons.

Running `/trp3 set` or `/trp3 set help` will print help text showing the full command and supported fields.

image

If the current profile is a default one, no modifications will be allowed. Additionally there is validation to ensure that a field is known and an appropriate error printed if not.

image


totalRP3/Core/Globals.lua
```diff
@@ -141,4 +141,5 @@ TRP3_API.globals.build = function()
 	end
 end
 
-TRP3_API.globals.addon = LibStub("AceAddon-3.0"):NewAddon(TRP3_API.globals.addon_name);
+TRP3_Addon = LibStub("AceAddon-3.0"):NewAddon("TRP3");
```
Solanya, line 144: Assuming this won't be an issue but `addon_name = "Total RP 3"` so it changes the behaviour

Meorawr: I'd be surprised if it breaks anything else - the addon object is only really used for our bootstrap logic. I just updated it so it looks a bit more sensible in `/dump` output, since Ace concats submodule names to it via an underscore and "Total RP 3_Foo" looks odd.

Solanya (approved):

PR #644, opened 2022-10-24
Use C_TooltipInfo APIs for companion owner names

Meorawr:
A future build of 10.0.2 will add a new UnitOwner tooltip line type that provides the GUID of the owning unit as part of its payload. We can use this to extract owner names and realm, thus removing the need to de-format localization strings and deal with locale-specific quirks like declensions.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -294,6 +294,28 @@ stds.wow = {
 			},
 		},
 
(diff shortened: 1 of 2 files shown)


Meorawr:
This is implemented in the latest beta build and works fine.

Solanya (approved):

PR #657, opened 2022-11-02
Plater Nameplates Support

Ghostamoose:
Opening PR now for collaboration.

If anyone is interested in testing, everything should be functional, and non-destructive. Note that you have to use the default Plater profile (for now) for this to work.

I require feedback.


Ghostamoose:
**Tasks**

Essential:
- [ ] Support for the higher-contrast setting
- [ ] Config localization
- [ ] Update changelog

Nice-to-haves:
- [ ] Better icon placement
- [ ] Friend coloring
- [ ] Guildmate coloring
- [ ] Support for non-default Plater profiles

"Go-long" ideas:
- [ ] OOC (character) names
- [ ] Customizable icon position (i.e. left, right, top, bottom)

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,355 @@
...
+	hooksecurefunc(nameplate.unitFramePlater, "UpdateHealthColor",
+		function(...) return self:UpdateNamePlateHealthBar(...); end);
+
+	hooksecurefunc(Plater, "UpdateUnitName",
+		function(...) return self:UpdateNamePlateNameText(...); end);
+
+	do
+		-- Icon widget.
+		if nameplate.actorType == "friendlyplayer" then
+			local iconWidget = nameplate:CreateTexture("TRP3PlaterIcon", "ARTWORK");
```
Meorawr, line 77: Drop the "TRP3PlaterIcon" name from this call and replace it with a nil; as-is every nameplate that runs through this function is going to create textures all named the same thing.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,355 @@
...
+			iconWidget:SetPoint("RIGHT", nameplate.CurrentUnitNameString, "LEFT", -4, 0);
+			iconWidget:Hide()
+
+			nameplate.unitFramePlater.TRP3PlaterIcon = iconWidget;
+		end
+	end
+	do
+		-- Title text widget.
+		if nameplate.actorType == "friendlyplayer" then
+			local titleWidget = nameplate:CreateFontString("TRP3PlaterTitle", "ARTWORK");
```
Meorawr, line 88: As above for the title fontstring.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,355 @@
...
+		else
+			return name
+		end
+	else
+		return name
+	end
+end
+
+function TRP3_PlaterNamePlates:AppendFullTitleToNameString(name, fullTitle)
+	return name .. "\n" .. "<" .. fullTitle .. ">"
```
Meorawr, line 116: From recollection neither of the existing decorators for nameplates surround the full title text in "<>". I think I'd gone with that due to the (very low) default cropping limit for nameplates and to cut down on visual noise. Not necessarily _opposed_ to Plater shoving them in, but might be worth seeing how they look without.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,355 @@
...
+	if not self:CanCustomizeNamePlate(nameplate) then
+		return;
+	end
+
+	local displayInfo = self:GetUnitDisplayInfo(nameplate.unitFramePlater.namePlateUnitToken);
+
+	if displayInfo and displayInfo.shouldHide then
+		if nameplate:IsShown() then
+			nameplate:Hide();
+		end
```
Meorawr, line 283: > Bob: Hey can you check if the door is open?
> Alice: Sure, lemme go do that.
> Bob: Yeah, the door is open.
> Alice: Can you go back and close it please?
> *Alice grumbles.*

Just tell the nameplate to hide, no need to ask.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,355 @@
...
+	else
+		if not nameplate:IsShown() then
+			nameplate:Show();
+		end
```
Meorawr, line 287: As above, except this time imagine Alice pulling out something sharp.

totalRP3/Locales/enUS.lua
```diff
@@ -632,7 +632,7 @@ The Kui |cff9966ffNameplates|r module adds several Total RP 3 customizations to
 	["NAMEPLATES_CONFIG_ICON_SIZE"] = "Icon size",
 	["NAMEPLATES_CONFIG_ICON_SIZE_HELP"] = "Configures the size of icons displayed on nameplates if the |cffffff00Show icons|r option is enabled.",
 	["NAMEPLATES_CONFIG_MENU_TITLE"] = "Nameplates",
-	["NAMEPLATES_CONFIG_PAGE_HELP"] = "Please note that only |cff449fe0Blizzard|r and |cff9966ffKui|r nameplates are currently supported. Refer to the help tip on each setting below for additional information.",
+	["NAMEPLATES_CONFIG_PAGE_HELP"] = "Please note that only |cff449fe0Blizzard|r, |cff9966ffKui|r, and |cffa8deffPlater|r nameplates are currently supported. Refer to the help tip on each setting below for additional information.\n\nIf using |cffa8deffPlater|r nameplates, you must be using the default profile for the RP names to work as intended.",
```
Meorawr, line 635: This file is autogenerated - remove this change or you'll cause more problems than you already do >:(

totalRP3/Locales/Locale.lua
```diff
@@ -1618,6 +1618,10 @@ We are aware of a current issue on Retail causing **quest item usage from the ob
 	KUI_NAMEPLATES_MODULE_DESCRIPTION = "Enables the customization of Kui nameplates.",
 	KUI_NAMEPLATES_WARN_OUTDATED_MODULE = "The Kui |cff9966ffNameplates|r plugin for Total RP 3 has been integrated directly into the main addon.|n|nThe old plugin has been disabled automatically, and |cffffcc00we recommend that you uninstall it|r as it is no longer needed.",
 
+	PLATER_NAMEPLATES_MODULE_NAME = "Plater Nameplates",
+	PLATER_NAMEPLATES_MODULE_DESCRIPTION = "Enables the customization of Plater nameplates.",
+	PLATER_NAMEPLATES_WARN_OUTDATED_MODULE = "Plater |cff9966ffNameplates|r is outdated.",
```
Meorawr, line 1623: The hex colour markup for Plater is inconsistent; in the `NAMEPLATES_CONFIG_PAGE_HELP` string it has its own unique colour but in this string it's using Kui's colour.

Meorawr (dismissed):

Meorawr:
Other misc. notes - I'm not _too_ enamoured at this point with the idea of settings that are specific to a single decorator. If icon placement becomes a thing it needs adding to all of them.

For OOC names, I couldn't make heads nor tails of that discussion. Isn't this already covered by ticking the "Disable customizations on OOC units" checkbox? I don't much like the idea of a modifier key toggling names, would need to think some more on that though.

I'm assuming friend/guild colouring is a Plater config thing that these customizations clash with, and not an option you're wanting to add in TRP.

Ghostamoose:
> Other misc. notes - I'm not _too_ enamoured at this point with the idea of settings that are specific to a single decorator. If icon placement becomes a thing it needs adding to all of them.

This seems reasonable enough, something I'll look into, but is probably just unnecessary work.

> For OOC names, I couldn't make heads nor tails of that discussion. Isn't this already covered by ticking the "Disable customizations on OOC units" checkbox? I don't much like the idea of a modifier key toggling names, would need to think some more on that though. It feels like if we had such a thing it should be a more global toggle - that it'd disable _all_ customization of both nameplates, tooltips, etc.

I could've been more specific. By OOC names I mean the non-RP character name. I understand there is _some_ value in having it easily accessible alongside the RP name, but honestly I feel like toggling nameplates gets the job done fine and probably isn't worth looking into on our end.

> I'm assuming friend/guild colouring is a Plater config thing that these customizations clash with, and not an option you're wanting to add in TRP.

Correct, the current coloring system overwrites any friend/guildmate coloring that names would have. I'm still not sure what the best way to go about this would be - changing all your friends to the same color takes away a bit of customization from that friend, and the same goes for guildmates. I was thinking of altering the outline color instead of the text color, but I think that may clash with certain profile colors. It's not an unsolvable problem, I'm just wondering what the best way to present that information would be, if at all.

Also working on your review points now. :)

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
Ghostamoose: Fixed in [12b3fa6](https://github.com/Total-RP/Total-RP-3/commit/12b3fa671de0903238b67ebcbc4e764019327305).

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
Ghostamoose: I went ahead and removed the "<>". Feels better suited for the guild name and creates unnecessary noise around the title that hurts readability. (not to mention the cropping limit)

Ghostamoose:
This PR is currently being demoted to a localization update, with the built-in Plater module being kept around, but disabled as an emergency backup plan in case something awful happens.

Meorawr:
I'd rather the file just be deleted in that case - if we need it again we can just pluck it from the history on this branch.

Ghostamoose:
The road to Plater Nameplates for TRP has been a long and bumpy one. I've created a Plater mod that holds some hope for the future, if anyone out there is reading this, I'd appreciate some extra testing hands.

https://wago.io/-IN-iZXDr

Ghostamoose:
Plater now allows developers to import their own mods from their addons, so that's what I'll be doing here. Rebuilding the Plater Nameplates module again to act as a sort of wrapper around the Plater mod.

Ghostamoose:
Re-opening for review. Everything works in it's current state, but it still needs a lot of improvement.

Which areas are the most egregious?

I know the module management side needs some help. One thing I also need to find a better solution for is the Plater import string. I doubt you'll be okay with just leaving that chonker in there as it is now.

Solanya (approved):

totalRP3/Locales/Locale.lua
```diff
@@ -1259,9 +1259,6 @@ If you wish to report %s's profile and you cannot target them you will need to o
 	REG_PLAYER_NOTES_ACCOUNT = "Common notes",
 	REG_PLAYER_NOTES_ACCOUNT_HELP = "These private notes are tied to your account and will be shared with all of your profiles.",
 
-	BINDING_NAME_TRP3_OPEN_TARGET_PROFILE = "Open target profile",
-	BINDING_NAME_TRP3_TOGGLE_CHARACTER_STATUS = "Toggle character status",
-
```
Meorawr, line 1264: Where did these go uwu?

totalRP3/Locales/Locale.lua
```diff
@@ -1310,7 +1307,7 @@ If you wish to report %s's profile and you cannot target them you will need to o
 
 	NAMEPLATES_CONFIG_MENU_TITLE = "Nameplates",
 	NAMEPLATES_CONFIG_PAGE_TEXT = "Nameplate settings",
-	NAMEPLATES_CONFIG_PAGE_HELP = "Please note that only |cff449fe0Blizzard|r and |cff9966ffKui|r nameplates are currently supported. Refer to the help tip on each setting below for additional information.",
+	NAMEPLATES_CONFIG_PAGE_HELP = "Please note that only |cff449fe0Blizzard|r, |cff9966ffKui|r, and |cffa8deffPlater|r nameplates are currently supported. Refer to the help tip on each setting below for additional information.",
```
Meorawr, line 1310: I'm fine with this string changing for now but I'm tempted to suggest moving it to the "Place new strings below here" marker so that Sol remembers to update it on CF.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,90 @@
...
+
+local TRP3_PlaterNamePlates = {} -- magic public table
```
Meorawr, line 12: May as well just drop the `local` here and not export it via `_G` later on - makes it easier to reason about on first-sight of this table.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -0,0 +1,90 @@
...
+	Plater.ForceTickOnAllNameplates();
+
+	self.PlaterMod = scriptAdded;
+end
+
+function TRP3_PlaterNamePlates:OnModuleDisable()
+	self.PlaterMod.Enabled = false;
+
+	if not TRP3_PlaterNamePlates.PlaterMod.Enabled then
+		TRP3_NAMEPLATES_ADDON = nil;
```
Meorawr, line 69: As hot-reloading is new - what would happen in the following circumstance:

1. I log in with Kui nameplates installed and enabled.
2. TRP initializes the Kui nameplates decorator, sets `TRP3_NAMEPLATES_ADDON` to "KuiNamePlates" (or whatever it was).
3. I enter the TRP settings and disable the Plater module.

In this case I'd expect that `TRP3_NAMEPLATES_ADDON` would inappropriately get unset despite having been activated for an entirely unrelated addon.

If so - it'd be wise to gate this instead on `if TRP3_NAMEPLATES_ADDON == "Plater"` I'd think.

Meorawr (changes requested):

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
Ghostamoose: I'll look into testing this more thoroughly next week but on my first set of mental gymnastics, I do think this is dangerous and will need to be looked at, will probably just implement your suggestion. It will need some additional work in there somewhere though as the mod will reset the flag to nil when it's disabled, independently of the TRP module.

tldr: needs fix and will fix next week

totalRP3/Locales/Locale.lua
Ghostamoose: I'll leave this one to Sol to resolve cause I have no idea where these went.

It does sound like these are the keys for the context menu options that were removed when edit mode was born. Were these replaced with this? 
image

Meorawr: No, those strings are for the keybindings.

totalRP3/Modules/NamePlates/NamePlates_Plater.lua
Ghostamoose: The modules would not be able to properly initialize if another one got to the bag first. In the module initialization function for both of them they check if `TRP3_NAMEPLATES_ADDON` is set and will disable themselves if it's not `nil`.

This means that while the scenario you mentioned _should_ be impossible, I am still going to implement your suggestion just to be safe.

Meorawr (approved):

PR #660, opened 2022-11-05
Allow opening profiles from player map scan pins

Meorawr:
This adds a right-click context menu that offers an "Open profile" option when clicking on pins for scanned players on the map.

If players are grouped, the entries will be coalesced into the displayed dropdown which will show (up-to) 25 "open profile" options, each showing the players' name.


Solanya (approved):

Solanya (changes requested):
I probably want to add "Right-click: ..." on the tooltip before merging

Meorawr:
I have made the changes for you, old man.

Solanya (approved):

PR #663, opened 2022-11-12
Add general taint workaround for UnitPopups

Meorawr:
When Edit Mode is entered it'll call UIDropDownMenu_AddButton for the "Layouts" dropdown without first initializing it via UIDropDownMenu_Initialize. This results in it writing buttons to the last-opened dropdown list frame, which if populated or modified by addons will have tainted 'numButtons' and 'maxWidth' fields. These are read during the UIDropDownMenu_AddButton call and spread taint back out into Edit Mode.

This workaround solves the issue by forcing DropDownList1 to be reinitialized upon being hidden. This triggers a secure assignment of constant zero to the aforementioned fields, preventing Edit Mode from tainting.

With this fix we should be able to re-enable UnitPopups by default again.


Meorawr:
Shelving as this fix breaks the "Set Focus" option and - with it - still manages to then break edit mode if you click "Set Focus" first followed by "Edit Mode".

PR #685, opened 2023-01-27
Persist name-only mode state in local settings

Meorawr:
An upcoming client build will be restoring the name-only mode cvar however won't persist it across client restarts, meaning that we'll need to store its state ourselves and restore it on load.

This commit attempts to implement that in such a way that we'll hopefully avoid confusing users and won't conflict with any other addons attempting to control the cvar.

The checkbox in the UI will represent the current effective state of name-only mode, showing a check if the cvar is enabled or not. If the state of the cvar doesn't match our locally saved state, the check will be desaturated - similar to how tristate checkbuttons work in the addon list. Clicking the checkbutton operates on our locally saved state, toggling based off that and not the effective one.

The cvar itself will only ever be enabled from us at runtime; if the user attempts to disable name-only mode within our settings they'll need to do a UI reload for the change to take effect. This is the same as the existing behaviour anyway, however the specific reason we won't toggle the cvar back to an off state is to avoid edge cases where we disable name-only mode after other addons in the environment have also tried to enable it.

On addon load we'll only set the cvar for name-only mode from within the Blizzard nameplates decorator. As such - if the user has Plater or Kui installed, we'll leave it to them to configure name-only mode. This might change based upon feedback.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -200,6 +200,13 @@ stds.wow = {
 			},
 		},
 
+		C_CVar = {
+			fields = {
+				"GetCVarBool",
+				"SetCVar",
+			},
+		},
+
 		C_FriendList = {
 			fields = {
 				"IsFriend",
--- a/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
+++ b/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
@@ -144,6 +144,7 @@ function TRP3_BlizzardNamePlates:OnModuleEnable()
 	self.initializedNameplates = {};
 
 	TRP3_NamePlates.RegisterCallback(self, "OnNamePlateDataUpdated");
+	TRP3_NamePlatesUtil.SetNameOnlyModeEnabled(TRP3_NamePlatesUtil.IsNameOnlyModePreferred());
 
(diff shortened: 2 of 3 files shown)


Meorawr:
One thing I'll need to think about is the latch-like behaviour on our setting of the name-only mode cvar. The cvar itself will act like typical temporary cvars in that it resets on a client restart rather than a UI reload.

The issue this then causes is that if we are the only addon the user has installed which is setting the cvar, by cautiously not allowing ourselves to unset it we're in a position whereby the cvar won't reset to the default state until the user fully exits the client - which is a bit more of an ask than a simple reload.

It's possible that we could just flat out unset the cvar and not attempt to play nice with other addons in the environment in such a case, but it's a bit risky. Probably need to discuss with a few addon authors to see if we can agree on a solution for managing this cvar that doesn't lead to everyone trampling on everyone else's toes.

Meorawr:
Supposedly this cvar will actually reset on UI reload, so the previous comment is effectively a non-issue now.

Meorawr:
This is good to go and works as intended in build 10.0.5.47871 which restored the cvar on Blizzards' end.

Solanya (approved):

PR #686, opened 2023-01-28
Integrate LibDropDownExtension for UnitPopup module

Meorawr:
This reworks our UnitPopup module to instead go through LibDropDownExtension. The library modifies unit popup menus through a slightly different mechanism of creating its own buttons and resizing the host menu, which avoids the edit mode taint issues that our old approach had since 10.0.

One downside is that LibDropDownExtension has a bug whereby check-style menu entries don't correctly set up the anchors for the text to display it beside the check texture - instead they overlap. As such the menu entry for toggling IC/OOC status has been disabled when right-clicking the players' own unitframe.


Meorawr:
Relevant upstream issue: https://github.com/Vladinator/wow-addon-libdropdownextension/issues/2

Solanya (approved):

PR #690, opened 2023-02-08
Fix dashboard dropdowns displaying incorrect selected values

Ghostamoose:
The menus were being initialized before they were opened, and were not correctly updating the selected values. I don't completely understand why this was happening, but my theory is that the initialization was happening before the options were created, or before the menu was open and populated, leading to them getting stuck with the old selected value instead of checking for the new values and updating accordingly.

Simply calling `self:RefreshMenu()` at the end of the menu initialization fixes this behavior. It could be a bit cleaner but that'd require some wider work on the system as a whole. This fix should be fine for now.


Meorawr:
If the title of that ticket is right I think the correct fix would actually just be to register for the `REGISTER_PROFILES_LOADED` event and make it refresh the menu.

Refreshing menus during the initialization callback might work for now but it's inherently not something I'd want to rely on.

Meorawr (approved):

PR #702, opened 2023-02-25
NPC nameplate support and additional settings

Meorawr:
This PR implements a few additional settings for nameplates, and provides a basic filter hook through which Extended will be able to customize NPC nameplates.

The following settings have been added:

- **Maximum name length**: Maximum length of the combined name + short title string before cropping is enforced. This defaults to the current hard-coded limit of 30 characters.
- **Maximum full title length**: As above, but for full titles.
- **Show only first names**: If checked, only displays the first names of players in nameplates. Won't work for companion profiles, nor anything received via MSP.
- **Disable customizations on NPC units**: Disables all nameplate customizations on NPC units. Disabled by default.
- **Disable customizations in instances**: Disables all nameplate customizations while in instances. Enabled by default.
- **Always show target unit**: Always shows the nameplate of the current target - overriding the other "Hide..." settings. Enabled by default as odds are if the user is targeting something, the nameplate will actually be of use.

With the NPC unit setting - it's important to note that NPC nameplates with this PR are now considered valid for customization; as such if the "Hide non-roleplay units" setting is enabled (which it isn't by default) NPC nameplates will be _hidden_. This is a change from the existing behaviour where they'd always be shown. As there's already other settings to disable customizations while in-combat and/or OOC, I'm hoping this change in behaviour won't be much of an issue - especially in combination with the new on-by-default "Disable customization in instances" and "Always show target unit" settings.

Additionally with this PR some optimizations have been employed to the decoration process:

- Nameplate settings are now cached in a table and don't rely on 3+ function calls on every single lookup.
- Cropped versions of the name and full title strings are now present in the display info tables for nameplates; decorators don't need to crop things manually anymore.

Finally for Extended, a filter hook system has been added to allow customizing the fields on the nameplate display info tables _after_ the fields on the tables are populated but before the decorators receive them. A basic filter would be as follows, which would turn everyone's name into a potato.

```lua
TRP3_NamePlates:RegisterDisplayInfoFilter(function(unitToken, displayInfo)
    displayInfo.name = "🥔";
end);
```

Filters are executed for all nameplates that are valid for customization - they won't be executed for example if the user has disabled nameplate customizations in-combat and is in the process of wiping in their premade M+ group.

For NPC nameplate customizations in particular - one thing Extended will need to do is set the `displayInfo.shouldHide` field to `false` if it's applying customizations for the supplied unit. That field will be set to true before the filter executes if the user has the "Hide non-roleplay units" option selected, and by default all NPC units are considered "non-roleplay units" as they don't have ~~phones~~ profiles.


Meorawr:
I've used the opportunity to also fix a long-standing footgun where users needed to ensure both the core nameplates module _and_ an associated decorator (like Blizzard Nameplates) were enabled.

There's been a few cases where users disable the core and get confused because "Blizzard Nameplates" or "Kui Nameplates" appear enabled in the module status list. Removing the ability to disable the core should resolve that; if users don't want nameplates they can just disable the decorators as was initially intended.

As a bonus - the nameplates core can now be disabled at runtime without a UI reload. This won't be hooked into existing decorators as part of this PR, but `/run TRP3_NamePlates:Disable()` (and `/run TRP3_NamePlates:Enable()`) should completely toggle all nameplate decorations on-demand, with the core fully unregistering all event handlers while disabled.

Solanya (approved):

PR #704, opened 2023-03-02
Add guild name/rank misc. info presets

Meorawr:
Two new presets have been added for guild name and rank to the additional info fields in a profile. Like pronouns, these fields are treated specially for both display in tooltips and via MSP comms.

When showing a tooltip for a player if they have a custom guild name set then it will fully override and replace the guild name that would have otherwise been shown based off their players' guild affiliation. The rank will be set to that of the "Guild rank" info field if present, or will default to a sensible "Member" string.

The IC/OOC membership field will *not* be shown if there's a custom guild set, as it doesn't really make sense to show it.

Custom guild name and rank

Custom guild name only

If the player has only a custom guild rank but no guild name, then only the rank of their players' actual guild will be replaced.

For MSP comms two new fields have been added for guild name and rank; "PG" and "PR" respectively. Use of the "G*" namespace was avoided due to that being used for "game info" fields like unit GUID.


Meorawr:
Due to the addition of nameplate fields for guild stuff this is dependent upon the nameplate crop settings, so merge that in first and I'll retarget this to main.

I've done a cleanup pass of all our misc. info handling code and added a new "ID" field to the structure that stores an integral constant identifying the preset from which a field was initially created. This is to resolve issues whereby it was impossible for us to properly deal with cases where the localized names present in fields were neither english nor from the same as the locale of the receiving player.

There's a new flyway patch involved in this which will retroactively add the field to all existing misc. info structures based off their current names. For backwards compatibility, when we try to identify the type of a field in a profile we check if the ID field is present before falling back to the current approach of name comparison.

Meorawr:
Added voice reference too because we're awful people who like to sneak extra features into unrelated PRs on a whim days before a release

Solanya (approved):

PR #710, opened 2023-03-13
Use CallbackHandler for event registries

Meorawr:
This reworks our event handing functionality to go through the more standard CallbackHandler library, which now powers all of our non-widget based registries. A thin set of utilities around these registries has been added.

Advantages:

- Modifying the list of registered callbacks has well-defined behaviour and won't either spuriously error or skip handlers.
- Performance against Blizzards' CallbackRegistryMixin should be greatly improved.
- Error handling behaviour is better; you'll be able to view locals in the error handler if a handler fails.
- Consistency with other addons in the environment.

To register a callback against a registry the main API you'll use is `TRP3_API.RegisterCallback(registry, "event", callback[, owner, ...])`. This creates and returns a CallbackRegistration object that can be dynamically toggled on-demand to efficiently unregister or re-register the managed callback. This is a more direct alternative to the old system of handler IDs.

In cases where you've got multiple callbacks that need registering and dynamically toggling you can make use of either a CallbackGroup or a CallbackGroupCollection.

A CallbackGroup is a flat list of registration objects with a simple Register/Unregister method pair to batch toggle every registration in the group. A CallbackGroupCollection builds upon this by creating "keyed" groups - allowing individual groups of callbacks to be toggled as needed. This is used by the changes in Extended for example when dealing with the tree of handlers for quest steps.

None of these object types attempt to maintain knowledge about whether or not the managed callback(s) are currently registered; this is by-design as it would add considerable complexity for something that isn't ever really needed.

Game events are exposed through a custom registry accessible via `TRP3_API.GameEvents`. This will attempt to validate any registered-for events and - if the client doesn't support them - will raise an error at the point of registration.

Addon events are exposed through our Ace3 addon table (`TRP3_Addon`). Events can be triggered via `TRP3_Addon:TriggerEvent("eventName", ...)` calls and - as with game events - will perform validation to ensure that the requested event actually exists both at the point of registration and triggering.


totalRP3/Core/CallbackRegistry.lua
```diff
@@ -0,0 +1,346 @@
...
+	function(f, a, b, c) return function(_, ...) return f(a, b, c, ...); end end,
+	function(f, a, b, c, d) return function(_, ...) return f(a, b, c, d, ...); end end,
+	function(f, a, b, c, d, e) return function(_, ...) return f(a, b, c, d, e, ...); end end,
+};
+
+local CallbackMethodClosureFactories =
+{
+	function(m) return function(_, o, ...) return o[m](o, ...); end; end,
+	function(m, o) return function(_, ...) return o[m](o, ...); end; end,
+	function(m, o, a) return function(_, ...) return o[m](a, ...); end end,
```
Solanya, line 327: Gonna be honest and admit I have no clue what's going on here, but every other line in this array has o as the first argument, so this one stands out as maybe a mistake?

Also inconsistent semicolons between the `end` on both arrays, first two lines on each have them, not the others.

totalRP3/Core/Configuration.lua
```diff
@@ -52,7 +52,7 @@ local function setValue(key, value)
 	local old = TRP3_Configuration[key];
 	TRP3_Configuration[key] = value;
 	if old ~= value then
-		TRP3_API.events.fireEvent("CONFIGURATION_CHANGED", key, value);
+		TRP3_Addon:TriggerEvent("CONFIGURATION_CHANGED", key, value);
```
Solanya, line 55:
```suggestion
		TRP3_Addon:TriggerEvent(TRP3_Addon.Events.CONFIGURATION_CHANGED, key, value);
```

Solanya:
Verry minor gripe
- TRP3_Addon:TriggerEvent
- TRP3_API.RegisterCallback

It feels weird to me how those calls are so different from each other. It's probably not worth changing (the old events system was itself very inconsistent between registering TRP and game events), just wondering if there could have been a way to keep them consistent.

(Otherwise approved once the arrays have been looked at)

totalRP3/Core/CallbackRegistry.lua
Meorawr: It's valid - added a comment to clarify things. It's technically not used but it would have added some awkward asymmetry without it.

Solanya (approved):

PR #711, opened 2023-03-13
The great color rework

Meorawr:
This commit reworks (and fully migrates) the color system to an approach more compatible with Blizzards' color structure.

A new Color prototype has been added which is *mostly* API-compatible with the default ColorMixin provided by Blizzard, plus some extension methods inherited from the Ellyb color system.

One difference is that we now treat colors as immutable globally - they provide no methods that directly mutate fields on the structure. This is to replace the old and broken concept of a "Frozen" color. Callees cannot know whether or not a color is frozen, so if they were to mutate it they'd need to always defensively make a copy anyway.

Some other Color methods have been removed; the individual "GetRed" style methods are gone - you can just read the structure normally.

To reduce memory churn when processing colors we now also cache them by a unique key formed from the hexadecimal representation of the color. This also gives a slight performance boost as a cache hit for an existing color ends up being faster than the trip to the allocator to make a new one from scratch.

All constant colors have been moved - without changes - to a new table ('TRP3_API.Colors'). Some additional tables are also present to group colors by type - eg. item quality colors.

Finally the contrast adjustment system has been reworked to make use of the APCA-W3 algorithm for lightness contrast and screen luminance. This has logarithmic performance characteristics rather than the old "just bump all the components by 0.001" linear behaviour and - visually - keeps colors closer to their original values as it only modifies the lightness of a color rather whilst keeping the hue and saturation the same.

In turn, the old setting for contrast adjustment is no longer a boolean but instead a tiered system that allows selecting a level of contrast to apply - the default being "Low" rather than the old default of "Off" - so users can now determine how badly they want colors to be adjusted, and gives us options for making contrast adjustment stronger in areas of the UI where text is perhaps smaller or harder to read.


totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -232,11 +229,6 @@ end
 
 --- refreshPsychoColor refreshes the color shown on a line item, updating
 --  the given named color field.
---
---  @param psychoLine The line item to update.
---  @param psychoColorField The color field being updated. Either LC or RC.
---  @param color The color to be applied. Must be an instance of Ellyb.Color,
---               or nil if resetting the color to a default.
```
Solanya, line 239: doc went poof

totalRP3/totalRP3.lua
```diff
@@ -109,9 +106,9 @@ function Globals.addon:OnEnable()
...
-	print(COLORS.ORANGE(("[TRP3: %s]"):format(TRP3_API.VERSION_DISPLAY)) .. " " .. COLORS.RED("Error during addon loading sequence:"));
-	print(COLORS.ORANGE("Sequence ID: ") .. " " .. MAIN_SEQUENCE_ID);
-	print(COLORS.ORANGE("Sub-sequence ID: ") .. " " .. MAIN_SEQUENCE_DETAIL);
-	print(COLORS.ORANGE("Error message: ") .. " " .. tostring(MAIN_SEQUENCE_ERROR));
-	print(COLORS.ITEM_ARTIFACT("Note: If you just recently updated the add-on, remember that you need to fully relaunch the game client. Updating while the game client is open will not load new files."));
+	print(TRP3_API.Colors.Orange(("[TRP3: %s]"):format(TRP3_API.VERSION_DISPLAY)) .. " " .. TRP3_API.Colors.Red("Error during addon loading sequence:"));
+	print(TRP3_API.Colors.Orange("Sequence ID: ") .. " " .. MAIN_SEQUENCE_ID);
+	print(TRP3_API.Colors.Orange("Sub-sequence ID: ") .. " " .. MAIN_SEQUENCE_DETAIL);
+	print(TRP3_API.Colors.Orange("Error message: ") .. " " .. tostring(MAIN_SEQUENCE_ERROR));
+	print(TRP3_API.ItemQualityColors.Artifact("Note: If you just recently updated the add-on, remember that you need to fully relaunch the game client. Updating while the game client is open will not load new files."));
```
Solanya, line 113: Completely unrelated but since I'm seeing it, we can remove that line, not been true for a while.

totalRP3/totalRP3.toc
```diff
@@ -7,7 +7,7 @@
 ## Version: @project-version@
 ## Notes: The best roleplaying addon for World of Warcraft.|n|n|cffffd200Version:|r @project-version@
 ## IconTexture: Interface\AddOns\totalRP3\Resources\trp3minimap
-## OptionalDeps: Ace3, CallbackHandler-1.0, LibChatAnims, LibCompress, LibDBIcon-1.0, LibDeflate, LibMSP, LibRPMedia, LibStub, Kui_Nameplates, Kui_Nameplates_Core, Prat-3.0
+## OptionalDeps: !ClassColors, Kui_Nameplates, Kui_Nameplates_Core, Prat-3.0
```
Solanya, line 10: why are they all gone 🤔

Solanya (changes requested):

totalRP3/totalRP3.toc
Solanya: Oh I might be comparing stuff that's not up-to-date with main, right

totalRP3/totalRP3.lua
Meorawr: Technically it's still accurate for non-Lua/XML files - thoughts on those?

Solanya: I don't know in what case it could create an error but I haven't thought about it for very long

Solanya (approved):

PR #712, opened 2023-03-15
Implement 10.1 addon compartment integration

Meorawr:
The latest build of 10.1.0 allows integrating ourselves into the Addon Compartment dropdown through TOC metadata fields, sidestepping the taint issues that the old manual registration method would have had.

Blizzard would like general feedback on the feature overall, so I'm unconditionally implementing support for it as a test case. 

image

There's a few outstanding issues that Blizzard ideally need to resolve - I'd like to add an option to select between the minimap button or the compartment for instance - but hopefully we can push for some of these to be looked at during 10.1 testing:

- https://github.com/Stanzilla/WoWUIBugs/issues/400
- https://github.com/Stanzilla/WoWUIBugs/issues/401


Meorawr:
The latest 10.1 build added right-click and tooltip support, so I've integrated that now. We show the tooltip in the default tooltip location (ie. the bottom-right of the screen) for two reasons:

1. There's not a lot of space near the compartment for tooltips to really look that good anyway.
2. Blizzard aren't passing us a reference to the dropdown list button - so we can't anchor to it even if we wanted to.

Solanya (approved):

PR #735, opened 2023-03-24
Profile automation

Meorawr:
This implements a very basic version of profile automation. Per the previous discussion on virtual profiles vs. automation, I think there's still some room for automation to exist for basic tasks like location, IC/OOC toggles, and profile switches, and some of our addon settings - so this feature is currently aimed at that case.

This implementation at-present contains the following actions. I'm not looking to expand it beyond these at this point unless there's some really low hanging fruit that are worth bashing up-front.

- Change profile
- Change roleplay status
- Toggle location broadcast
- Toggle nameplate customizations
- Toggle friendly player nameplates
- Toggle friendly NPC nameplates

A few macro conditions have been added to the system as a whole:

- "[afk]" - Checks if the player is flagged as AFK.
- "[character:name]" - Checks if the players' current character matches the specified name.
- "[dnd]" - Checks if the player is flagged as DND.
- "[instance]" / "[instance:type]" - Checks if the player is in an instanced zone, optionally of a specific type.
- "[pvp]" - Checks if the player is flagged for PvP.
- "[warmode]"  - Checks if War Mode is active. Always evaluates to false in Classic.
- "[realm:name]" - Checks if the players' current realm matches the specified name.

To avoid CPU usage overhead, automation processes updates every 1 seconds. It will immediately update at the end of the current game frame if a limited set of game events fire. We explicitly do **not** listen to our own events (like `ROLEPLAY_STATUS_CHANGED`) to avoid cases where someone sets up a silly condition that alternates their IC/OOC status every frame.

The UI is extremely spartan and just consists of a dropdown to select the action and an editbox which users are expected to enter a macro conditional string. These support the custom macro conditions we implemented previously for the `/trp3 set` command and friends. These dropdown entries all have descriptive tooltips in the dropdown when hovered.

I'd rather get feedback on the usefulness of the feature from users before improving the UI further; if no one ends up using it then there's not too much point in adding a more "basic" editor. One potential point of improvement would be to add something that lists all our custom macro conditionals - this might be a good fit for the "Help & Support" menu.

image
image


Meorawr:
I went ahead and added that `[altform]` macro conditional you wanted for Worgen/Dracthyr forms. Perhaps counterintuitively `[altform]` is true when you're in your _human_ form for those races, but I'd rather just defer to the infinite wisdom of the API than special case this.

Additionally there's now a new action to change equipment sets. Unlike the profile swap it requires an exact set name rather than a partial match. You can also use a set ID but that won't be documented in the help text.

Meorawr:
One thing to note is that on Classic Wrath the edit field for automation behaves a bit weirdly - the placeholder example text won't clear on-click and the font is always greyed out.

I'd rather just let Blizzard fix that problem as it's specific to 3.4.1 and doesn't impact anything else.

Solanya (approved):

PR #741, opened 2023-04-14
Update Plater mod and move nameplate customization to TRP3

Ghostamoose:
First off, this is a bit of a disaster in it's current state, but works for the most part.

Moving the customization to the TRP3 module allows greater visibility into the inner workings of the mod and allows me to work somewhere other than the ghastly Plater mod UI in-game.

There have been a lot of changes to the underlying systems that this module uses, so the way I make use of them here may or may not be correct.

Things that need to be fixed:
- [x] The `DisplayInfo` lags behind when the `CONFIGURATION_CHANGED` event is fired, so occasionally the options will get desynced and start doing the opposite of what's intended. Re-adding the nameplate to the screen fixes this, but it seems that the nameplates are being customized before `DisplayInfo` is updated with the new configuration.
- [x] Decrease jank


totalRP3/Modules/NamePlates/NamePlates_Plater.lua
```diff
@@ -40,30 +169,31 @@ function TRP3_PlaterNamePlates:OnModuleEnable()
...
 		local newScript = Plater.BuildScriptObjectFromIndexTable(newScriptTable, scriptType);
 
-		for i = 1, #scriptDB do
-			local scriptObject = scriptDB[i];
+		for _, scriptObject in ipairs(scriptDB) do
 			if scriptObject.Name == newScript.Name then
 				scriptAdded = scriptObject;
 			end
 		end
 	end
```
Meorawr, line 177: Is it possible for there to be multiple scripts with the same name, or would it be safe for this to break after finding the first one?


```suggestion
		for _, scriptObject in ipairs(scriptDB) do
			if scriptObject.Name == newScript.Name then
				scriptAdded = scriptObject;
				break;
			end
		end
	end
```

Meorawr (dismissed):
Looks fine though haven't tested it. One comment but it's not urgent.

Meorawr (commented):
Looks fine though haven't tested it. One comment but it's not urgent.

Meorawr (approved):

PR #743, opened 2023-04-23
Party Time

Solanya:
Cleaning up after April Fools but fatcat won't leave and brought a friend, so I sent them away to the dashboard and told them to stay hidden until the party's started.


totalRP3/Modules/Dashboard/Dashboard.xml
```diff
@@ -16,6 +16,56 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
 	<Include file="TabFrame.xml"/>
 	<Include file="StatusPanel.xml"/>
 
+	<Button name="TRP3_PartyTime" virtual="true">
```
Meorawr, line 19: Can't these just be normal Frames?

Meorawr (approved):
Obviously this won't work on Classic Era but 1.15.0 should be out in a few months.

PR #765, opened 2023-05-05
Plater mod bug fixes

### LINKED ISSUE #760: Odd TRP3/Plater Behavior
## Describe the issue
Nameplates are getting "caught" towards the edge of the screen if they contain a title (NPC) or guild (player).

## Expected behaviour
Nameplates should not remain on screen when source is not visible.

## Issue checklist

> You can replace the "[ ]" with "[x]" for each option below to tick it.

- [X] The issue occurs with the [latest version] of the addon.
- [ ] The issue occurs with Total RP 3 as the only enabled addon.
- [X] The issue occurs only with other addon(s) enabled (please provide a list).
      - Plater
      - Plater TRP3 Mod Enabled (possible culprit)

## Steps to reproduce
1)Have Plater activated with TRP3 Mod enabled AND TRP3 addon enabled
2)Run around a populated area (Valdrakken, for instance)
3) Issue occurs

## Additional files
[Screenshot](https://ibb.co/z7pRRxK)

---

[error logs]: https://www.curseforge.com/wow/addons/bugsack
[profile data files]: https://github.com/Total-RP/Total-RP-3/wiki/Saved-Variables
[latest version]: https://www.curseforge.com/wow/addons/total-rp-3/files

Ghostamoose:
- Fixes errors being thrown in chat if TRP3 was disabled but the Plater mod was enabled.
     - The mod itself is to be enabled/disabled by TRP3 via the "Plater Nameplates" module in TRP3's module settings. While it will **not** do anything with TRP disabled, due to a limitation with Plater mods, it will remain active if it was active before TRP was disabled.
- Fixes #760, a bug that caused player nameplates to get stuck on screen and/or not being properly removed from the screen when they should be.

Still working on verifying the second fix, as I **still** cannot reliably reproduce it.


Ghostamoose:
Update: fix for the nameplates getting stuck does not work :))))))))

Ghostamoose:
Found the fix for the nameplates getting stuck. Details below!

Short answer: We were updating the nameplate visibility without checking if the nameplate was still on the screen, leading to us showing nameplates that should not be on the screen.

Long answer: For whatever reason, there is some amount of delay between the `NAMEPLATE_UNIT_REMOVED` event being fired, and the nameplate visibility (`nameplate:IsShown()`) being updated, meaning we were executing our code in that little window between when the nameplate was removed from the screen, and it being properly hidden by the game or by Plater.

The fix: Just check `nameplate.unitFrame.PlaterOnScreen` for the most reliable gauge of whether the nameplate is actually on the screen or not.

Side note: For some reason, I was entirely unable to get this to repro on live, but the bug linked in the original post shows that it can happen. I was able to get almost a 100% repro rate on Classic though, which leads me to believe that there is some difference in either, how Plater handles `NAMEPLATE_UNIT_REMOVED` on Classic, or the underlying Blizzard system works differently there, making it more likely for the bug to occur.

Solanya (approved):

PR #770, opened 2023-05-14
Event trace integration

Meorawr:
This adds some hooks to integrate our callback registries into the event trace panel so that we can actually see the events firing through things without needing to revert back to the stone age with print calls and the like. The integration is only enabled in debug mode to avoid any potential conflicts.

image


totalRP3/Core/Debug.lua
```diff
@@ -0,0 +1,74 @@
...
+local function OnCallbackEventFired(registry, event, ...)
+	if not EventTracingRegistries[registry] then
+		return;
+	elseif not EventTrace or not EventTrace.LogLine then
+		return;
+	end
+
+	if TRP3_API.IsEventValid(TRP3_API.GameEvents, event) then
+		return;
+	end
```
Meorawr, line 37: Will remove this later - this is a leftover from before I added the RemoveFromEventTraceWindow function.

Solanya: Are we still removing this later

totalRP3/Modules/Map/MapScannersManager.lua
```diff
@@ -49,7 +41,7 @@ function MapScannersManager.launch(scanID)
 	scan:ResetScanData();
 
 	local function OnScanTimerElapsed()
-		TRP3_Addon:TriggerEvent(Events.MAP_SCAN_ENDED);
+		TRP3_Addon:TriggerEvent("MAP_SCAN_ENDED");
```
Solanya, line 44:
```suggestion
		TRP3_Addon:TriggerEvent(TRP3_Addon.Events.MAP_SCAN_ENDED);
```

totalRP3/Modules/Map/MapScannersManager.lua
```diff
@@ -61,7 +53,7 @@ function MapScannersManager.launch(scanID)
 	end
 
 
-	TRP3_Addon:TriggerEvent(Events.MAP_SCAN_STARTED, scan.duration);
+	TRP3_Addon:TriggerEvent("MAP_SCAN_STARTED", scan.duration);
```
Solanya, line 56:
```suggestion
	TRP3_Addon:TriggerEvent(TRP3_Addon.Events.MAP_SCAN_STARTED, scan.duration);
```

Solanya (approved):

PR #774, opened 2023-05-27
Banish malicious dice objects

Ghostamoose:
Ensures that dice roll objects received from other players match the expected structure of a dice roll to prevent arbitrary messages from being displayed to the end user if the object they received is invalid. Dice rolls should only contain numbers or nil values.

image

### DIFF
--- a/totalRP3/Core/Slash.lua
+++ b/totalRP3/Core/Slash.lua
@@ -147,6 +147,22 @@ function TRP3_API.slash.rollDices(...)
 	return total, i;
 end
 
+local function isValidDiceObject(diceObject)
+	if type(diceObject) ~= "table" or not diceObject.t or #diceObject > 4 then
+		return false;
+	end
+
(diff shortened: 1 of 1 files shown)


totalRP3/Core/Slash.lua
```diff
@@ -147,6 +147,22 @@ function TRP3_API.slash.rollDices(...)
...
 
+local function isValidDiceObject(diceObject)
+	if type(diceObject) ~= "table" or not diceObject.t or #diceObject > 4 then
+		return false;
+	end
+
+	for _, v in pairs(diceObject) do
+		local valueType = type(v);
+
+		if valueType ~= "number" and valueType ~= "nil" then
```
Meorawr, line 158: It's impossible for the type to ever be nil.

Meorawr (changes requested):

Solanya (approved):

PR #780, opened 2023-06-23
Fix potential SetText error

Meorawr:
Per Discord bug report - unsure what the cause is but as it can happen...

image


Meorawr:
Luacheck errors are unrelated to the PR; I guess it got updated upstream and has additional things now.

Solanya (approved):

PR #788, opened 2023-07-20
Add support for automation profiles

### LINKED ISSUE #777: Add character-specific automation settings
Tentatively would like to do this for 2.6, but it might slip a bit beyond that.

Meorawr:
Closes #777,

The current automation system doesn't scale too well when dealing with multiple characters. The "Change profile" action ends up being with rules that are often character-specific, meaning users need to populate this rule with long and complicated rulesets if they play multiple characters - and then need to update it each time they make a new character.

To alleviate this problem a new profile system has been added that functions similarly to profiles in Ace-based addons.

The dropdown to the right of the automation panel allows users to create, enable, or modify a profile. Profiles can be activated on a per-character basis, but will always default to a shared "Default" profile when the addon is first loaded on a new character.


Meorawr:
Check failure not related to PR; CF being shit as usual.

Solanya (approved):

PR #789, opened 2023-07-22
Regroup TRP3_MainFrame OnLoad scripts

Solanya:
While investigating the Extended resizing issue, I found out there were two Scripts sections in TRP3_MainFrame. As both TRP3 and Extended have to include themselves in the UISpecialFrames, I thought it made sense to just auto-include that line in the mixin to remove that extra section.


Meorawr (approved):
How disgustingly sensible

PR #809, opened 2023-11-08
Sort map scans by name in dropdown

Meorawr:
Nothing irks me more than an unsorted list.

image


Solanya:
I'm sorry, I was napping, but other languages exist, we should probably use something that is locale-independent :(

Solanya (approved):

PR #813, opened 2023-11-12
Icon Browser Improvements

Meorawr:
It's nearly 4am but basically:

- The selected icon now shows first in the grid with a yellow highlight.
- Made icons look less ugly in the tooltip - size 70 was a weird choice, okay?
- Made icons look less ugly in the grid - we mask off the corners and have an atlas to give them a bit of a background and a proper pressed state.
- Made the search box look less awful - it's now at the top-right of the browser where search bars normally are, and inherits the same template as every other search input in the UI.
- Made the search box behave less awful - we now don't care about punctuation characters; "inv misc poop" is treated the same as "inv_misc-poop".
- Added an empty state; if no icons are found we display some nice text in-place of the icon grid.
- Removed the icon count text - users probably won't really ever care to know how many icons there are, and it takes up space.
- Added an input blocker to prevent clicking icons while a search is in progress. This is to work around API issues; ScrollBox list views seem to render their elements unclickable for a frame whenever the data provider updates, and as we're searching every game frame the icons can't be clicked anyway until the search completes. The input blocker is more intended to represent this visually by applying a dark overlay over the icons until the search is complete.
- Relocated the progress bar to sit atop everything, complete with some wicked animations.
- The scrollable area for icons has been greatly increased.

Couple of screenshots and videos from various points in these changes;

Selected icon highlight:
image

Empty state:
https://github.com/Total-RP/Total-RP-3/assets/287102/6bc0246f-743a-4790-9e16-b104aefd79a4

Input blocker + progress animation:
https://github.com/Total-RP/Total-RP-3/assets/287102/e9e56b82-13b0-42bb-bb7c-cc4526b86e61


Meorawr:
This is basically good to go at this point - only thing that's outstanding is whether or not I want to make the browser sort icons that are in the players' spellbook and/or equipped items first, similar to how the icon browser works in the macro window.

Solanya (approved):

PR #816, opened 2023-11-16
Fix altform trigger on druid forms

DJScias:
# Preface
According to the [Macro Conditionals](https://github.com/Total-RP/Total-RP-3/wiki/Macro-conditionals) wiki page:
> `[altform]`: (Worgen/Evoker only) Checks if the character is in their human form (Example: `[altform] Human Profile; Worgen Profile`).

However, in its current form, `altform` will also trigger as a positive if you are in druid form as a Worgen. This is not desirable, as the macro conditional is only supposed to be true for their 'True Form' (Dracthyr / Worgen).

# Example to reproduce
Look at the following automation profile:
```
[altform] "Human"; "Worgen
```
This turns on the Worgen profile when in Worgen form as expected, but also turns it on when in Druid form.

# The problem explained
`C_PlayerInfo.GetAlternateFormInfo()` returns the following:
- Alternate form (Human / Visage): `true true`.
- True Form (Dracthyr / Worgen): `true false`.
- Druid forms (as Worgen): `true false`.

Since TRP relies on the second result being false to determine if you're in your True Form, it also triggers on druid forms for Worgens.

# The fix explained
We now always assume that `inAlternateForm` starts as true (true means we're in our true form as a dracthyr or Worgen).

Then, when an `altform` macro conditional is used, we check to see if `C_PlayerInfo.GetAlternateFormInfo` exists AND NOT a druid that is shapeshifted shapeshifted (`and not (TRP3_API.utils.str.GetClass("player") == "DRUID" and GetShapeshiftForm() ~= 0)`), because `true forms` are always considered by the game as not shapeshifted. The druid check is because other classes also have non-0 results for shapeshift forms like a rogue in stealth, or a warrior's specific stance.

This has the side effect that druid forms, if not handled by multiple macro conditionals (as they should be), will now trigger the alternate form (Human for Worgens). But this seems more desirable than before, as strictly speaking a druid form is still NOT your 'true form'. Druid forms should be handled by the `[form:x]` macro conditional as is.

# Regressions tested
I've re-tested the automation profiles to check for problems:
```
[altform] "Human"; "Worgen
```
Correctly switches to Worgen in Worgen form, and does not trigger the Worgen profile when the user is shapeshifted (and thus not in their 'true form').

Then again with multiple conditionals:
```
[form:3] "Travel"; [form:4] "Moonkin"; [altform] "Human"; "Worgen"
```
This works just fine before AND after my fix, because you made sure to handle the druid forms first.

This means there are no regressions and everything else continues to work as it should.

### DIFF
--- a/totalRP3/Modules/Automation/Automation_Conditions.lua
+++ b/totalRP3/Modules/Automation/Automation_Conditions.lua
@@ -147,9 +147,9 @@ TRP3_AutomationUtil.RegisterCondition({
 	tokens = { "altform" },
 
 	Evaluate = function()
-		local inAlternateForm = false;
+		local inAlternateForm = true;
 
-		if C_PlayerInfo.GetAlternateFormInfo then
+		if C_PlayerInfo.GetAlternateFormInfo and GetShapeshiftForm() == 0 then
 			inAlternateForm = select(2, C_PlayerInfo.GetAlternateFormInfo());
 		end
 
--- a/totalRP3/Modules/Automation/Automation_Conditions.lua
+++ b/totalRP3/Modules/Automation/Automation_Conditions.lua
@@ -149,7 +149,7 @@ TRP3_AutomationUtil.RegisterCondition({
 	Evaluate = function()
 		local inAlternateForm = true;
 
-		if C_PlayerInfo.GetAlternateFormInfo and GetShapeshiftForm() == 0 then
+		if C_PlayerInfo.GetAlternateFormInfo and not (TRP3_API.utils.str.GetClass("player") == "DRUID" and GetShapeshiftForm() ~= 0) then
 			inAlternateForm = select(2, C_PlayerInfo.GetAlternateFormInfo());
 		end
 


Meorawr:
To add a bit more:

People will probably be using `[altform]` in existing non-character-specific conditionals that potentially also run on characters that are neither Worgen/Evoker, so having the default flip state flip from false to true will break them.

With respect to Worgen Druids, it's certainly not too convenient that the API works this way but _at this stage_ I'd rather not have a hack in place to work around one specific race/class combination when I think one can just use `[altform,form:0]` for a human-only check and just `[form]` for "am I in any Druid form?". Macro conditionals should in general map in a fairly 1:1 manner to underlying APIs, quirks included.

DJScias:
@Solanya 
> > We now always assume that `inAlternateForm` starts as true (true means we're in our true form as a dracthyr or Worgen).
> 
> That is wrong, true means alternate form, as the boolean name clearly states.

This was a typo and not how the code was working that I wrote in the first place, I apologize for the typo as it was meant to be:
> We now always assume that inAlternateForm starts as true (true means we're in our alt form as human/visage or druid form).

I now understand, as per our discord chat, that only Human/Visage is considered an alt form and that **druid forms are not considered alt forms** for this implementation. My mistake was reading the doc and thinking "anything that is NOT the true dracthyr/worgen forms should be considered alt forms", which to me had druid forms in the list of thinking.
  
@Meorawr 
> To add a bit more:
> 
> People will probably be using `[altform]` in existing non-character-specific conditionals that potentially also run on characters that are neither Worgen/Evoker, so having the default state flip from false to true will break them.
> 
> With respect to Worgen Druids, it's certainly not too convenient that the API works this way but _at this stage_ I'd rather not have a hack in place to work around one specific race/class combination when I think one can just use `[altform,form:0]` for a human-only check and just `[form]` for "am I in any Druid form?". Macro conditionals should in general map in a fairly 1:1 manner to underlying APIs, quirks included.

This makes a lot of sense. You want to avoid specific race/class checks, I can totally understand why that would be a bit much since it only happens for a single combination right now. Maybe the future API will define druid forms differently from the alternate form idea, but who knows.

Both of your explanations together make a lot of sense, thanks for that. The PR is indeed moot, thank you for closing. 👍

PR #819, opened 2023-11-16
Prat duplicate names on preset emotes

Solanya:
Fixes the names showing on lines they shouldn't in Prat.
image

As to what caused the issue to arise is anyone's guess, didn't notice anything related glancing at the recent commits, but I'm blind.


Meorawr (approved):
I must scream but I have no `message.PLAYER`.

PR #824, opened 2023-12-10
Customize own pet nameplate

Solanya:
We looked up our own pet profile in the directory of others players' pet profiles, couldn't find it, so never customized the nameplate. I just check who the owner is before calling the appropriate function.


Meorawr (approved):
Now I'm curious why I never noticed this to begin with - I guess I'd just tested it with other player's pets.

PR #835, opened 2024-01-14
Misc preset renaming issues

### LINKED ISSUE #834: Preset additional info names can be modified
As some users have been confused about which text box to modify for additional info presets, we should make it so preset names cannot be modified. This should keep in mind that fields can be reordered (and enable/disable appropriately during reorder).

As for handling existing fields, a flyway patch should check the current name against the original preset name, turning the field into a custom field (by clearing the ID value) if the name does not match.

Solanya:
Fixes #834.

Preset misc info names can no longer be edited, and a flyway patch will convert preset fields to custom fields if they have been renamed.

Imported profiles from before the ID conversion were also not given the new ID, this is now fixed, along with the same custom field conversion if there is an ID not matching the name.

### DIFF
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -600,15 +600,9 @@ function TRP3_API.profile.init()
 					-- Misc info ID conversion
 					if data.player and data.player.characteristics and data.player.characteristics.MI then
 						for _, miscData in ipairs(data.player.characteristics.MI) do
-							if not miscData.ID then
-								-- No ID = retrieve ID from name
+							if not miscData.ID or miscData.ID ~= TRP3_API.MiscInfoType.Custom then
+								-- Adding ID from name if ID missing, or setting a preset to custom if renamed
 								miscData.ID = TRP3_API.GetMiscInfoTypeFromData(miscData);
-							elseif miscData.ID ~= TRP3_API.MiscInfoType.Custom then
-								-- Existing non-custom ID = check name matches ID, convert to custom otherwise
-								local miscInfoData = TRP3_API.GetMiscTypeInfo(miscData.ID)
-								if miscData.NA ~= miscInfoData.localizedName and miscData.NA ~= miscInfoData.englishName then
-									miscData.ID = TRP3_API.MiscInfoType.Custom;
-								end
 							end
 						end
 					end
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -602,7 +602,7 @@ function TRP3_API.profile.init()
 						for _, miscData in ipairs(data.player.characteristics.MI) do
 							if not miscData.ID or miscData.ID ~= TRP3_API.MiscInfoType.Custom then
 								-- Adding ID from name if ID missing, or setting a preset to custom if renamed
-								miscData.ID = TRP3_API.GetMiscInfoTypeFromData(miscData);
+								miscData.ID = TRP3_API.GetMiscInfoTypeByName(miscData.NA);
 							end
 						end
 					end
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -596,6 +596,22 @@ function TRP3_API.profile.init()
 					if data.player and data.player.about and data.player.about.MU and type(data.player.about.MU) == "string" then
 						profiles[profileID].player.about.MU = Utils.music.convertPathToID(data.player.about.MU);
 					end
+
+					-- Misc info ID conversion
+					if data.player and data.player.characteristics and data.player.characteristics.MI then
+						for _, miscData in ipairs(data.player.characteristics.MI) do
+							if not miscData.ID then
+								-- No ID = retrieve ID from name
+								miscData.ID = TRP3_API.GetMiscInfoTypeFromData(miscData);
+							elseif miscData.ID ~= TRP3_API.MiscInfoType.Custom then
+								-- Existing non-custom ID = check name matches ID, convert to custom otherwise
+								local miscInfoData = TRP3_API.GetMiscTypeInfo(miscData.ID)
+								if miscData.NA ~= miscInfoData.localizedName and miscData.NA ~= miscInfoData.englishName then
+									miscData.ID = TRP3_API.MiscInfoType.Custom;
+								end
+							end
+						end
+					end
 				end
 
 				TRP3_ProfileImport:Hide();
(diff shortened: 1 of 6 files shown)


totalRP3/Core/Profiles.lua
```diff
@@ -596,6 +596,22 @@ function TRP3_API.profile.init()
...
+						for _, miscData in ipairs(data.player.characteristics.MI) do
+							if not miscData.ID then
+								-- No ID = retrieve ID from name
+								miscData.ID = TRP3_API.GetMiscInfoTypeFromData(miscData);
+							elseif miscData.ID ~= TRP3_API.MiscInfoType.Custom then
+								-- Existing non-custom ID = check name matches ID, convert to custom otherwise
+								local miscInfoData = TRP3_API.GetMiscTypeInfo(miscData.ID)
+								if miscData.NA ~= miscInfoData.localizedName and miscData.NA ~= miscInfoData.englishName then
+									miscData.ID = TRP3_API.MiscInfoType.Custom;
+								end
```
Meorawr, line 611: I'd probably just suggest ignoring the "this is a legacy function" comment and checking if `miscData.ID ~= TRP3_API.GetMiscInfoTypeByName(miscData.NA)` when resetting to a custom ID.

The comment was more written because at the time there wasn't anything that needed to explicitly do name-based checks; as this is now changing the comment can just be removed and function kept around.

totalRP3/Modules/Flyway/FlywayPatches.lua
```diff
@@ -263,3 +263,27 @@ TRP3_API.flyway.patches["16"] = function()
...
+	for _, profile in pairs(TRP3_Profiles) do
+		local miscInfo = SafeGet(profile, "player", "characteristics", "MI");
+
+		if miscInfo then
+			for _, miscData in ipairs(miscInfo) do
+				if miscData.ID ~= TRP3_API.MiscInfoType.Custom then
+					local miscInfoData = TRP3_API.GetMiscTypeInfo(miscData.ID)
+					if miscData.NA ~= miscInfoData.localizedName and miscData.NA ~= miscInfoData.englishName then
+						miscData.ID = TRP3_API.MiscInfoType.Custom;
+					end
```
Meorawr, line 284: As above.

totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -1119,6 +1119,15 @@ function setEditDisplay()
 		frame.miscIndex = frameIndex;
 		_G[frame:GetName() .. "Icon"].IC = miscStructure.IC or TRP3_InterfaceIcons.Default;
 		_G[frame:GetName() .. "NameField"]:SetText(miscStructure.NA or loc.CM_NAME);
+		-- Disabling name edition on presets
```
Meorawr, line 1122:
```suggestion
		-- Disable name editing on presets
```

totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -1119,6 +1119,15 @@ function setEditDisplay()
...
+		if miscStructure.ID == TRP3_API.MiscInfoType.Custom then
+			_G[frame:GetName() .. "NameField"]:Show();
+			_G[frame:GetName() .. "PresetNameText"]:Hide();
+		else
+			_G[frame:GetName() .. "NameField"]:Hide();
+			_G[frame:GetName() .. "PresetNameText"]:Show();
+			_G[frame:GetName() .. "PresetNameText"]:SetText(miscStructure.NA or loc.CM_NAME);
```
Meorawr, line 1129:
```suggestion
			frame.PresetName:Hide();
		else
			_G[frame:GetName() .. "NameField"]:Hide();
			frame.PresetName:Show();
			frame.PresetName:SetText(miscStructure.NA or loc.CM_NAME);
```

I imagine the use of global names here was more for consistency reasons, but I think it just makes sense to use parent keys going forward where possible for newer stuff even if it makes for some ugliness like this.

totalRP3/Modules/Register/Characters/RegisterUICharacteristics.xml
```diff
@@ -53,6 +53,17 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
...
 			</Button>
 		</Frames>
+		<Layers>
+			<Layer level="OVERLAY">
+				<FontString name="$parentPresetNameText" parentKey="PresetName" text="[preset]" inherits="GameFontNormal" justifyH="LEFT">
+					<Size x="100" y="18"/>
+					<Anchors>
+						<Anchor point="LEFT" x="15" y="0" relativePoint="RIGHT" relativeTo="$parentIcon"/>
+					</Anchors>
+					<Color r="1.0" g="1.0" b="1.0"/>
```
Meorawr, line 63:
```suggestion
				<FontString parentKey="PresetName" text="[preset]" inherits="GameFontHighlight" justifyH="LEFT">
					<Size x="100" y="18"/>
					<Anchors>
						<Anchor point="LEFT" x="15" y="0" relativePoint="RIGHT" relativeTo="$parentIcon"/>
					</Anchors>
```
With the above we don't need a name anymore, and we can remove the colour if we just use GameFontHighlight.

Meorawr (changes requested):

totalRP3/Core/Profiles.lua
```diff
@@ -596,6 +596,16 @@ function TRP3_API.profile.init()
 					if data.player and data.player.about and data.player.about.MU and type(data.player.about.MU) == "string" then
 						profiles[profileID].player.about.MU = Utils.music.convertPathToID(data.player.about.MU);
 					end
+
+					-- Misc info ID conversion
+					if data.player and data.player.characteristics and data.player.characteristics.MI then
+						for _, miscData in ipairs(data.player.characteristics.MI) do
+							if not miscData.ID or miscData.ID ~= TRP3_API.MiscInfoType.Custom then
+								-- Adding ID from name if ID missing, or setting a preset to custom if renamed
+								miscData.ID = TRP3_API.GetMiscInfoTypeFromData(miscData);
```
Meorawr, line 605:
```suggestion
								miscData.ID = TRP3_API.GetMiscInfoTypeByName(miscData.NA);
```

GetMiscInfoTypeFromData will just return `miscData.ID` if it's not nil.

Meorawr (changes requested):

Meorawr (approved):
good boy

PR #872, opened 2024-05-06
Spaces clean up

Solanya:
Removed spaces between | and c/r
Moved some spaces around
Removed some when there were two consecutive ones
Fixed some wildly incorrect color codes when I saw them
Removed some trailing spaces
Removed like a couple spaces before punctuation that aren't supposed to be there in English but y'know how it is

This is all fascinating to look at really


Meorawr (approved):
please clap

PR #875, opened 2024-05-06
Add 'Open Profile' option to pet menus

Meorawr:
Yet another terrible idea has been foist upon us by the dregs of our own Discord. Woe is us, forced into this 'ternal suffering where we must appease the whims and fancies of those who proclaim themselves to be thinkers. Hark! My agony calls.


Solanya (approved):
More is thy due than more than all can pay.

PR #878, opened 2024-05-07
Ensure ChatThrottleLib is packaged

Meorawr:
It's marked as a dependency of Chomp in its own pkgmeta, but for some reason we neglected to actually ship it ourselves. Not a huge issue as Chomp has coped until now without it, but was still technically incorrect anyway - and now fatally so.


Meorawr:
Seems to all work fine - tested loading TRP with nothing else at all and it's got all the right versions, no errors, and can fundamentally pull profiles.

Solanya (approved):

PR #887, opened 2024-05-17
Restore the Timerunner icon on customized chat names

Solanya:
Because the timerunner icon is added directly in the GetColoredName function, we're eating it when applying the chat name customization, since we replace the entire function. This should restore the hourglass in front of everything else so you still know who's a timerunner.

Note: When the setting to show the icon on chat names is enabled, it'll take priority over the hourglass. Seems like a Blizzard bug that we can't have two textures in there.


Meorawr (approved):
This is most poggers.

PR #894, opened 2024-06-01
Handling of default profile in customizations and directory

Solanya:
The way the default profile works leads to weird displays when it comes to customization. To be able to quickly identify whether a profile ID belongs to a regular profile or the default one, the last character is now a special one (I chose an asterisk but I'm not picky, we can pick something else).

(Old default profile IDs get replaced by the flyway, and the old links will automatically reselect the new profile ID when the old one can't be found)

This way, I can remove customization from chat and nameplates, and hide profiles from the directory if they're default. There might be other places where this should be done, I'll look around, but wanted to get the first draft out in case there's major things to do better.

(Oh yeah, also fixed a bug where the fallback function for GetColoredName was called with the wrong arguments when the unitID or realm were missing, leading to chat messages having incorrect displays)

### DIFF
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -104,10 +104,13 @@ end
 TRP3_API.profile.isProfileNameAvailable = isProfileNameAvailable;
 
 -- Duplicate an existing profile
-local function duplicateProfile(duplicatedProfile, profileName)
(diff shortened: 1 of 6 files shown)


Meorawr (dismissed):
I'm Commander Shepard and this is my favorite PR on the repository.

Meorawr (approved):

PR #900, opened 2024-06-08
Migrate to new menu system

Meorawr:
This lengthy and frankly under-commented mess of changes converts all of our dropdown and context menus in the core addon to use Blizzard's new menu system added in 11.0.

The changes largely mean that we now follow the new way of doing things and have completely excised the MSA dropdown library and LibDropDownExtension. For Classic compatibility, there's a lightweight backport of the new API that I've dubbed as "legacy" menus which maps the subset of the new concepts that we need back to UIDropDownMenu.

The new menu system will be used based upon the `TRP3_USE_MODERN_MENUS` global which is currently set to true if we detect one of the new templates added in 11.0 specific to menus. If it isn't set, then the legacy renderer kicks in and UIDropDownMenu will be used behind the scenes.

With this set of changes the TRP3_DropDownMenuTemplate template has been deprecated. All uses of it in the core addon are gone and the replacement TRP3_DropdownButtonTemplate is used instead.

As part of this change, Extended will need to fix up the position of all of its dropdowns, approximately nudging them by 13-16 units left or right based upon their current anchors. The reason for this is that the old dropdowns (based on UIDropDownMenuTemplate) had a lot of padding on those edges which isn't present with new menus.

Otherwise in terms of functionality - all menus seem to work fine without any casualties.

On the unit popups module, this has undergone more significant changes and has been intentionally disabled unless modern menus are available.

The reason here is that unlike the other menus where the differences between the two renderers are abstracted away, it's a bit harder to do that in the context of unit popups because it needs to append to existing menus _and_ deal with the fact that there's a library in the way rather than simply talking straight to UIDropDownMenu. As that'd be a lot of effort we'll instead just disable it until the new menu API makes its way back to Classic. On the plus side however, the "Character Status" option on the players' own menu has been re-added now that we no longer need to deal with LibDropDownExtension.


totalRP3/Modules/Dashboard/StatusPanel.lua
```diff
@@ -1,324 +1,109 @@
...
-local Enums = AddOn_TotalRP3.Enums;
+
 local Events = TRP3_Addon.Events;
 local Globals = TRP3_API.globals;
-local UI = TRP3_API.ui;
 local Utils = TRP3_API.utils;
 
--- Increments the character version data number for a given player model.
---
--- @param The player model to increment version data for.
```
Solanya, line 17: did you mean to delete the doc

totalRP3/Deprecated.xml
```diff
@@ -0,0 +1,19 @@
...
+	SPDX-License-Identifier: Apache-2.0
+
+	Anything in this file is deprecated and will be removed in future versions
+	of TRP, potentially without warning.
+-->
+<Ui xmlns="http://www.blizzard.com/wow/ui/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.blizzard.com/wow/ui/
+https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
+	<Include file="Deprecated.lua"/>
+	<!--
+		Use TRP3_DropdownButton template instead.
```
Solanya, line 12: TRP3_DropdownButtonTemplate

totalRP3/Modules/Register/Characters/PlayerMapScan/PlayerMapPin.lua
```diff
@@ -131,50 +146,11 @@ function TRP3_PlayerMapPinMixin:OnMouseDown(button)
...
-		local info = {};
-		info.text = string.format("%1$s: %2$s", loc.CL_OPEN_PROFILE, pin.playerNameColored);
-		info.func = function() TRP3_API.slash.openProfile(sender); end;
-		info.notCheckable = true;
-
-		MSA_DropDownMenu_AddButton(info);
-	end
-
-	MSA_DropDownMenu_AddSeparator({});
-	MSA_DropDownMenu_AddButton({ text = CANCEL, notCheckable = true });
```
Solanya, line 179: I'll have to check but does that mean no cancel button anymore?

Solanya (dismissed):

totalRP3/Modules/Register/Characters/PlayerMapScan/PlayerMapPin.lua
Meorawr: Yeah - all cancel buttons in modern menus have been removed, and the same will be the case for Classic when it gets them.

Unfortunately this means right now in Classic closing a menu is a bit tricky - you need to click the dropdown button itself to dismiss the menu. For context menus, the same _should_ apply but need to test. Worst case I can bodge the menus for Classic to just force a Cancel button to be re-added.

totalRP3/Modules/Dashboard/StatusPanel.lua
Meorawr: yes

totalRP3/Core/UITools.lua
```diff
@@ -116,162 +116,126 @@ end
...
+	local function SetSelectedElement(elementData)
+		local value = elementData[2];
+		dropdown:SetSelectedValue(value);
 	end
-end
 
-local function listBoxGetValue(self)
-	return self.selectedValue;
-end
+	local function GenerateMenuDescription(menuItems, menuDescription)
```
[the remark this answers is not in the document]
Meorawr: Intentionally electing to do nothing for now.

totalRP3/Deprecated.xml
Meorawr:
```suggestion
		Use TRP3_DropdownButtonTemplate instead.
```

Solanya (approved):

PR #902, opened 2024-06-10
Check launcher action validity for tooltip instructions

Solanya:
When writing the launcher tooltip instructions, we parse each action's name to add it at the bottom. However, if an action was bound to a module (say, Extended) that is subsequently disabled, the action doesn't exist anymore, causing the tooltip to error out.

This change just checks whether the action is valid before trying to add it at the bottom of the tooltip, so the instruction just gets skipped otherwise.


Meorawr (approved):
a most poggersful change

PR #905, opened 2024-06-21
Add companion custom name and color to Tooltips

Raenore:
This adds support for tooltips to show companions their custom names and colors.

Take a companion like this:
Wow_GROaAraMLl

**Before this change,** tooltip with someone targetting that companion had this:
Wow_DPUkRN9svD
Basically OOC name always and the WARRIOR color.

**After this change,** it now properly gets the RP name and color set:
Wow_vA8Fo1kdrO
Look how pretty.

_Don't decimate me too hard, Meow._

### DIFF
--- a/totalRP3/Modules/Register/Main/RegisterTooltip.lua
+++ b/totalRP3/Modules/Register/Main/RegisterTooltip.lua
@@ -765,9 +765,6 @@ local function writeTooltipForCharacter(targetID, _, targetType)
 		local name = UnitName(targetType .. "target");
 		local targetTargetID = getUnitID(targetType .. "target");
 		if targetTargetID then
-			---@type Player
-			local targetTarget = AddOn_TotalRP3.Player.static.CreateFromCharacterID(targetTargetID)
-			local _, targetEnglishClass = UnitClass(targetType .. "target");
 			local unitType = TRP3_API.ui.misc.getTargetType(targetType .. "target");
 
 			local targetClassColor;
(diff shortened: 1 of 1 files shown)


Meorawr:
Let the games begin.

totalRP3/Modules/Register/Main/RegisterTooltip.lua
```diff
@@ -768,14 +768,38 @@ local function writeTooltipForCharacter(targetID, _, targetType)
 			---@type Player
 			local targetTarget = AddOn_TotalRP3.Player.static.CreateFromCharacterID(targetTargetID)
 			local _, targetEnglishClass = UnitClass(targetType .. "target");
-			local targetInfo = getCharacterInfoTab(targetTargetID);
-			local targetClassColor = targetEnglishClass and TRP3_API.GetClassDisplayColor(targetEnglishClass) or TRP3_API.CreateColor(1, 1, 1, 1);
+			local unitType = TRP3_API.ui.misc.getTargetType("target");
```
Meorawr, line 771: This queries the type of the unit you're actively targeting, rather than the thing the tooltip is actually looking at.

Meorawr (changes requested):

totalRP3/Modules/Register/Main/RegisterTooltip.lua
Raenore: Thank you, no clue why I missed that one. Should be fixed in latest commit.

Meorawr (dismissed):
If I were a butthole I'd ask you to reduce the scope of `targetTarget` but I think you've survived this first hazing - both in this PR and in-game.

You did a job, Buzzy. I'm neutral on the matter.

Raenore:
> If I were a butthole I'd ask you to reduce the scope of `targetTarget` but I think you've survived this first hazing - both in this PR and in-game.
> 
> You did a job, Buzzy. I'm neutral on the matter.

I'll do you one better, I also moved `targetEnglishClass` to the proper scope. That's two whole lines for the price of one.

Meorawr (approved):
I am sated. Still won't say anything nice though, Buzzy.

PR #907, opened 2024-06-29
We call these fields "statuses" in TRP, name them so + cleanup

Raenore:
To show the outcome:
Wow_xHFh3neczJ

--> `REG_PLAYER_STATUS`
It felt silly to have these fields called "Currently" while one is.. also called "Currently" and the other is "Other information (OOC)". Besides, we already refer to this as "statuses" in `DB_STATUS_CURRENTLY_COMMON`.

--> `DB_STATUS_CURRENTLY` in favor of `REG_PLAYER_CURRENT`
While renaming "Currently (IC)" to just "Currently" as everything is implied IC. The added benefit is that this means we can once again use this string in the tooltip as its counterpart for OOC used there is `DB_STATUS_CURRENTLY_OOC`.

--> `REG_PLAYER_CURRENTOOC`, `REG_PLAYER_CURRENT_OOC`
Also remove some strings that are not used anywhere, at all.

> [!IMPORTANT] 
> This will require a new localization for `REG_PLAYER_STATUS` and `DB_STATUS_CURRENTLY` to have the `()` parts removed.


Meorawr (dismissed):
Approved noting that we need to resolve the modified localization strings - either in CF, or by force-uploading edited ones from the repo. Also need to purge the deleted keys.

Katoriehooves:
I'm giving this my hoof seal of approval.

Ghostopheles:
I'm giving this 3.5 Tombstones out of 5.

keyboardturner:
i rate this +2/3 jermas

Meorawr:
:AYAYAweird:

Meorawr (approved):

PR #908, opened 2024-06-29
Show RP names in dice roll output

Meorawr:
Currently when we receive a dice signal message the character name printed to the chat is the raw 'sender' string supplied to us by the comms subsystem, which even if the sender is on the same realm typically ends up being an ugly as heck "name-realm" string.

Using my superior intellect and absolutely no prodding from someone else, it has been decided that this is subpar and can be improved. As such, we now display the senders' RP name and color it according to their class appropriately.

image

### DIFF
--- a/totalRP3/Core/Slash.lua
+++ b/totalRP3/Core/Slash.lua
@@ -177,11 +177,14 @@ TRP3_API.RegisterCallback(TRP3_Addon, TRP3_Addon.Events.WORKFLOW_ON_LOADED, func
 			return;
 		end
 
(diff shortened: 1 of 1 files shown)


Solanya (approved):

Raenore:
> Using my superior intellect and absolutely no prodding from someone else, it has been decided that this is subpar and can be improved. As such, we now display the senders' RP name and color it according to their class appropriately.

Yeah of course, totally.

It's a good idea, as RPers do tend to have their char box filled with other people's IC names. So going by IC name, when set, is the best course of acttion.

PR #915, opened 2024-06-30
Adjust fonts without breaking alphabet swaps

Meorawr:
The SetFont API is a trap that has the effect of irreversibly breaking the ability for a fontstring to swap which font file is used for rendering strings when dealing with text content that goes from one language to another, eg. English to Russian.

We've had font utilities for a little while now, so let's fix up all the places calling SetFont and prevent them from breaking.


totalRP3/Core/UITools.lua
```diff
@@ -394,25 +394,21 @@ local function refreshTooltip(Frame)
...
 			TRP3_MainTooltipTextLeft1:SetNonSpaceWrap(true);
 			TRP3_MainTooltipTextLeft1:SetTextColor(1, 1, 1);
 		end
 		if Frame.bodyText then
 			TRP3_MainTooltip:AddLine(Frame.bodyText, 1, 0.6666, 0, true);
-			local font, _, flag = TRP3_MainTooltipTextLeft2:GetFont();
-			TRP3_MainTooltipTextLeft2:SetFont(font, getTooltipSize(), flag);
+			TRP3_TooltipUtil.SetLineFontOptions(TRP3_MainTooltip, 2, getTooltipSize());
 			TRP3_MainTooltipTextLeft2:SetNonSpaceWrap(true);
-			TRP3_MainTooltipTextLeft2:SetTextColor(1, 0.75, 0);
```
Solanya, line 415: Just double checking, this one was removed because it's the default color, right?

Solanya (approved):

totalRP3/Core/UITools.lua
Meorawr: It's still a shade of gold, at least.

PR #916, opened 2024-06-30
Remove info tooltip size setting

Meorawr:
The "Info tooltip size" setting was used to change the size of informational help tooltips shown in the UI.

This setting has a few issues; firstly, the default size is wrong. Compared to tooltips around the rest of the game interface, the size is 11 compared to what should be 12 - so by default our tooltips are less readable for people with bad eyesight.

The more important issue however is that the default value assumes that font heights across different languages are uniform, which they're not. Many game fonts in asian languages - particularly Korean - have their sizes specifically increased to make glyphs clearer.

The adjustment to fonts isn't something that's particularly uniform. At larger sizes these per-alphabet adjustments eventually stop, but at smaller sizes they tend to be greater.

As such, we've elected to remove the setting for now and to see who complains. It should be noted that the game client itself supports DPI scaling in addition to the UI scaling feature - so users on 4k monitors and the like already have plenty of knobs they can tune that work across the entire game, instead of needing to tweak those _and_ our weird setting.

### DIFF
--- a/totalRP3/Core/UITools.lua
+++ b/totalRP3/Core/UITools.lua
@@ -19,7 +19,7 @@ local loc = TRP3_API.loc;
 local floor, tinsert, pairs, wipe, assert, _G, tostring, table, type, strconcat = floor, tinsert, pairs, wipe, assert, _G, tostring, table, type, strconcat;
 local math = math;
 local MouseIsOver, CreateFrame = MouseIsOver, CreateFrame;
-local TRP3_MainTooltip, TRP3_MainTooltipTextRight1, TRP3_MainTooltipTextLeft1, TRP3_MainTooltipTextLeft2 = TRP3_MainTooltip, TRP3_MainTooltipTextRight1, TRP3_MainTooltipTextLeft1, TRP3_MainTooltipTextLeft2;
+local TRP3_MainTooltip = TRP3_MainTooltip;
 local shiftDown = IsShiftKeyDown;
 local UnitIsPlayer = UnitIsPlayer;
 local getUnitID = TRP3_API.utils.str.getUnitID;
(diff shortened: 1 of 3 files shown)


totalRP3/Core/UITools.lua
```diff
@@ -438,7 +414,7 @@ local function setTooltipForFrame(Frame, GenFrame, GenFrameAnch, GenFrameX, GenF
 		if GenFrameAnch then
 			Frame.GenFrameAnch = "ANCHOR_"..GenFrameAnch;
 		else
-			Frame.GenFrameAnch = "ANCHOR_TOPRIGHT";
```
Meorawr, line 441: To ward off the question - intentional. All call sites seem to be specifying this parameter, and most of them set it to either "TOP" or "RIGHT". I've noticed no breakages with this change at all.

totalRP3/Core/UITools.lua
```diff
@@ -19,7 +19,7 @@ local loc = TRP3_API.loc;
 local floor, tinsert, pairs, wipe, assert, _G, tostring, table, type, strconcat = floor, tinsert, pairs, wipe, assert, _G, tostring, table, type, strconcat;
 local math = math;
 local MouseIsOver, CreateFrame = MouseIsOver, CreateFrame;
-local TRP3_MainTooltip, TRP3_MainTooltipTextRight1, TRP3_MainTooltipTextLeft1, TRP3_MainTooltipTextLeft2 = TRP3_MainTooltip, TRP3_MainTooltipTextRight1, TRP3_MainTooltipTextLeft1, TRP3_MainTooltipTextLeft2;
```
Solanya, line 22: could have cleaned more imports while you were there 🧹

totalRP3/Core/UITools.lua
```diff
@@ -373,43 +373,19 @@ end
...
-			TRP3_TooltipUtil.SetLineFontOptions(TRP3_MainTooltip, 1, getTooltipSize() + 4);
-			TRP3_MainTooltipTextRight1:SetNonSpaceWrap(true);
-			TRP3_MainTooltipTextRight1:SetTextColor(1, 1, 1);
-		end
-		do
-			TRP3_MainTooltipTextLeft1:SetNonSpaceWrap(true);
-			TRP3_MainTooltipTextLeft1:SetTextColor(1, 1, 1);
-		end
+		TRP3_MainTooltip:SetOwner(Frame.GenFrame, Frame.GenFrameAnch, Frame.GenFrameX, Frame.GenFrameY);
+		GameTooltip_SetTitle(TRP3_MainTooltip, Frame.titleText);
```
Solanya, line 383: There was no mention of rightText removal but there doesn't seem to be any case where it was used so makes sense

Solanya (approved):

totalRP3/Core/UITools.lua
Meorawr: challenge accepted

totalRP3/Core/UITools.lua
Meorawr: Oh right - yeah, I'd noticed it was entirely unused.

PR #918, opened 2024-06-30
Eliminate egregious and undesirable upvalues

Meorawr:
This isn't a total cleanup of all our upvalues ever - there's a lot of stupid crap still around like "get" and "getValue" but those can be cleaned up when I get around to TRP4.


totalRP3/Core/Slash.lua
```diff
@@ -8,8 +8,6 @@ local AddOn_TotalRP3 = AddOn_TotalRP3;
 
 local loc = TRP3_API.loc;
 local displayMessage = TRP3_API.utils.message.displayMessage;
-local tonumber, math, tinsert, type, assert, tostring, pairs, sort = tonumber, math, tinsert, type, assert, tostring, pairs, table.sort;
```
Meorawr, line 11: An astute reviewer would notice that `sort = table.sort` in this case, but `sort` is a global anyway so it works fine even with this removed.

Solanya (approved):

Issue #178, opened 2018-07-24
Review roleplay experience values


Solanya:
Currently, Experienced roleplayer is the default state for roleplay experience, which probably isn't reflecting the actual experience for players who don't pay attention to that dropdown.

There should be a default value (Either not set, or "normal" roleplayer), with New/Experienced/Volunteer being manually selected. Maybe add an icon for experienced ?

Additionally, this flag should be account-wide.

Connor-Macleod:
Definitely agree with all points here.
I might even suggest having a popup suggesting you pick an option at first login. as currently it's missed by those who most need it (new roleplayers)

Ellypse:
The popup on first login is planned see #179

Raenore:
This should have been resolved in #927.

PR #927, opened 2024-07-02
Make RP experience a global setting

Meorawr:
Logically - it makes no sense for the RP experience field to be something that's profile-specific. Beginner users who want to elect into displaying this status on their profiles need to remember to set it up manually on each individual one, as it always defaults to experienced.

This changeset makes it such that the RP experience field is now stored as a configuration variable (not dissimilar to default profile IDs).

All logic for accessing and mutating the XP field has been relocated to the Player class accordingly. The XP field is no longer transmitted as part of "character" data but is now part of the vernum exchange, similar to the account trial information flag.

The XP status on player-owned profiles is set to nil with a migration to prevent us unnecessarily sending out stale data to older clients. We don't clear it on stored register profiles; the expectation is as people update the XP field will clear itself out progressively.

One downside to this is that the XP field as part of this change will now effectively reset for all users. It'd be a bit problematic to guess which experience level is correct if there's a mixture of profiles, so it's probably best to not do so and instead just make a note in the changelog. This does however give us a chance to look at resolving #178 by adding a fourth option and setting a new default that's not "Experienced".


totalRP3/Core/Player.lua
```diff
@@ -450,6 +450,25 @@ function CurrentUser:SetCustomRace(race)
...
+	-- transmitted as part of character data - so it must trigger a version
+	-- update and register-related events but can't go through the standard
+	-- UpdateProfileField path to do so.
+
+	local character = self:GetInfo("character");
+	local playerName = TRP3_API.globals.player_id;
+	local profileID = self:GetProfileID();
+
+	TRP3_API.configuration.setValue("roleplay_experience", experience);
+	character.v = TRP3_API.utils.math.incrementNumber(character.v or 1, 2);
```
Solanya, line 468: I'm not sure there's a good way around it, but incrementing only the current profile means if someone sees you on another profile instead, they won't get the updated status. Probably unlikely to happen often anyway, just thinking out loud.

Solanya (approved):

totalRP3/Core/Player.lua
Meorawr: Hmm, that's a good point. Not entirely opposed to just incrementing the versions across *all* profiles, but will think.

Solanya (dismissed):

Meorawr:
Let's kill six birds with one bazooka and also take care of https://github.com/Total-RP/Total-RP-3/issues/178 while we're here, as the changes to XP to now be part of the vernum exchange is implicitly incompatible anyway.

Solanya:
Didn't notice there was a question.
Normal sounds weird to me right now, but I don't have a much better word to suggest, so we could just use that as default.

Meorawr:
Updated levels (and also corrected the field label itself):
Wow_2024-07-03_23-32-37
Wow_2024-07-03_23-32-40
Wow_2024-07-03_23-32-43
Wow_2024-07-03_23-32-46

New tooltip icons for retail:
Wow_2024-07-03_23-25-40
Wow_2024-07-03_23-25-44

totalRP3/Modules/Flyway/FlywayPatches.lua
```diff
@@ -323,3 +322,23 @@ TRP3_API.flyway.patches["19"] = function()
...
+	if not TRP3_Profiles then
+		return;
+	end
+
+	for _, profile in pairs(TRP3_Profiles) do
+		local character = profile.player and profile.player.character;
+
+		if character then
+			character.XP = nil;
+			character.v = TRP3_API.utils.math.incrementNumber(character.v or 1, 2);
```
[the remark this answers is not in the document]
Meorawr: It both is and isn't - we want players to send an updated misc out with the XP field removed without having to wait until they touch their CU/CO fields to increment the version.

Solanya (approved):

PR #941, opened 2024-07-04
New tooltip description and rendering system

Meorawr:
This implements the basics of the new tooltip builder logic. This works similarly to menu descriptions in 11.0, where we build up a giant description of the tooltip contents line by line and have methods to control how it ultimately gets rendered.

One thing that this system offers us which we don't have today is the ability to hook our tooltips; tooltip descriptions (and lines) can be tagged, so once migrated this system would allow for modules to modify character tooltips in an extensible manner without being forced to always append contents to the end of the tooltip.


totalRP3/UI/Tooltip/Tooltip.lua
```diff
@@ -0,0 +1,473 @@
...
+
+	if type(anchor) ~= "table" then
+		tooltip:SetOwner(owner, anchor, description:GetAnchorOffset());
+	else
+		tooltip:ClearAllPoints();
+		tooltip:SetPoint(anchor:Get());
+		tooltip:SetOwner(owner, "ANCHOR_PRESERVE");
+	end
+
+	tooltip:SetOwner(owner, "ANCHOR_RIGHT");
```
Solanya, line 445: why are we setting the owner again after setting it right before?

Meorawr: you saw nothing

Solanya (approved):

Issue #484, opened 2021-01-17
[Suggestion!] "Accept Character Theft" option in Miscellaneous!


AwesomeXona:
**Is your feature request related to a problem? Please describe.**
- For role purposes, playing a thief can be a tricky thing. Trade is incredibly important to developing multiple storylines, though it doesn't seem like it's always obviously used. Stealing items, whether through TRP or through emotes can be taboo on occasion - especially if it's interfering with current plotlines.

**Describe the solution you'd like**
In Miscellaneous, there's a "Accept Character Death", "Accept Character Injury", "Accept Character Romance", I'd like to suggest adding a "Accept Character Theft" or "Accept Character Larceny" to the list. This helps rogue/thief characters continue to develop naturally, and give them clear storylines and abilities to follow.

**Describe alternatives you've considered**
At the moment this entirely relies on the working of other people -- adding notes to profiles, or adding to IC/OOC notes. This is an overlooked thing for everyone other than in a rogue type of role, and is often not put into use -- which generates a lot of footwork for RPers to ask OOC permissions beforehand.

**Additional context**

Meorawr:
Generally we've preferred not to add more fields for that page since there's a _large_ number of character archetypes that someone could possibly play out, all of which may have varying ideas on what options would be useful to show there.

To me personally it feels a bit like the proper solution here would be to either allow customization of that page - which would be a significant task - or to somehow push information like that into the existing "Additional Information" system on the Characteristics page.

AwesomeXona:
Fair. Though the option to disclose certain information or not seems really helpful with that. The "Do Not Show" really reduces how much is there. Granted, worst case scenario if that list entirely grows too large, you could limit the use of actual factors. Say only 5-6 options can actually show at a time.

I dunno how all that works, but I would think it'd maybe help streamline it a little bit. c:

PR #944, opened 2024-07-04
Export RP style options via MSP

Meorawr:
This is to enable their support in MSP-based addons down the line. Fields are as follows:

- Accept character injuries ("RI")
- Accept character death ("RD")
- Accept character romance ("RR")
- Accept criminal activity ("RM")
- Accept loss of control ("RO")

Values exported are the same as our dropdown options as-is. May adjust this based upon feedback.


Meorawr:
Tentatively, adding "Accept criminal activity" ("RM") and "Accept loss of control" ("RO") for our plans around #484.

Solanya (dismissed):

Solanya (approved):

PR #945, opened 2024-07-04
Fix progress spinner wobble

Meorawr:
The progress spinner on the tooltip has a bit of a visual wobble when animating, with the cause being that by default it has texel and grid snapping enabled for image sharpening.

Disabling this fixes the wobble and makes the animation much smoother.


Solanya (approved):

Raenore:
But who doesn't like a bit of wobble?

PR #946, opened 2024-07-04
Enable missing phrase deletion for locale uploads

Meorawr:
This could go horribly wrong, or it could go horribly right. Either way if it does go wrong we have all the strings in the repo for restoration.


Solanya (approved):

Solanya:
yolo

PR #950, opened 2024-07-05
Misc. UI Tweaks

Meorawr:
Just a few assorted tweaks that don't really fit in elsewhere. Nothing blocking, but good to have.

- The scrollbars on profile pages now reset to the top when changing the selected profile page.
- T2 icons are now bordered to appease the Holy Rae.
- Icon browser icons are now larger, because whatever I keep telling myself I honestly wish it was bigger.


Raenore (commented):
Some pages should potentially be revisited for some bottom padding, placing it here as per your request.

An example is the Characteristics page:
image

Solanya (approved):

PR #937, opened 2024-07-03
UI Wars Episode 1: The Phantom Backdrop

Meorawr:
This is the first part of the large wave of UI updates for our 3.0.0 version targetting backdrops and UI control skins. Most of these changes are linked together unfortunately and can't be broken down further, for example changing sliders requires changes to the configuration pages, which in turn requires its new backdrops.

Key notes:

- Backdrop borders are now consistently positioned and sized across the whole addon.
- Backdrop borders are now gold, matching the achievement UI.
- Checkboxes, sliders, editboxes, and tabs now use modern styles where appropriate.
- Most scrollbars in the UI are now fixed and no longer bleed outside of their containing frames. Additionally, we also now consistently use the "minimal" slider style. Scrollbars also now hide if not required to be shown.
- Icons shown in profile pages are now bordered.
- Configuration pages have had a visual overhaul.
- Various controls now display tooltips if their textual contents have been truncated.

A picture says a billion words, so here's a bunch.

Wow_2024-07-03_21-57-23
Wow_2024-07-03_21-57-04
Wow_2024-07-03_21-57-12
Wow_2024-07-03_21-57-32
Wow_2024-07-03_22-01-37

PR #958, opened 2024-07-05
Teach ElvUI skinning about parentKeys instead of just frame names

Raenore:
| Before | After  |
|  ----    |  ----   |
| image | image |

This fixes the skinning issues presented with the TWW UI rework (#937) which removed `TRP3_TargetFrameCaptionPanel` but instead use the parentKey of `Header` from `TRP3_TargetFrame`.
To use these for skinning, you use "FrameName.ParentKeyName" within `SKINNABLE_FRAMES` for ElvUI.

There's also some minor additions of semi-colons where they were missing.

### DIFF
--- a/totalRP3/Modules/TooltipSkins/ElvUI.lua
+++ b/totalRP3/Modules/TooltipSkins/ElvUI.lua
@@ -41,7 +41,7 @@ TRP3_API.module.registerModule({
 
 		local SKINNABLE_FRAMES = {
 			"TRP3_TargetFrame",
-			"TRP3_TargetFrameCaptionPanel",
+			"TRP3_TargetFrame.Header",
 			"TRP3_GlanceBar",
 			"TRP3_GlanceBarSlot1",
 			"TRP3_GlanceBarSlot2",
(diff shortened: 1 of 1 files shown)


totalRP3/Modules/TooltipSkins/ElvUI.lua
```diff
@@ -108,24 +108,35 @@ TRP3_API.module.registerModule({
...
 
 			function skinTargetFrame()
 				-- Go through each skinnable frames from our table
 				for _, frame in pairs(SKINNABLE_FRAMES) do
+					local parentKey;
+
+					-- Check if a skinnable frame includes a parentKey
+					if frame:find("%.") then
+						frame, parentKey = frame:match("(.+)%.(.+)");
+					end
```
Meorawr, line 131:
```suggestion
					local parentKey;
					frame, parentKey = string.split(".", frame, 2);
```

Believe this is functionally equivalent - do test.

totalRP3/Modules/TooltipSkins/ElvUI.lua
```diff
@@ -108,24 +108,35 @@ TRP3_API.module.registerModule({
...
+
 					if _G[frame] then
-						-- We check that the frame exists and then add it to ElvUI
-						TT:SecureHookScript(_G[frame], 'OnShow', SetStyleForTargetFrame)
+						-- If parentKey is not nil, check if a frame exists with said parentKey within this frame
+						if parentKey ~= nil and _G[frame][parentKey] then
+							TT:SecureHookScript(_G[frame][parentKey], 'OnShow', SetStyleForTargetFrame);
+						else
```
Meorawr, line 137:
```suggestion
						elseif parentKey == nil then
```
If `_G[frame][parentKey]` doesn't exist, the unguarded `else` could go and hook scripts on entirely the wrong thing. We want to ensure that if the parentKey is defined it's used if available, and otherwise entirely skipped if not found.

Meorawr (changes requested):

totalRP3/Modules/TooltipSkins/ElvUI.lua
Raenore: Confirmed works the same, I'll batch it in.

totalRP3/Modules/TooltipSkins/ElvUI.lua
Raenore: Good catch, I hadn't though about that that possibility.

Meorawr (approved):

PR #960, opened 2024-07-06
Rename Masters to Owners

Solanya:
Most places in the addon refer to the person in charge of a pet as its owner, but this one dropdown in companion profiles.
Not sure it's worth resetting other languages just for that, but we can at least be consistent here.


Meorawr (approved):
get owned

PR #963, opened 2024-07-06
Relationship adjustments

Solanya:
Couple of small adjustments I forgot about regarding relationships:
- The action button on profile view has been made bigger with the newer icon button template.
image

- The target button tooltip has been changed to fit our newly defined tooltip standards. Subject to change if we're not happy with the current version.
image

I also moved the relation button registration to the newly added relations file, rather than its previous existence in the ignore file.


Meorawr (dismissed):
Looks good though I think having an extra blank line between the tooltip body text and instructions would help.

Meorawr (approved):

PR #971, opened 2024-07-08
Add TRP3 toggle window to toolbar

Solanya:
Another certified Rae banger, adding a toolbar button to toggle the main frame.

Short of rewriting the toolbar system to be able to handle the custom file path to the minimap logo (and in turn needing to use the bordered icon template to account for the border inconsistency between regular icons and that one), it was preferred to use a book icon to convey the "main frame" idea. Open to using a different one though.

image


Solanya:
She's freeeeee

PR #977, opened 2024-07-09
Tooltips with just clicks should have no newline

Raenore:
Oops.


Solanya (approved):
Can't believe someone would let that pass review, what were they thinking

PR #982, opened 2024-07-10
Consistency pass of settings pages

Solanya:
- Repeating "settings" on the tab names made them crop when the section title already indicates that these are "Settings". Keys in all languages we could somewhat interpret where changed to remove this.
- As an added thing, "Frames settings" was split into "Toolbar" and "Target Frame".

Two things worth discussing:
- There is the small matter of the map button settings. I moved it in General for now as this was the simplest solution. It's possible it should be in the Directory category instead next to the Location settings, but that requires a lot more changes, so I'd like to double-check if we have an alternative idea first.
- Register was renamed to Directory, which means the non-special categories are not in alphabetical order anymore. Then again, other languages were also not in alphabetical order. Do we still want to change it?

image


Meorawr (approved):
One comment that "Chat frames" is used as the string in enUS but a bunch of others are still "Chat".

Solanya:
Put the name singular and reused the global string from Blizzard. Should be good now.

totalRP3/Locales/frFR.lua
```diff
@@ -224,7 +224,7 @@ Gardez à l'esprit que changer ces paramètres peut altérer votre expérience a
 	["CO_ANCHOR_TOP"] = "Haut",
 	["CO_ANCHOR_TOP_LEFT"] = "Haut-gauche",
 	["CO_ANCHOR_TOP_RIGHT"] = "Haut-droite",
-	["CO_CHAT"] = "Discussions",
+	["CO_CHAT"] = "	Fenêtre de discussion",
```
Meorawr, line 227: Why the leading space bro

Meorawr (changes requested):

Meorawr (approved):
I approve of this ramen.

PR #990, opened 2024-07-11
Added unread icon on About tab

Solanya:
When the About tab is unread, we add the unread icon to it so people get used to it by association.
This will have to be adjusted when we change the tooltip icons.

image

In order to avoid issues with languages that have very long "About" tab names, I added a parameter to change the margin on the text for them. I've been told it will conflict with half the tab changes. We'll just have to find out.

image

### DIFF
--- a/totalRP3/Core/UITools.lua
+++ b/totalRP3/Core/UITools.lua
@@ -722,10 +722,10 @@ local function tabBar_onSelect(tabGroup, index)
 	for i=1, #tabGroup.tabs do
 		local widget = tabGroup.tabs[i];
 		if i == index then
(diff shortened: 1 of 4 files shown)


Meorawr:
I'm tempted to just split out the tab button changes which gives you a proper `:SetIcon()` method on the button itself, as well as fixing the awkwardness you had around disabling tabs.

Solanya:
I can wait for that one and rework it to use the new stuff once it's there, shouldn't take too long to adjust then.

Meorawr (approved):
I hereby approve of my own changes

Raenore:
> I hereby approve of my own changes

images

PR #991, opened 2024-07-12
Add resize cursor while resizing the main window

Solanya:
This is very hacky so please sit down before reviewing.

Added a SetCursor call to the main window resizing. However, because the cursor gets reset when we get out of frames capturing the mouse inputs, moving the cursor to make the frame bigger does not keep the cursor.
To that effect (and this is the hack), I added a mouse-capturing frame to show only when we click on the button and hide when releasing or drag stop. This now lets us keep the resize cursor even when moving out of the frame. 

If this is too much of a hack, down to discard it, but hey at least it works.


Meorawr:
I was going to probably just get rid of our weird shadow frame resize hack anyway and let the frame resize properly. This in turn would already use the correct cursor.

Meorawr (approved):
If I kill the resize shadow stuff I'll do so in another PR; this hack is probably fine for now.

PR #992, opened 2024-07-12
Fix profile errors while still receiving data

Solanya:
- When the tooltip for the target frame button to Open profile was modified to include the character name, the getCompleteName call was given a `nil` fallback name, which makes it error out when targeting someone for whom we haven't yet received the characteristics tab.
I've added the fallback string found in other places in the very file, knowing the target frame gets refreshed once we receive the characteristics data to display the proper name.

- In the same vein, opening a profile while receiving data could throw an error trying to access a character table that wasn't received yet. I just added a check to prevent it from errorring out.


Raenore (approved):
This seems reasonable.. I wonder who didn't test this out extensively before before. 😲

PR #995, opened 2024-07-13
The Great Tooltip Purge (the purge giveth and taketh)

Raenore:
An effort to make all (read: the most common) our tooltips go through a consistent format, sprinkled with some fixes and localization checks.
This is an on-going effort and stuff probably might still change and/or gets missed still.

> [!NOTE]
> Extended might follow soon(TM), I've got a few more things to tackle that isn't specifically a "purge" for that beast (like porting the dropdown changes to Extended).

The format works as follows:
```
- Titles are always white.
- Descriptions are default gold.
- Highlights and/or keywords in Descriptions are green.
- Actions (clicks, what the tooltip's element DOES) are in green.
```

> [!TIP]
> The following best practices were observed (which you can deviate from at any point of time, if you'd like):
> - If a tooltip has both a Description and Actions, the Actions are separated by a whiteline to stand on their own. If there is no Description, there is no whiteline.
> - If a button is self-explanatory (Save, Cancel, Remove), it will not have a tooltip unless its action cannot be inferred from the button itself (for example, an additional right-click menu).
> - A tooltip will open from the top-right edge ("ANCHOR_RIGHT") on the element it is attached to, however there are some exceptions:
>   - The Toolbar and Targetframe tooltips will open right above the elements ("ANCHOR_TOP").
>   - Certain popups where the tooltip would obstruct too much (glance editor), anchor to the left ("ANCHOR_LEFT").
>   - Browsers (icon browser, companion browser) do not have an y 5, they are tied directly to the element they belong to.

> [!IMPORTANT]
> This is not a strict set of rules! Features will come in that may not adhere to these practices at first (or ever), because sometimes things can't or won't fit into a consistent idea. Anyone is free to open a PR to resolve any inconsistencies, which can then be reviewed on a case-by-case basis.

# Changelog
This is just what I remember, it's possible I might've missed a thing or two (or three..).

### General
- A lot of tooltips have received standardized instructions (left-click, right-click, etc.) to properly convey what is possible.

**Glances**
- Clicking on a glance slot will hide its tooltip so there is no overlap with the editor that is opened.

**Profile - About**
- About page received an extra tutorial for the formatting tools.

**Relations**
- Icon picker for relations has been given a tooltip with instructions.

### General
- Color codes were tightened and using global colors where applicable.
- Nearly every tooltip known to mankind has been changed in some shape or form if it deviated from the current format.
- The resize button (main frame, glance editor) tooltip was removed, NOTE: It still requires the cursor to be changed upon drag to inform the users it does something logically.
- Tooltips, except for rare cases, will now always open starting from the top-right edge of an element.

**Companion Browser**
- Companion (Pet/Mount) browser now has the tooltip starting from its element like the Icon browser.

**Profile - About**
- Remove button in About Template 2 had its tooltip removed (action can be inferred from button).

**Profile - Characteristics**
- End of an era: Personality Traits (AKA: Psycho) keys have been renamed from their old French to English.
- Remove button for Misc Info and Traits had its tooltip removed (action can be inferred from button).
- Reorder button for Misc Info and Traits had its tooltip removed (action can be inferred from button).

**Profile - Notes**
- Notes now use your profile name when it states whose notes it is showing, this way the user knows which profile set the notes (as profiles characters can have the same name).
- Notes had their localization changed a bit ("Common Notes" are "Account-wide notes" now), plus a note mentioning that it adds a tooltip (Note: If we have a new note icon, this should be changed!).

**Target Frame**
- Target frame glance tooltips no longer can have their anchor configured, they will now respect our consistent format.
- Unread icon overlay on character, companion and mount target frame buttons has been updated to the new one (Note: might still change again!).

**Toolbar Frame**
- Toolbar tooltips now open above the button instead of to the left of them, mimicking the target frame behavior.

## Fixes

**Directory**
- Actions button on directory is now aligned decently.
- Player tooltips in directory had some color issues, this has been fixed.
- Some tutorial tooltips have been reworked and will not go off the frame even at its minimum size.

**Profile - About**
- About page tutorial button is now correctly removed when you save/cancel editing your templates.
- Character music theme picker actually works, tooltips are no longer cut off and display info reliably.
- `TRP3_RegisterAbout_Edit_Template1` had a duplicate frame (with the same name) that did absolutely nothing (we hope), this was fixed.

**Profile - Characteristics**
- Residence coordinates button now properly explains clicking it will open the world map with a pin.

**Profile - Misc**
- Removed invisible tooltips (did they ever work?) on currently and other information boxes in the Misc tab for other players.

**Relations**
- Relation color picker now has a tooltip with some explanation, like all the other color pickers.

### General
- A lot of other strings that no longer make any sense/do not point to anything.

**Dashboard**
- TRP3 Currently Frame mentions removed, this is now built into the addon (3rd party addon still works of course!).

**Profile - Characteristics**
- Characteristics background changing strings have been removed, this feature never made it in.


totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -1045,7 +1053,11 @@ TRP3_API.RegisterCallback(TRP3_Addon, TRP3_Addon.Events.WORKFLOW_ON_LOADED, func
 			end,
 			onClick = function(unitID)
 				openMainFrame();
-				openPageByUnitID(unitID);
+				if unitID == Globals.player_id then
+					TRP3_API.navigation.menu.selectMenu("main_10_player");
```
Solanya, line 1057: I'm not quite sure what that was supposed to do, I imagine it was meant to be modified so it wouldn't reselect the player menu if that was already the active menu, but right now it just does the exact same thing openPageByUnitID was doing when unitID was the player.

totalRP3/Modules/TargetFrame/TargetFrame.lua
```diff
@@ -133,7 +133,7 @@ local function onStart()
 				uiAlert:Show();
 				uiAlert:SetWidth(buttonSize / 2);
 				uiAlert:SetHeight(buttonSize / 2);
-				uiAlert:SetTexture(buttonStructure.alertIcon);
+				uiAlert:SetAtlas(buttonStructure.alertIcon);
```
Meorawr, line 136: I guess our only uses of alertIcon are for atlases now, but it'd probably be worth supporting them both anyway.

```suggestion
				if C_Texture.GetAtlasInfo(buttonStructure.alertIcon) then
					local useAtlasSize = false;
					uiAlert:SetAtlas(buttonStructure.alertIcon, useAtlasSize);
				else
					uiAlert:SetTexture(buttonStructure.alertIcon);
				end
```

Solanya (dismissed):
Finally through it all.

totalRP3/Modules/Register/Main/RegisterList.lua
Raenore: This wasn't meant to go in, as it does not do anything indeed. Sorry for that, will remove in 'future' push.

Raenore:
> Finally through it all.

Thank you for that, I went through all the things to see if you missed something while you checked for things I missed.
All looks fine so far, but I'm sure there's still stuff hidden that Meow or even more re-reads will find out.

Meorawr (dismissed):
In theory it all looks fine but it now conflicts with the profile manager updates so pls fix :)

Solanya (approved):

PR #998, opened 2024-07-13
Add companion icon to sidebar list entries

Meorawr:
Could have sworn this was always here, and we have an interface icon entry for it, but seemingly we never used it? Did we remove it by accident? Did Rae eat it? idk.

image


Raenore:
Having done some expert digging, I can confirm it wasn't there even before I came on to the project. So I will vehemently deny having touched it, even if the paw looks ruff.

https://github.com/Total-RP/Total-RP-3/blob/c29dfda6790def95ee4968b25616367ca9454d45/totalRP3/Modules/Register/Main/RegisterList.lua#L118

Solanya (approved):

PR #1002, opened 2024-07-13
Disable About/Ignored tabs when no data in them

Solanya:
This is a work in progress waiting for the tab changes to be done so it can be adapted to them, and as such should not be merged now.

- The About tab gets disabled on other players profiles if there is no relevant data in that section (including music theme).
- The Ignored tab gets disabled in the Directory if no player is ignored.


totalRP3/UI/TabSystem.lua
```diff
@@ -43,6 +43,19 @@ function TRP3_TabButtonMixin:SetTooltip(tooltipFunction)
...
+	return GetValueOrCallFunction(self, "locked");
+end
+
+function TRP3_TabButtonMixin:SetTabLocked(locked)
+	self.locked = locked;
+	self:MarkDirty();
+end
+
+function TRP3_TabButtonMixin:MarkDirty()
+	self:SetScript("OnUpdate", self.Update);
```
Meorawr: We set an OnUpdate script here but, seemingly, never unset it in the Update method?

At any rate, I'd prefer that we:

1. Add an OnUpdate method that calls self:Update().
2. Hook that up in XML.
3. Add `onUpdateMode="disabled"` to the template.
4. In MarkDirty, call `self:SetOnUpdateMode(Enum.OnUpdateMode.RunWhenVisibleOnce)`. No need to add any calls to Update.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -569,10 +569,26 @@ local function showDefaultTab()
 	TRP3_Addon:TriggerEvent(TRP3_Addon.Events.NAVIGATION_TUTORIAL_REFRESH, "player_main");
 end
 
+local function getAboutDataExists(profile)
+	if profile and profile.about then
+		if profile.about.MU then
+			return true;
+		elseif profile.about.T1 and profile.about.T1.TX then
```
Meorawr: What about scenarios where `TX = ""`?

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -569,10 +569,26 @@ local function showDefaultTab()
...
+			return true;
+		elseif profile.about.T2 and CountTable(profile.about.T2) ~= 0 then
```
Meorawr: We're linearly counting an entire table to determine emptiness?

```suggestion
		elseif profile.about.T2 and not TableIsEmpty(profile.about.T2) then
```

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -958,12 +958,17 @@ end
...
+local tabGroup;
+
 function refreshList()
 	local lines;
 	local initializer;
 	TRP3_RegisterListEmpty:Hide();
 	TRP3_RegisterListHeaderActions:Hide();
 
+	-- Disable ignored tab when no character ignored
+	tabGroup.tabs[3]:SetTabLocked(CountTable(getIgnoredList()) == 0);
```
Meorawr: We're linearly counting a table to determine emptiness?

```suggestion
	tabGroup.tabs[3]:SetTabLocked(not TableIsEmpty(getIgnoredList()));
```

totalRP3/UI/TabSystem.xml
```diff
@@ -95,6 +97,7 @@
 			<OnLoad method="OnLoad"/>
 			<OnEnter method="OnEnter"/>
 			<OnLeave method="OnLeave"/>
+			<OnLeave method="OnUpdate"/>
```
Meorawr:
```suggestion
			<OnUpdate method="OnUpdate"/>
```

Meorawr (approved):

totalRP3/Core/UITools.lua
```diff
@@ -729,104 +729,150 @@ end
 -- Tab bar
 --*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
-local tabBar_index = 0;
+---@class TRP3.TabGroup
+---@field tabs { [number]: Button }
+---@field enabled { [number]: boolean }
+---@field current number?
+---@field tabBar Frame
+local TabGroup = {};
```
Meorawr, line 737: The tab group code offended me, so I had Copilot prototype-ify it. It did okay, need to update its instructions a bit. Originally, it was really focused on "do the smallest possible change" and literally defined a `local tabGroup = { MethodName = oldLocalFunctionName, ... }` table.

Only functional change here really is adding SetTabEnabled and UpdateTabStates. I opted for this over storing locked state on tabs because the tab group should really be the controller of that state, and it's simpler if it remains the sole caller of SetTabState on individual buttons.

Otherwise, it's just your standard "take the old functions and make them methods on this table" conversion. A few things were reordered, so the diff may look messy.

Tab rework in 2027 I promise.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -473,14 +473,19 @@ local function updateAboutTabIcon(context)
...
 		if not context.isPlayer and profileID == context.profileID then
-			if infoType == registerInfoTypes.ABOUT and tabGroup.current == 2 then
-				showAboutTab();
-				updateAboutTabIcon(context);
+			if infoType == registerInfoTypes.ABOUT then
+				updateAboutTabState(context);
+				if tabGroup.current == 2 then
+					showAboutTab();
+					updateAboutTabIcon(context);
+				end
```
Meorawr, line 488: I don't know why we only update the icon when the tab is selected. That seems like a bug, but I'm not gonna change that.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -569,10 +574,43 @@ local function showDefaultTab()
...
+		elseif profile.about.T2 and not TableIsEmpty(profile.about.T2) then
+			return true;
+		elseif profile.about.T3 then
+			local T3 = profile.about.T3;
+
+			for _, field in ipairs({ "PH", "PS", "HI" }) do
+				if T3[field] and Utils.str.emptyToNil(T3[field].TX) ~= nil then
+					return true;
+				end
+			end
```
Meorawr, line 598: Noting that intermediate PH/PS/HI fields can be nil.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -569,10 +574,43 @@ local function showDefaultTab()
...
+
+function updateAboutTabState(context)
+	local isEnabled = context.isPlayer or getAboutDataExists(context.profile);
+	tabGroup:SetTabEnabled(2, isEnabled);
+
+	-- If an update comes in that deletes About page content while we've got
+	-- that tab selected, swap to the Characteristics tab.
+	if not isEnabled and tabGroup.current == 2 and tabGroup.tabs[1]:IsShown() then
+		tabGroup:SelectTab(1);
+	end
```
Meorawr: Not a huge fan, but I think it's weirder UX in this state to have the user keep a visibly disabled tab selected... Hopefully an edge case.

totalRP3/Modules/Register/Characters/RegisterNotes.lua
```diff
@@ -111,7 +111,7 @@ function TRP3_API.register.inits.notesInit()
 		end
 
 		local openPageByUnitID = TRP3_API.register.openPageByUnitID;
-		local openNotesTab = TRP3_TabBar_Tab_5:GetScript("OnClick");    -- This was a quick workaround for RP.IO, is there a better option ?
+		local openNotesTab = function() TRP3_API.register.player.tabGroup:SelectTab(4); end
```
Meorawr, line 114: Hacks no more.

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -1005,10 +1010,8 @@ end
 -- Init
 --*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
-local tabGroup;
-
 local function createTabBar()
-	local frame = CreateFrame("Frame", "TRP3_RegisterMainTabBar", TRP3_RegisterList);
+	local frame = CreateFrame("Frame", "TRP3_RegisterListTabBar", TRP3_RegisterList);
```
Meorawr, line 1011: TRP3_RegisterMainTabBar is actually a duplicate name used for frames in two files. I changed this one.

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -958,12 +958,17 @@ end
...
+local tabGroup;
+
 function refreshList()
 	local lines;
 	local initializer;
 	TRP3_RegisterListEmpty:Hide();
 	TRP3_RegisterListHeaderActions:Hide();
 
+	-- Disable ignored tab when no character ignored
+	tabGroup:SetTabEnabled(3, not TableIsEmpty(getIgnoredList()));
```
Raenore:
```suggestion
	local hasIgnored = not TableIsEmpty(getIgnoredList());
	tabGroup:SetTabEnabled(3, hasIgnored);
	if not hasIgnored and currentMode == MODE_IGNORE then
		tabGroup:SelectTab(1);
		return;
	end
```
If the ignored list goes empty, we should probably move the user off this tab too.

Given you un-ignore from the ignored list, this'll happen if you only got one person ignored and then remove them.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -569,10 +574,43 @@ local function showDefaultTab()
...
 end
 
+local function getAboutDataExists(profile)
+	if profile and profile.about then
+		if profile.about.MU then
+			return true;
+		elseif profile.about.T1 and Utils.str.emptyToNil(profile.about.T1.TX) then
+			return true;
+		elseif profile.about.T2 and not TableIsEmpty(profile.about.T2) then
+			return true;
```
Raenore:
```suggestion
		elseif profile.about.T2 then
			for _, frameTab in pairs(profile.about.T2) do
				if Utils.str.emptyToNil(frameTab.TX) then
					return true;
				end
			end
```
1. Use Template 2
2. Create a Frame
3. Wipe its default text.
4. Save

It will save as:
`profile.about.T2 = { { TX = nil } }`

Which would make the tab remain clickable, but it would actually be empty.
This fixes that.

Raenore (changes requested):
Tested on live 12.1

- About tab is disabled when empty.
- About tab works fine in other profiles.
- Your own about tab is never greyed out.
- If you're on an about that gets wiped, you are pushed to characteristics.
- Ignored list when empty is greyed out.

Given we always go to tab 1 on new page, meaning there's no real other ways to get 'stuck' on an About we shouldn't, I think I've covered all interactions?

I did mark two things that can/should be improved, with suggestions:
- Ignored tab when empty and disabled does not move you to another tab, my suggestion makes you go back to the first tab (like you did with about).
- Template 2 could still show an empty page due to an empty text frame not being "empty" but carrying `{ TX = nil}' (see the suggestion). My suggestion should handle that as well.

Raenore (approved):
Tested anew on 12.1 live.

- About tab is disabled when empty.
- About tab works fine in other profiles.
- Your own about tab is never greyed out.
- If you're on an about that gets wiped, you remain on it but moving off it makes it disabled.
- Template 2 also is disabled even with 'empty' frame.

Seems in order this time.

PR #1003, opened 2024-07-14
Convert preset additional info & traits to custom

Raenore:
https://github.com/user-attachments/assets/0225bd9a-5d8c-4355-9df3-f1dcd2b6c42a

Basically allows you to convert a preset additional info or trait to be custom. On conversion, it will keep its icon, values, names, etc. It just becomes customizeable.
image

> [!NOTE]  
> The action button SetScript had to be moved out of the `if frame == nil then` block as it would otherwise not know its IDs got changed, which in turn would make it never know to stop hiding "convert to custom" from the user.

If an additional info was one that added something to the tooltip, the user gets a warning that mentions it will be gone after conversion. For the above behavior to work, a `shownOnTooltip` boolean field was added to all `MiscInfoTypeData`.
image

> [!IMPORTANT]
> Two new localizations have been added:
> - `REG_PLAYER_CONVERT`
> - `REG_PLAYER_CONVERT_WARNING`


totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -750,6 +759,27 @@ local function onPsychoDuplicate(self)
 	setEditDisplay();
 end
 
+local function onPyschoConvert(self)
```
Solanya, line 762:
```suggestion
local function onPsychoConvert(self)
```

totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -1267,6 +1300,17 @@ function setEditDisplay()
 			tinsert(psychoEditCharFrame, frame);
 		end
 
+		frame.ActionButton:SetScript("OnMouseDown", function(self)
+			TRP3_MenuUtil.CreateContextMenu(self, function(_, description)
+				-- If not custom, allow convert to custom.
+				if psychoStructure.ID then
+					description:CreateButton(loc.REG_PLAYER_CONVERT, onPyschoConvert, self);
```
Solanya, line 1307:
```suggestion
					description:CreateButton(loc.REG_PLAYER_CONVERT, onPsychoConvert, self);
```

Solanya (commented):
The Meowster has also agreed to show the tooltip on additional info creation for presets that gets shown on the tooltip, if we want to add that before merging.

Meorawr (dismissed):

Meorawr (approved):

Raenore:
For the (lesser) Ancients, I have added icons and tooltips explaining which additional infos show on the tooltip.



(Video was without icons, but it shows the tooltip.)
https://github.com/user-attachments/assets/d87731af-7e2a-4abb-ae52-c34fdc4d6a94

Meorawr (approved):

PR #1006, opened 2024-07-14
Convert profile managers to scrollbox list views

Meorawr:
This converts the player and companion profile managers to use scrollbox list views rather than the old and crappy UI list system that forced those things to be memes for years and only support a fixed number of frames.

The code is a mess, but that's because frankly rewriting the entire profile manager would take a lot longer than we really have time for and I just want to be done with this shit. Additionally while there's some changes for the profile importer, that has not been converted to a list view because it's another special beast entirely. I've no idea how that tab even looks right now with these changes.


totalRP3/UI/ProfileManager/ProfileManagerList.lua
```diff
@@ -0,0 +1,85 @@
...
+	ScrollUtil.AddManagedScrollBarVisibilityBehavior(self.ScrollBox, self.ScrollBar, scrollBoxAnchorsWithBar, scrollBoxAnchorsWithoutBar);
+end
+
+function TRP3_ProfileManagerListMixin:OnShow()
+	self.CreateButton:SetText(L.PR_CREATE_PROFILE);
+	self.ScrollBox.EmptyText:SetText(L.PR_PROFILEMANAGER_EMPTY);
+end
+
+function TRP3_ProfileManagerListMixin:GetSearchText()
+	return self.SearchBox:GetText("");
```
Solanya, line 36: Is there a reason for the empty string here

Solanya (dismissed):

totalRP3/UI/ProfileManager/ProfileManagerList.lua
Meorawr: copy/paste

Meorawr, line 36:
```suggestion
	return self.SearchBox:GetText();
```

Solanya (approved):

PR #1009, opened 2024-07-14
Don't register the button when target frame doesn't exist

Solanya:
Forgot to check that the target frame was enabled before registering the relation button when I moved it around. This restores the same test that is done everywhere else.


Meorawr (approved):
very pog

PR #1011, opened 2024-07-14
You sit on a throne of lies - only left-click works

Raenore:
image

will-ferrell-elf


Solanya:
Ok but if you're gonna change that one, at least change the one in the companion list too

Raenore:
> Ok but if you're gonna change that one, at least change the one in the companion list too

No.


...Okay.

Raenore:
Will this tyranny never end of me being forced upon with more work..

image

Solanya (approved):
Never

PR #1012, opened 2024-07-14
Make profile list less snappy

Meorawr:
Don't update the scroll position each time the list is rebuilt, rather just do so when the search text changes or upon entering the tab.


Solanya (dismissed):

Raenore (approved):
Seems to be in working order after the latest commit. Good job Meow.

PR #1021, opened 2024-07-16
Mark empty abouts as read (combined flyway with inactive glances)

Raenore:
About changes were always marked as unread, which is fine 9/10 of the time. However due to an incorrect check in `onAboutReceived` we also applied unread to empty abouts. These should not marked as unread.

This includes a flyway (21) to update people's directory to mark all the profiles with empty abouts that were wrongly marked as unread.. As read.

Updated: This also adds to the same flyway the removal of inactive glance data that we were not intending to send (handled in #980).

### DIFF
--- a/totalRP3/Modules/Flyway/FlywayPatches.lua
+++ b/totalRP3/Modules/Flyway/FlywayPatches.lua
@@ -359,7 +359,7 @@ TRP3_API.flyway.patches["21"] = function()
 	local profileList = TRP3_Register.profiles;
 	for profileID, profile in pairs(profileList) do
 		-- Don't check default profiles
-		if not TRP3_API.profile.isDefaultProfile(profileID) and profile.characteristics and next(profile.characteristics) ~= nil then
+		if not TRP3_API.profile.isDefaultProfile(profileID) then
 			-- Handle the wrong about unread situation first.
 			local aboutData = profile.about;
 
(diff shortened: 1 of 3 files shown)


totalRP3/Modules/Flyway/FlywayPatches.lua
```diff
@@ -342,3 +342,77 @@ TRP3_API.flyway.patches["20"] = function()
...
+				local filledDescription = false;
+				if aboutData.TE == 1 then
+					local templateData = aboutData.T1 or {};
+					filledDescription = templateData.TX and strtrim(templateData.TX):len() > 0;
+				elseif aboutData.TE == 2 then
+					local templateData = aboutData.T2 or {};
+					local atLeastOneFrame = false;
+					for _, frameTab in pairs(templateData) do
+						if frameTab.TX and strtrim(frameTab.TX):len() > 0 then
+							atLeastOneFrame = true;
```
Meorawr, line 379:
```suggestion
							atLeastOneFrame = true;
							break;
```

No need to continue looking at other parts of T2 once we've found one.

totalRP3/Modules/Flyway/FlywayPatches.lua
```diff
@@ -342,3 +342,77 @@ TRP3_API.flyway.patches["20"] = function()
...
+					end
+					filledDescription = atLeastOneFrame;
+				elseif aboutData.TE == 3 then
+					local atLeastOneFrame = false;
+					local templateData = aboutData.T3 or {};
+					local datas = {templateData.PH, templateData.PS, templateData.HI};
+					for i=1, 3 do
+						local data = datas[i] or {};
+						if data.TX and strtrim(data.TX):len() > 0 then
+							atLeastOneFrame = true;
```
Meorawr, line 390:
```suggestion
							atLeastOneFrame = true;
							break;
```

Meorawr (changes requested):

Meorawr (approved):

Solanya (approved):

PR #1033, opened 2024-07-18
Fix Template 2 edit background

Solanya:
The backgrounds for template 2 frames weren't properly set when showing template 2. We're adding a call to set the proper background on each frame of template 2 when showing it, similar to template 3.

As an addendum, since setBkg is literally just a call to TRP3_API.ui.frame.setBackdropToBackground with the same parameters, I just yeeted the function and called the original one everywhere.

### DIFF
--- a/totalRP3/Modules/Register/Characters/RegisterAbout.lua
+++ b/totalRP3/Modules/Register/Characters/RegisterAbout.lua
@@ -104,17 +104,13 @@ getDefaultProfile().player.about = {
 
 local draftData;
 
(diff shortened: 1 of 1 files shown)


Solanya:
Maybe Rae suggested the function replacement, if I don't mention that I'll get in trouble probably

Meorawr:
This was an excellent idea Sol, very smart of you to come up with it by yourself

Solanya:
Thanks

Meorawr (approved):

PR #1036, opened 2024-07-18
Add option to display full relations in tooltips

Meorawr:
This expands the "Show relation" option converting it to an enum and allows displaying the full relation description in the tooltip if set to the new "Show full description" value.

The default is "Show name only" which matches the existing behavior of only displaying a 'Relation: Potato' line in the tooltip.

Flyway code from the boolean setting is not included; we've not done a release with the new setting, so not much point.

image
image


Solanya (approved):

Raenore (approved):
We should probably clean up "Do not show" dupliccate keys as well before 3.0.0.

Also somewhat make the dropdowns consistent in options.. But I know my place.

Meorawr:
Thanks for volunteering to do that.

Raenore:
> Thanks for volunteering to do that.

a8f3cfa6718a94a4c004bc3c83cb243fdc9ac625

PR #1037, opened 2024-07-19
Reset isFading on showing the toast

Solanya:
As soon as the toast starts fading once, it will never fade again, as isFading was never reset. We're now resetting it as soon as we show the toast.

Might have been noticed by some void elf whose name starts with an R.


Meorawr (approved):
> Might have been noticed by some void elf whose name starts with an R.
Weird way of spelling Meorawr.

PR #1038, opened 2024-07-19
Popups on escape now just close themselves

Raenore:
https://github.com/user-attachments/assets/b61ea850-fc15-4896-87e1-d82580603864

This fixes the fact that escaping a popup would close the whole TRP frame with it as well. Now it behaves similarly to the close button on the popups itself.

Also the popups code is a freaking nightmare, do not recommend.

### DIFF
--- a/totalRP3/Core/Popup.lua
+++ b/totalRP3/Core/Popup.lua
@@ -1285,7 +1285,7 @@ function TRP3_API.popup.showPopup(popupID, popupPosition, popupArgs)
 	popup.frame:SetScript("OnKeyDown", function(_, key)
 		-- Do not steal input if we're in combat.
 		if InCombatLockdown() then return; end
-		
+
 		if key == "ESCAPE" then
 			PlaySound(SOUNDKIT.IG_MAINMENU_OPTION_CHECKBOX_ON);
 			popup.frame:SetPropagateKeyboardInput(false);
--- a/totalRP3/Core/Popup.lua
+++ b/totalRP3/Core/Popup.lua
@@ -766,6 +766,20 @@ function TRP3_API.popup.showCompanionBrowser(onSelectCallback, onCancelCallback,
 	else
 		TRP3_CompanionBrowserTitle:SetText(loc.REG_COMPANION_BROWSER_MOUNT);
 	end
+	-- Do not steal input if we're in combat.
+	if not InCombatLockdown() then
+		local frame = TRP3_CompanionBrowser;
+		frame:SetScript("OnKeyDown", function(_, key)
+			if key == "ESCAPE" then
+				PlaySound(SOUNDKIT.IG_MAINMENU_OPTION_CHECKBOX_ON);
+				frame:SetPropagateKeyboardInput(false);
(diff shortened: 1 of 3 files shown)


Meorawr (commented):
I'll look at this tomorrow but one thing we need to consider is that some (all?) popups support cancellation callbacks when the close button is pressed. An escape press should count as a cancel operation and slowly invoke those callbacks if set.

Raenore:
> I'll look at this tomorrow but one thing we need to consider is that some (all?) popups support cancellation callbacks when the close button is pressed. An escape press should count as a cancel operation and slowly invoke those callbacks if set.

I'll cancel you.

But I can look into this, I guess.

Raenore:
I have checked most of these cases, and in just about every situation we currently seem to be feeding `nil` to the browsers for cancel callbacks.

However, I think I've at least figured out how to handle the callbacks specifically for Companion and Pet browsers in the latest commit (I could be wrong ofc, this is all very weird code).
For the others, I don't currently know how to do it properly, so I'll wait for your input/suggestions.

totalRP3/Core/Popup.lua
```diff
@@ -766,8 +766,25 @@ function TRP3_API.popup.showCompanionBrowser(onSelectCallback, onCancelCallback,
...
+		frame:SetScript("OnKeyDown", function(_, key)
+			if key == "ESCAPE" then
+				PlaySound(SOUNDKIT.IG_MAINMENU_OPTION_CHECKBOX_ON);
+				frame:SetPropagateKeyboardInput(false);
+				-- Hiding frames & Cancel callback is handled within.
+				onCompanionClose();
+			else
+				frame:SetPropagateKeyboardInput(true);
+			end
+		end);
```
Meorawr, line 785: The InCombatLockdown check needs to be inside the SetScript callback, not outside.

Meorawr (changes requested):

totalRP3/Core/Popup.lua
Raenore: Should be resolved in latest commit, thank you.

Raenore:
> If cancel callbacks are used nowhere I'd be tempted to just remove support for them entirely. Alternatively, the method of closing each popup could be changed to just `:Click()` each close button - but that feels disgusting.

I mean, if they all are nil/not handled right now then I feel it might just be better to just leave it as-is until the rework?

If the new escape options don't do stuff, and the normal callbacks don't either, there will be a zero sum change at the end. /shrug

Meorawr:
Only issue with leaving them as-is is the risk of us forgetting that they don't consistently work - but similarly, I suspect it unlikely that we'll run into that trap anyway so I'm probably fine with leaving things as-is.

Solanya (approved):

Raenore:
Allow me to block my own PR for a second to see if I can implement escape closing on the glance editor too.

Solanya:
Sure

Raenore:
Done.

Meorawr (approved):

Solanya (approved):

PR #1044, opened 2024-07-20
Rework notification icons

Meorawr:
This reworks the notification icons (again) to try and resolve some issues we didn't much like during our testing.

- The walk-up icon is no longer a smiley face. Instead we've gone with an icon that's a bit more like the three-person-group indicator that's used by Blizzard in stuff like the group manager UI.
- The unread notification is no longer an "!" as it was a bit too confusing given the use of quest assets. Instead we've gone for two icons that both incorporate an "\*", with the tooltip asset in particular being a book with a "\*" at the top-right.
- The note icon has been slightly touched up and now has a consistent style with the other indicators.
- A currently unused "edit" pencil icon is included for future use.

image
image
image


Raenore (approved):
Seems to be all in order, nice add of the note on the notes page before I sent this.

Just the same reminder I put on Discord:
We need to make sure to eventually implement the "unread description" book icon in the About for unread so people will learn to associate it unconsciously.

Solanya (approved):

PR #1047, opened 2024-07-20
Add an absolute unit of an edit button

Meorawr:
This tweaks our "Edit" button for characteristics and about pages to instead be a big red button with a pencil icon inside of it. The rationale here is that our "Edit" button doesn't actually fit its text in certain locales and looks stupid as a result, so let's try something a bit special.

Wow_2024-07-20_17-27-53


Solanya (approved):

Raenore (changes requested):
The tooltip for this button should either be:
- Gone, as its intent is clear (and its placement is the same as previous edit).
- Have a description, as we try to avoid just "title text" tooltips where possible.

Raenore (approved):

Solanya (approved):

PR #1050, opened 2024-07-20
All dropdowns get consistent + Some delete relation stuff

Raenore:
Dropdowns will generally be:
Change nothing > Do not show > Change something > Change all (the bigger/more permissive the change, the lower in the list. Where applicable!).

----

Make the "Delete relation" option red as it is a destructive one.

Deleting relations can also hurt your feelings maybe. :(


Meorawr (approved):

Raenore:
I will be adding to this.

Raenore:
Add the fact that in use relations will have "Delete relation" greyed out.

Mind you, due to how the menu list works.. If you disable an option it has no "hover" state so the tooltip doesn't actually trigger.

Solanya (approved):

Meorawr (approved):
legally reviewing

PR #1054, opened 2024-07-21
Tooltip on about unread tab

Meorawr:
This adds a tooltip when hovering the About tab while its unread icon is present to better explain what the icon means. Unlike the initial attempt, this tooltip will show irrespective of what part of the tab button you hover over - so you don't need to solely mouseover the icon to get an explanation.

image

Also tweaks the toolbar button tooltips to use long-form text and omit names for all but mounts.

image
image
image


Solanya (approved):

Raenore (approved):
Looks fine, let's do this again sometime.

PR #1055, opened 2024-07-21
Only test Y coordinate axis when dragging MI/PS

Meorawr:
The old hit rect logic for detecting which element the cursor was over when dragging MI/PS lines was rather strict and wanted the user to not drag their cursor outside of the horizontal bounds of the item.

This was a bit dumb - if the user has clicked and begun dragging, the only thing we should really care about is whether or not they've moved the cursor up or down.

We'll no longer test the left/right bounds of the rect and the horizontal positioning of the cursor. This resolves the issue we were seeing where dragging MI/PS fields and having your cursor toward the left edge of the drag button would result in nothing being moved.

Also I cleaned up half the comments because to hell with rewriting them.


Solanya (approved):

Raenore (approved):
Tested and confirmed, ship it.

But honestly damn you for removing annotations.

Meorawr:
1v1 me

PR #1056, opened 2024-07-21
Toggle IC/OOC status on toolbutton click

Meorawr:
The toolbar button for toggling RP status for some reason uses an onMouseDown handler rather than onClick. This is inconsistent with every other button on the toolbar which wait for the "up" part of the click.

Thus, bring harmony and consistency to the world once more and make this button only work on clicks. Additionally route the status change via the Player class so we've got one less awkward place that's manually modifying profile fields and incrementing vernums.


Solanya (approved):

Raenore (approved):
"For menus I'd suggest onmousedown because it's consistent with blizzard" - Meorawr

PR #1057, opened 2024-07-21
Debounce CU/CO field edits

Meorawr:
When a user edits their CU or CO fields we don't debounce inputs, and so each keystroke immediately applies the change to the field and increments the vernum rapidly.

We now only commit changes to these fields 250ms after users stop inputting text for them, both from within the Misc. panel and from our new shiny "Currently" editor on the toolbar. This should stop the vernum field cycling quite so eagerly and us firing so many events while editing the fields.

Additionally as with #1056 the Misc. page was directly modifying these fields instead of going through the Player class. This is no longer the case, so we've got two less instances of manual vernum updating going on.


Solanya (approved):

Raenore (dismissed):
This is fine, but I feel considering people tend to remove everything and then type a few words (or even lines).. you could probably even put the debounce at 500ms or an actual second.

Raenore (approved):

PR #1058, opened 2024-07-21
Bump TOC versions

Meorawr:
The game client supports comma-delimited interface versions on all clients at this point, so we can now just go down to using the single `Interface` line with all the versions and disable TOC splitting in the packager.

Note that Classic users will need to be told to fully exit the game before updating. If they upgrade while the client is open, TRP3 will disappear from their addon list until they restart the game. Reason being that with TOC splitting being disabled the "_Cata.toc" and "_Vanilla.toc" files in the release packages will no longer exist, but the client doesn't process file deletions until a full restart. For Retail users they should notice nothing as the packager doesn't create "_Mainline.toc" files for us today.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Create Package
         uses: BigWigsMods/packager@v2
         with:
-          args: -d -S -z
+          args: -dz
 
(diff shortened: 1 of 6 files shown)


Solanya (approved):

Raenore (approved):
Sounds like a "skill issue" for Classic.

PR #1060, opened 2024-07-21
Use fixed height for notification lines in all TTs

Meorawr:
Rae's tweak to make notifications use a fixed height of 10 only applied to tooltips shown for player characters and not companions, which can show the line for unread abouts.


Solanya (approved):

Raenore (approved):
Keeping you guys on your toes and teaching you all to be more consistent.

Good job!

PR #1063, opened 2024-07-21
Colon-delimit mount info in tooltip

Meorawr:
Looked a bit weird.

image


Solanya (approved):

Raenore (approved):
Interesting way to write; "Thanks Rae for pointing it out, I didn't even notice!"

PR #1064, opened 2024-07-21
Fix map scan context menu parent

Meorawr:
Rae broke it


Solanya (approved):

Raenore (approved):
You both reviewed it.

Meorawr:
No, we both looked at it and fell asleep.

PR #1066, opened 2024-07-21
Remove internal logging buffer

Meorawr:
Our logging functions write to three sinks - any chat frame named "Logs", the DebugLog addon, and an internal circular buffer.

The intent of the circular buffer sink was for debugging and maybe exposing recent log messages in the addon as part of a troubleshooting tool, however in the couple of years since adding this we've never once directed a user to run the necessary commands to dump the buffer nor have we added tooling to visit it.

As such - let's just remove it. We can always revive the buffer later on if we need it, but there's no point having all users pay a memory cost for our logging that they'll never benefit from otherwise.

Also killed a few unused utility functions with this.


Solanya (approved):

Raenore (approved):
"Who needs logging anyways?"

PR #1074, opened 2024-07-21
Fix guild membership style localization

Raenore:


Solanya (approved):

Raenore:
> wow, no description, not even a "thanks Sol for noticing it"
Thanks Sol.

Meorawr (approved):

PR #1077, opened 2024-07-21
Don't overwrite CU/CO with other peoples data

Meorawr:
It passed review you tools


Raenore (approved):
This seems in order now.

We passed the review because we trust you.. A wrong choice to make.

tumblr_medl7nNZDR1qdeqbpo2_500

Solanya (approved):

PR #1079, opened 2024-07-21
Centralize soundkit definitions and add more sounds

Meorawr:
This defines a new global (TRP3_InterfaceSounds) for any common sound kits that we're using on interactions across the UI. Not all sounds have entries in this table - if a specific interaction is unique enough to have its own soundkits, it's free to call PlaySound with the SKID directly. This is more just a centralization of the *commonly used* ones, or ones that are otherwise a pain in the rear to manage.

Additionally, some additional sounds are now played upon interactions:

- Dragging glances now plays a sound effect similar to dragging MI/PS fields.
- The color picker "Presets" button no longer double-plays a sound.
- The music browser plays a click sound when selecting a song.
- The mature profile editor plays click sounds on various interactions.


totalRP3/Modules/Toolbar/Toolbar.lua
```diff
@@ -50,9 +50,9 @@ function ToolbarMixin:Toggle()
 	self.forcedVisibility = nil;
 
 	if self:IsShown() then
-		TRP3_API.ui.misc.playUISound(SOUNDKIT.IG_MAINMENU_OPEN);
+		PlaySound(SOUNDKIT.IG_MAINMENU_OPEN);
 	else
-		TRP3_API.ui.misc.playUISound(SOUNDKIT.IG_MAINMENU_CLOSE);
+		PlaySound(SOUNDKIT.IG_MAINMENU_CLOSE);
```
Solanya, line 55: Should those not use one of the TRP3_InterfaceSounds?

Meorawr: I'd thought about it but the toolbar is a bit of a special case, so I'm happy to just let it directly go out and figure out its soundkits for now.

Solanya (approved):
k

PR #1080, opened 2024-07-22
Limit maximum open directory pages

Meorawr:
This limits the number of maximum directory pages that can be opened at once, with the goal of preventing users from too easily opening a lot of profile pages and clogging up their menus unnecessarily.

At the moment this is hardcoded to only keep open the five most recent pages. We could consider making this a setting if users request it. The limit is shared across both player and companion profiles.


Solanya (approved):

Raenore (changes requested):
This doesn't work correctly when you remove a profile from the middle of the list.

Meorawr:
Yeah that'd require a menu rewrite to fix, or at least uncomfortable changes to somehow resync the list across individual item closes. Shame.

PR #1084, opened 2024-07-22
Fix tutorial on profiles list

Raenore:
Imagine if the thing that's meant to teach you how stuff works.. Doesn't work..


Meorawr (approved):
this is ghost's bug

Solanya (approved):

PR #1085, opened 2024-07-22
Remove old changelogs

Solanya:
While working on the translations, realized we still had a few old changelogs lying around. Might as well throw them all out.


Meorawr (approved):
gigabytes of memory saved

PR #1087, opened 2024-07-22
Convert music browser to use scrollbox list view

Meorawr:
This converts the music browser to use a scrollbox based listview instead of our old crusty and ugly list view logic.

The browser isn't changed too much apart from a small modernization of the code behind it otherwise, so should remain familiar to users. 

A few small usability changes:
- The search filter now behaves similarly to the icon browser in that it'll perform searches with punctuation characters stripped and case insensitively.
- The tooltip now formats durations in human readable durations rather than seconds.

image


Meorawr:
Alright good to go, do your worst.

totalRP3/UI/Browsers/MusicBrowser.lua
```diff
@@ -0,0 +1,160 @@
+-- Copyright The Total RP 3 Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+local L = TRP3_API.loc;
+
+local LRPM12 = LibStub:GetLibrary("LibRPMedia-1.2");
+
+local function GenerateSearchableString(str)
+	return string.utf8lower(string.trim((string.gsub(str, "%p+", " "))));
```
Solanya, line 9: As per #1090 there was mention of lower being enough, don't know if you care about replacing it here or if you're just waiting for it to be centralized

Solanya (approved):

totalRP3/UI/Browsers/MusicBrowser.lua
Meorawr: Yeah just waiting.

PR #1088, opened 2024-07-22
Remove special handling of empty custom class on MSP

Solanya:
Once upon a time, special handling was added for the custom class field coming from MSP, presumably so that it wouldn't use the game fallback and instead just show no class, as this was how MRP displayed it.

The Kat Archives mention that the custom class field was only added to MRP in BfA. Presumably, the behaviour was unintended and MRP was fixed to use the game class as a fallback, but we never got the memo. As such, this has been broken forever and we wouldn't show a class if the field was left empty.

I've removed the special handling so that we use the game class as a fallback once again.

### DIFF
--- a/totalRP3/Modules/Register/MSP/RegisterMSP.lua
+++ b/totalRP3/Modules/Register/MSP/RegisterMSP.lua
@@ -371,10 +371,6 @@ local function onStart()
 					local value = data[field];
 					if value then
 						value = emptyToNil(strtrim(value));
-						-- Preserve empty class field.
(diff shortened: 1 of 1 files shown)


Meorawr (approved):
Not sure I understand the change but I'm going to assume you tested and saw whatever it was working.

Solanya:
Before it would put " " as the class so that it wouldn't be an empty string, therefore not using game class as fallback and appear with no class.
Now, we leave the class string empty, so it uses the game class as fallback once again.

Meorawr:
Ah, right.

Raenore (approved):

PR #1090, opened 2024-07-22
Add centralized search string cache

Meorawr:
The "make-searchable-string" function in all the browsers is similar-ish to a function already present in StringUtil, so let's make everything use that instead and also add a cache there to speed up the generation of search strings.

Note I've worked this atop the music browser changes in #1087. We could merge this in for 3.0, but until that's merged this will remain a draft and the diff will be a bit messy.

### DIFF
--- a/totalRP3/Core/StringUtil.lua
+++ b/totalRP3/Core/StringUtil.lua
@@ -27,10 +27,22 @@ function TRP3_StringUtil.SortCompareStrings(a, b)
 	return strcmputf8i(a, b) < 0;
 end
 
-local function CanonicalizeSearchString(str)
-	str = string.gsub(str, "[%p%c%s]", "-");
-	str = string.utf8lower(str);
+local function GenerateSearchableString(str)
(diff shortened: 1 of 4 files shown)


totalRP3/Core/StringUtil.lua
```diff
@@ -27,12 +27,24 @@ function TRP3_StringUtil.SortCompareStrings(a, b)
 	return strcmputf8i(a, b) < 0;
 end
 
-local function CanonicalizeSearchString(str)
-	str = string.gsub(str, "[%p%c%s]", "-");
-	str = string.utf8lower(str);
+local function GenerateSearchableString(str)
+	str = string.gsub(str, "[%p%c%s]+", "");
+	str = string.lower(str);
```
Meorawr, line 32: To also note; at some point in recent client history string.lower became UTF-8 aware. We don't know when this happened, but it means that we can stop using the slow-as-balls `string.utf8lower` library function generally.

Solanya (approved):

PR #1099, opened 2024-07-26
Add extra highlight to LoC

Raenore:
image

Given that it's a more verbose explanation already, I feel we can get away with an extra highlight for people quickly 'hovering' over it and want to know its most obvious examples.

The highlight is also far enough away from the other highlight that it works without much issues, at least in my opinion.

Also updated the baguette version with this change already.

Is this a nit? Yes.. Yes it is.


Meorawr (approved):
It's a good nit.

Solanya (approved):
We're not highlighting the mugged example for criminal activity. I don't know how relevant it is of a comparison, but it's the only thing that bothers me a little bit about that change. Not enough to block it though. 🤷

Raenore:
> We're not highlighting the mugged example for criminal activity. I don't know how relevant it is of a comparison, but it's the only thing that bothers me a little bit about that change. Not enough to block it though. 🤷

See point two:
image

It would live way too close the previous highlight with only a few words in-between. which also spawns a way bigger highlight which in turn will make people ignore it as at that point you've marked half the text as a highlight in green.

Solanya:
I'm aware of point 2, and not saying mugged should be highlighted. I'm just not sure the example for LoC absolutely needed a highlight. But again, not super attached one way or another. 🤷

Raenore:
> I'm aware of point 2, and not saying mugged should be highlighted. I'm just not sure the example for LoC absolutely needed a highlight. But again, not super attached one way or another. 🤷

I think the reason I can mostly place it for myself is because everyone knows what "theft" entails when you read that word.

"Loss of control" is something less clearly defined, so giving actual examples that might aid people's perception of it seems more logical.

PR #1105, opened 2024-07-29
Issue profile requests on target bar interactions

Meorawr:
This adjusts the "Open profile" buttons for players and companions to issue a profile request when clicked, to avoid any edge cases where someone updates their profile after having been stared at for 5 minutes by someone else who hasn't bothered to mouseover them even once.

Also I s/unitID/characterID/g'd. Get over it.


Raenore:
What a genius person who came up with the fact we shouldn't give people stale/old data if they're our target and we open their profile from the bar, huh!

Raenore (approved):

Solanya (approved):

PR #1108, opened 2024-07-31
Toolbar Modernization (and also IC/OOC icons)

Meorawr:
What begun as a simple foray into making the IC/OOC icons less problematic for people who suffer from colorblindness escalated into a discovery that the hacks around padding icons in the layout logic of the toolbar were absolute nonsense that didn't work unless you had the exact number of buttons on the toolbar that we seem to enable by default, and with the same row size as the default.

So I basically said "fuck this" and rewrote most of the toolbar code to properly separate out all the responsibilities. Buttons now have mixins and are responsible for handling all their own stuff themselves without requiring us to basically micromanage their displays from five different locations of the main toolbar script. The toolbar frame itself outsources the layout logic to a combination of ResizeLayoutFrame and AnchorUtil.GridLayout, completely solving the woes with the added button padding.

A few notable changes:

- Removed onMouseDown support from toolbar buttons to try and force us into consistently using onClick handlers.
- Added untested support for a `tooltipInstructions` key to button descriptions that can be used to populate tooltip instruction lines without having to do the stupid amounts of string concatenation logic that we're doing elsewhere.
- The tooltip code is using our new TooltipScriptMixin and tooltip description logic. Yay.

Note that this isn't the full toolbar rewrite from ages back, but it gets us like 50% of the way there and will make implementing button reordering and drag-drop a bit easier.

### DIFF
--- a/totalRP3/Core/Configuration.lua
+++ b/totalRP3/Core/Configuration.lua
@@ -485,7 +485,5 @@ end);
 AddOn_TotalRP3.Configuration = {}
 
 function TRP3_API.configuration.constructConfigPage()
-	TRP3_API.configuration.registerConfigurationPage(TRP3_API.configuration.CONFIG_TOOLBAR_PAGE);
-	TRP3_API.configuration.registerConfigurationPage(TRP3_API.configuration.CONFIG_TARGETFRAME_PAGE);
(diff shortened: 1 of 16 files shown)


Raenore:
After that misordering fix, ElvUI is back to normal.

Will review the rest in a bit while playing around with it.

Solanya (approved):

Raenore (approved):

PR #1109, opened 2024-08-01
Update crop text tooltip

Raenore:
We have more (and updated) croppings since 3.0.0, but our settings tooltip didn't reflect that.

Where known translations exist for these options (Name, Pronouns, Volice ref, etc.) is where I applied these.

Translations that didn't exist for certain words (mostly guild name/rank) have been kept in English and can be translated later.


Meorawr (changes requested):
If we're doing this I'm tempted to make a few suggestions;

1. Those `FIELDS_TO_CROP` tables in RegisterTooltip.lua are (essentially) duplicated three times. Make the one from `writeTooltipForCharacter` global (call it `TRP3_TooltipCroppingConstants`) and adjust all of the `FIELDS_TO_CROP` lookups to consult that table.
2. Convert these localization strings to just have format specifiers like %d, and then format in the crop sizes from that global table. That way we won't forget to update them in the future, and don't need to deal with importing localization into CF.

Meorawr:
As an alternative I'd be totally down for just also _removing_ the crop lengths from the settings tooltip entirely.

totalRP3/Modules/Register/Main/RegisterTooltip.lua
```diff
@@ -84,6 +84,17 @@ local ConfigKeys = {
...
+TRP3_TooltipCroppingConstants = {
+	Title = 100,
+	Name = 50,
+	Race = 30,
+	Class = 30,
+	Pronouns = 30,
+	GuildName = 30,
+	GuildRank = 30,
+	VoiceReference = 30,
+}
```
Meorawr, line 96:
```suggestion
};
```

WHERE'S THE LAMB SAUCE

Raenore:
Did the global and removed the info from the tooltip setting as it'd just bloat if we do add more cropping.

Meorawr (changes requested):

Meorawr (approved):

Solanya (approved):

PR #1114, opened 2024-08-03
Remove transmission support for PS VAs

Meorawr:
This removes support for sending the old 0-6 range VA field as part of trait comms. The VA field was deprecated and replaced by V2 all the way back in 2018, so there should be nothing on the network that requires it.

For receipt, we still process the VA field if there's no V2 field in viewed profile data. The cost of keeping this around is minimal for now, and we can kill it off some other time if that ever changes.

Additionally some improvements have been made to our serialization of custom trait colors for transmission to try and reduce wire consumption.

Unfortunately when custom colors were added they were transmitted in RGB table form rather than hex colors, so are subject to a bit of blow-up if color components are a bit too exact. We now try and round them to 2 significant digits prior to transmission, so there's a bit of lost information but it should be mostly accurate.


Solanya (commented):
Something that looks weird to me:
- The flyway patch talks about keeping the rendering code but clearly deletes the VA field, so keeping the rendering code is only relevant to displaying the values on the session the old profile is imported.
- Shouldn't we just convert the VA field to V2 upon import if it exists (and V2 doesn't) instead of keeping the old rendering code then?

Meorawr:
I suspect no one has an import string old enough that just has a VA field and no V2, so testing a change to the import logic would be a pain anyway. Honestly - maybe we just kill the VA support completely and say 6+ year old profile imports aren't supported.

Raenore (commented):
I wonder if profiles from 6 years ago have nothing bigger to worry about than this VA field..

Perhaps 6 year old profiles is a very fair cut-off date to give to the users, instead of retaining code for this specific case that will likely affect less than even a single percent?

Honestly in favor of getting rid of it, to be fair.

Solanya (approved):

Raenore (approved):
TRP's on a diet it seems, losing all this old weight.

PR #1115, opened 2024-08-12
Fix constant related error in pet browser

Meorawr:
Blizzard keep renaming the constants used to determine the number of pet slots a hunter can have, so we're just going to hardcode it and hope they never bother to add more stable slots.


Raenore (approved):
Keeping 200 pets at once sounds unethical, but issue's fixed.

> we're just going to hardcode it and hope they never bother to add more stable slots.

Blizzard reading this going "Bet".

Solanya (approved):

PR #1116, opened 2024-08-12
Don't propagate buttons on map pins in combat

Meorawr:
This is mostly required to work around Blizzard's dumb logic where they try to initialize our pins on acquisition and configure their button passthrough while in combat, resulting in blocked script errors - however we also call that function ourselves for player map scan pins so we're equally at fault.


Raenore (approved):
Yeetus deletus the error on map scan.

Solanya (approved):

PR #1118, opened 2024-08-13
Fix Plater mod

Ghostopheles:
There were deprecated functions being used in there. It's a dark place there - you don't want to go looking.


Solanya (approved):
One day we should probably put the actual code or data somewhere on the repo

Meorawr (approved):

PR #1119, opened 2024-08-14
Fix pin button passthrough for 11.0.2

Meorawr:
Blizzard changed the logic for map pins such that they attempt to configure button passthrough *after* the OnAcquire method is called on the pin.

This causes issues with our player map scan as we only respond to right-clicks for that pin, and disable the passthrough in its OnAcquire handler. Blizzard then goes and resets the passthrough to include right-clicks again, thus eliminating our change entirely.

To resolve this we can just nop out the CheckMouseButtonPassthrough method which is called by Blizzard to apply the passthrough change on their end.


Solanya (approved):

DJScias (approved):

DJScias:
Review was unintentional, I was looking at the code (As I had this issue myself) and clicked that by accident.

Please disregard.

Raenore (approved):

Solanya:
> Review was unintentional, I was looking at the code (As I had this issue myself) and clicked that by accident.
> 
> Please disregard.

You're now legally liable for any issue that happens following this fix.

PR #1120, opened 2024-08-14
Color swatch tweaks

Meorawr:
Aligns the color swatch with the residence button rather than indenting the text, and also reduces its size very slightly in the characteristics page.

image


Solanya (approved):

Raenore (approved):
TotalRP is saved!

PR #1121, opened 2024-08-14
Add monochromatic contrast level

Meorawr:
Had intended to add this with the characteristics changes but forgot. This adds a new color contrast level ("Monochrome") that if selected will force text to either use pure white or black depending upon the background color.

In our actual use cases, this is currently always white.

### DIFF
--- a/totalRP3/Core/Color.lua
+++ b/totalRP3/Core/Color.lua
@@ -475,6 +475,7 @@ TRP3_ColorContrastOption =
 	MediumHigh = 5,
 	High = 6,
 	VeryHigh = 7,
+	Monochromatic = 8,
 
 	-- Aliases and other magical values.
 	Default = 3,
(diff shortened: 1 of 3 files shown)


totalRP3/Core/Color.lua
```diff
@@ -604,7 +606,7 @@ function TRP3_API.GenerateReadableColor(foregroundColor, backgroundColor, target
 	local FgYs = CalculateLuminance(FgR, FgG, FgB);
 	local FgLc = CalculateLightnessContrast(FgYs, BgYs);
 
-	if math.abs(FgLc) < targetContrast then
```
Meorawr, line 607: Note; only reason for this change is to just skip the contained loop which attempts to brighten the color until it meets the target contrast.

As the contrast for monochromatic mode is infinity, the loop is pointless and we can bypass it and just go straight to the snapping logic right afterwards.

Solanya (approved):

Raenore (approved):

PR #1123, opened 2024-08-14
Kill menu test code

Meorawr:
This was meant to be removed before merging the menu stuff in but seemingly got forgotten; was only used for testing the compatibility layer for old menus -> new system.


Raenore (approved):
Oh no this broke the whole addon.. Is what I'd like to say.

Funny how we didn't notice this before.

PR #1124, opened 2024-08-14
Improve error display logic for modules

Meorawr:
Module functions like onInit, onStart, or onDisable can return a (bool, string) pair that act similarly to the returns of a pcall to trigger a "soft error" for the module and display it as failed in the module listing.

Some of our modules use this when gracefully disabling themselves, but confusingly end up being displayed as errored in the module listing - chief among them are the Blizzard nameplates when they've been disabled due to the presence of other known nameplate addons.

We now support returning a module status enum code from the module functions in-place of the leading boolean, which will force the status of the module to appear as returned.

Additionally to better support the use case by Blizzard nameplates, a new status code has been added for conflicted modules. These will display as grey in the list rather than appearing as errors.


Solanya (approved):

Raenore (approved):
Accepted due to no conflict

PR #1125, opened 2024-08-14
Fix hiding of icons and titles on Kui nameplates

Meorawr:
The Kui nameplate logic determines if it "can" customize a nameplate based upon both upon our internal requirements for its state (ie. is it initialized, does it have a unit, etc.) but also based upon whether or not it's a personal nameplate or a nameplate for a unit that has a valid reaction.

The problem is that this combined logic results in us not triggering the name text update for nameplates that have transitioned from "can-update" to "cannot-update". This results in the icon and full title updates never being processed, as these are invoked from within the name text update hook.

This commit splits the "can" logic into "can" and "should", with most functions converted to checking "should". When updating a nameplate if we determine that we shouldn't update it, we'll forcefully update the icon and full titles anyway in order to properly hide them.

### DIFF
--- a/totalRP3/Modules/NamePlates/NamePlates_Kui.lua
+++ b/totalRP3/Modules/NamePlates/NamePlates_Kui.lua
@@ -195,7 +195,7 @@ function TRP3_KuiNamePlates:OnNamePlateCreate(nameplate)
 end
 
 function TRP3_KuiNamePlates:OnNameplateNameTextUpdated(nameplate)
-	if not self:CanCustomizeNamePlate(nameplate) then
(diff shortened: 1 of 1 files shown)


Meorawr:
Alright, well at the very least nothing errors or _looks_ broken with these changes. Testing if it fixes the actual issue however is a pain as it's quite random.

Solanya (approved):

Raenore (approved):
Went around with this on in Stormwind, I can't really trigger issues so I feel that even if it has or hasn't fixed stuff..  At least it isn't error-ing.

Might fix that rare "nameplate has the wrong guild attached to an NPC or player" issue I have.

PR #1127, opened 2024-08-15
Rename menu settings

Solanya:
We're going to forget again otherwise, we settled on "Unit menus" as the new name.
Not resetting the keys but could do with an update on the German one which is the only other one for Latin character languages.


Raenore (dismissed):
I want to deny this on the basis one of your commits is called `fr fr`.

Raenore (approved):
I was forced to contribute to this PR, I'm approving it for that reason.

PR #1129, opened 2024-08-17
Scrolling HTML rework

Meorawr:
Opening this PR now so that we can at least get the fundamentals merged in while I work on the integration of these changes into the About page.

This set of changes is aimed at unifying the various HTML utility mixins and templates that we've got lying around the place into one supreme system that rules over all others.

The new `TRP3_HTMLFrameTemplate` can be applied to SimpleHTML frames and acts as a lightweight prefab with sensible font selections, line spacing, and applies a mixin that exposes a few common-sense methods for assigning textual content of varying types - plain text, "rich text" (ie. our markup), and raw HTML text. This region will automatically resize its height to fit the displayed contents, and additionally will do so if it is resized.

Note that changes to other properties of the frame (such as fonts) will *not* trigger an automatic re-layout; one can call the `:Layout()` method to force an immediate update in such a case, or `:MarkDirty()` to batch it at the end of the tick.

For cases where we need scrollable HTML content, the new `TRP3_HTMLScrollFrameTemplate` can be used. This is a Frame (note, not ScrollFrame) template that contains a scrollbox, scrollbar, and internal content region that derives from a user-controllable SimpleHTML template - defaulting to `TRP3_HTMLFrameTemplate`.

As with the HTML frame template, assigning text/html/"rich text" contents fo this frame automatically handles the sizing of the content region. If the frame is resized, the contents will be refreshed appropriately and the scrollable extent updated. The pan extent of the scrollbox is derived from the height of paragraph text, defaulting to scrolling 2.5 lines of text per tick of the mouse wheel. As we're using scrollbox for the internals, there's full support for configuring the padding around the HTML content region (defaulting to 10 units on each edge) and interpolated scrolling is also permitted.

Both templates support a slightly extended version of the hyperlink handler framework that was present in the dashboard code. Hyperlinks displayed in the content regions are expected to be of the form `type:data`, and link handlers can be associated with individual link types via the `:CreateLinkHandler(type, clickCallback[, tooltipCallback])` method on the frames.

The LinkUtil.lua file drives this entire infrastructure and provides a "standard" link handler set that we use commonly everywhere. This set defines link handlers for Twitter URLs ("twitter:profile") and external URLs. Both of these open a new link copy dialog.

Default link tooltip
Link copy dialog

With all this - the Dashboard code has been simplified and consolidated quite a bit, and the HTML views it was previously using have been fully migrated to the new system. As the contents of these views are quite dense in terms of text, interpolated scrolling has been enabled. We also now display bulletpoint images for list markup in these tabs, however *for now* this support is exclusive to the dashboard and is not generalized for use in profiles.

Dashboard view updates

https://github.com/user-attachments/assets/9b82344a-398b-4b80-96fb-977c3adbe316


totalRP3/Modules/Dashboard/DashboardChangelog.lua
```diff
@@ -206,67 +171,8 @@ Thanks to {col:00ff00}Raenore{/col} for the help on some of these features, and
...
-
-	-- Register all the URL strings that allow people to toggle settings
-	-- from the view directly. Whoever thought of this feature was a genius,
-	-- by the way.
-	for url, handler in pairs(URL_HANDLERS) do
-		self.dashboard:RegisterHyperlink(url, handler);
-	end
+function TRP3_DashboardUtil.GenerateChangelog()
+	local text = CHANGELOG_TEXT;
+	text = TRP3_DashboardUtil.GenerateListMarkup(text);
```
Solanya, line 176: Not that it matters but couldn't it have just been `local text = TRP3_DashboardUtil.GenerateListMarkup(CHANGELOG_TEXT);`

Meorawr: Yeah, leftover from some extra stuff that the code was doing when I put in bulleted lists I think.

Solanya (approved):

PR #1130, opened 2024-08-17
Fix tooltips for party units on other maps

Meorawr:
The tooltip logic when the mouseover unit changes explicitly consults the "mouseover" unit or falls back to the world cursor support in mainline when enabled.

For units on different maps to the current player, the client never associates the "mouseover" unit to them - it'll instead display the tooltip with their party/raid unit token which we've never checked.

The current tooltip unit logic has been adjusted to consult the default GTT if shown and use its unit if set, before then falling back to the "mouseover" token and world cursor support.

**Note:** This may in fact break things. Extensive testing required because this logic is garbage.

### DIFF
--- a/totalRP3/Modules/Register/Main/RegisterTooltip.lua
+++ b/totalRP3/Modules/Register/Main/RegisterTooltip.lua
@@ -1446,9 +1446,15 @@ end
 local function GetCurrentTooltipUnit()
 	local unitToken;
 
-	if UnitExists("mouseover") then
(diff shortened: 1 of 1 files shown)


Raenore:
I have been testing this for the past two nights during radiant echoes and moving about maps. So far it's worked.

Not green checking it just yet as I'm sure I can find more extensive test cases still, but figured I'd report that in already.

Solanya (approved):

PR #1131, opened 2024-08-17
Add framebuffer attribute to main frame

Meorawr:
Mostly for Rae's benefit and whoever else is now deciding to alpha-ify our main frame when moving.

As our main frame is a composition of various texture regions, adjusting the alpha of the frame has a few ugly side effects where individual textures representing the borders and sidebar divider end up blending together. This can be seen below - note the top and bottom of the dividing bar between the sidebar and the main content area.

Before

This can be resolved by just enabling the framebuffer attribute on the frame which will force the frame to be rendered to a target at full opacity before being composited into the final UI scene with transparency applied.

Net result - none of the borders end up blending together. As an extra add, as Keyboard pointed out our sidebar texture was smaller than its actual space so the anchors have been adjusted slightly to compensate. See below for the after.

After

### DIFF
--- a/totalRP3/UI/Main.xml
+++ b/totalRP3/UI/Main.xml
@@ -205,7 +205,7 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
 		</Frames>
 	</Frame>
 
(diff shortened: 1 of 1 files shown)


Solanya (approved):

Raenore (approved):
Fade deez frames.

PR #1144, opened 2024-09-09
Upgrade upload-artifact action

Meorawr:
I'm not even gonna bother trying to ascertain what they changed that's necessitated two major version bumps. In fact, I might go into a tirade about how absolutely dire the state of CI tooling broadly is if someone doesn't hold me ba--

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,7 +30,7 @@ jobs:
         with:
           args: -dz
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


Meorawr:
Alright it works and it only required one superfluous commit. I sure love that the intended path for debugging CI's is to constantly commit and push incremental changes, because having local execution and debuggers is so 90's.

Raenore:
> I'm not even gonna bother trying to ascertain what they changed that's necessitated two major version bumps. In fact, I might go into a tirade about how absolutely dire the state of CI tooling broadly is if someone doesn't hold me ba--

<Holds you back.>

Solanya (approved):

PR #1154, opened 2024-12-15
Add warning when chat is disabled

### LINKED ISSUE #1152: Add warning when chat is disabled
Disabling chat also disables addon messages, so it could be useful to add a warning somewhere that warns players who disable chat that the addon will not work.

A combination of CHAT_DISABLED_CHANGED and/or C_SocialRestrictions.IsChatDisabled will likely do the trick.

Solanya:
Fixes #1152.

A warning appears in chat when logging in or changing the "Disable chat" setting if chat is disabled.

image

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -312,6 +312,12 @@ stds.wow = {
 			},
 		},
 
+		C_SocialRestrictions = {
+			fields = {
+				"IsChatDisabled",
+			},
+		},
+
 		C_Spell = {
 			fields = {
 				"DoesSpellExist",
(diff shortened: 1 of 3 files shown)


Meorawr (approved):
Maybe just a nit on the wording; I think we can drop the "Warning: " prefix.

Meorawr (approved):
Seems fine - out of interest, does the warning on login actually work on a fresh client start? 

Wondering if it might only appear after a reload, if chat restriction data is for some reason not available early on in the leoading process.

Raenore (approved):

Solanya:
> Seems fine - out of interest, does the warning on login actually work on a fresh client start?
> 
> Wondering if it might only appear after a reload, if chat restriction data is for some reason not available early on in the leoading process.

It does seem to.

PR #1155, opened 2024-12-16
Improve toolbar object addon compatibility

Ghostopheles:
Addons like ElvUI, praise be to their name, use the `text` field on LDB objects when creating displays for those objects. Our toolbar objects were missing this field, so they'd show up in the DataText selection but would be completely blank.

I've added a function that allows objects to signal when they're updated, as well as added `text` attributes to some of the buttons people might want to use as DataTexts. Maybe it's worth going through and standardizing the way these LDB objects are handled but until then, this makes the objects functional for LDB displayer addons.

### DIFF
--- a/totalRP3/Modules/Toolbar/Toolbar.lua
+++ b/totalRP3/Modules/Toolbar/Toolbar.lua
@@ -47,16 +47,32 @@ local function onStart()
 
 	local LDB = LibStub:GetLibrary("LibDataBroker-1.1");
 
+	-- in order to turn a toolbar object into an LDBObject that can be used by other addons (ElvUI) to display information:
+	-- 	modify the object's 'text' attribute to the desired display text
+	-- 	call TRP3_API.toolbar.SignalLDBObjectUpdate, passing the object as the only arg
+	-- 	ensure the default 'text' attribute is a suitable placeholder, as it will be shown temporarily when the object is updated (at least in the case of ElvUI)
+
+	local OBJECT_NAME_FORMAT = TRP3_API.globals.addon_name_short .. " — %s"; -- what is this dash character??
+
+	local function SignalLDBObjectUpdate(object)
+		local name = OBJECT_NAME_FORMAT:format(object.configText);
+		local callbackName = "LibDataBroker_AttributeChanged_" .. name;
+		LDB.callbacks:Fire(callbackName, name, nil, object.text, object);
+	end
+	TRP3_API.toolbar.SignalLDBObjectUpdate = SignalLDBObjectUpdate;
+
 	---
 	-- Register a Databroker plugin using a button structure
 	-- @param buttonStructure
 	--
 	local function registerDatabrokerButton(buttonStructure)
+		local objectName = OBJECT_NAME_FORMAT:format(buttonStructure.configText);
 		local LDBObject = LDB:NewDataObject(
-			TRP3_API.globals.addon_name_short .. " — " .. buttonStructure.configText,
+			objectName,
 			{
-				type= "data source",
+				type = "data source",
 				icon = Utils.getIconTexture(buttonStructure.icon),
+				text = buttonStructure.text,
 				OnClick = function(Uibutton, button)
 					if buttonStructure.onClick then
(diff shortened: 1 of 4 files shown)


totalRP3/Modules/Toolbar/Toolbar.lua
```diff
@@ -47,16 +47,39 @@ local function onStart()
...
+		-- have to fire both of these events because some addons only listen to one or the other - peak design
+		local callbackNames = {
+			"LibDataBroker_AttributeChanged_" .. name,
+			"LibDataBroker_AttributeChanged_" .. name .. "_" .. attr,
+		};
+		for _, callback in ipairs(callbackNames) do
+			LDB.callbacks:Fire(callback, name, attr, object[attr], object);
+		end
+	end
+	TRP3_API.toolbar.SignalLDBObjectUpdate = SignalLDBObjectUpdate;
```
Ghostopheles, line 69: You would probably think that manually firing these callbacks is unnecessary, as the objects seem to emit their own events when their respective fields are updated.

However, for whatever reason, only icon field changes get picked up by LDB displayer addons. Naturally, I tried to fix this by manually triggering the `LibDataBroker_AttributeChanged_<name here>` callback and specifying "text" as the updated field.

Now we've run into a bit of a conundrum. Some addons don't listen to that event, and instead only register callbacks for the slightly more specific version `LibDataBroker_AttributeChanged_<name>_<field>` - resulting in the need to trigger both callbacks for our objects to actually be updated correctly across all* addons.

*only ElvUI and TitanPanel were tested

Ghostopheles, line 69: You would probably think that manually firing these callbacks is unnecessary, as the objects seem to emit their own events when their respective fields are updated.

However, for whatever reason, only icon field changes get picked up by LDB displayer addons. Naturally, I tried to fix this by manually triggering the `LibDataBroker_AttributeChanged_<name here>` callback and specifying "text" as the updated field.

Now we've run into a bit of a conundrum. Some addons don't listen to that event, and instead only register callbacks for the slightly more specific version `LibDataBroker_AttributeChanged_<name>_<field>` - resulting in the need to trigger both callbacks for our objects to actually be updated correctly across all* addons.

*only ElvUI and TitanPanel were tested

Meorawr:
I don't see why we're manually firing callbacks still - they're fired automatically. If other addons aren't listening to them then us firing them twice by force accomplishes nothing.

Ghostopheles:
> I don't see why we're manually firing callbacks still - they're fired automatically. If other addons aren't listening to them then us firing them twice by force accomplishes nothing.

well boy do I have news for you

totalRP3/Modules/Toolbar/Toolbar.lua
```diff
@@ -46,17 +46,20 @@ local function onStart()
 	--*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
 	local LDB = LibStub:GetLibrary("LibDataBroker-1.1");
+	local OBJECT_NAME_FORMAT = TRP3_API.globals.addon_name_short .. " — %s"; -- what is this dash character??
```
Meorawr, line 49:
```suggestion
	local OBJECT_NAME_FORMAT = TRP3_API.globals.addon_name_short .. " — %s";
```

totalRP3/Modules/Toolbar/Toolbar.lua
```diff
@@ -92,6 +94,7 @@ local function onStart()
 		assert(LDBButton, "Could not find a registered LDB object for id " .. buttonStructure.id)
 
 		LDBButton.icon = Utils.getIconTexture(buttonStructure.icon);
+		LDBButton.text = buttonStructure.text; -- need to manually update each attribute here in order to correctly propagate object changes
```
Meorawr, line 97:
```suggestion
		LDBButton.text = buttonStructure.text;
```
Superfluous comment; it should be no surprise that a function whose sole purpose is to update LDB attributes has to indeed update attributes.

Meorawr (changes requested):

Meorawr (approved):

PR #1158, opened 2024-12-31
Fix Plater optional friendly nameplates handling

### LINKED ISSUE #1157: Name plates don't work when Plater is enabled but Plater Friendly Player module is not enabled
## Steps to reproduce

- TRP3 and Plater AddOns are enabled.
- In Plater, `Friendly Player` > `Module Enabled` is **unchecked**. (For context, I never enable this because it seems completely broken in instanced content.)
- Enable IC status in TRP3.

## Expected behaviour

Since the Plater Friendly Player name plates module is not enabled, TRP3 friendly name plates should still work and display the TRP names and icons when IC.

## Actual behavior

TRP3 friendly name plates no longer work. If I disable the Plater addon, TRP3 friendly name plates will work again.

## Workaround

For now, I have resorted to commenting out `"Plater",` (line 98) from `Modules/NamePlates/NamePlates_Blizzard.lua`.

## Issue checklist

- [x] The issue occurs with version 3.0.6 of the addon.
- [ ] The issue occurs with Total RP 3 as the only enabled addon.

Ghostopheles:
Friendly nameplates are an optional module in Plater, like with ElvUI - this PR ensures that if friendly nameplates are disabled in the Plater config, that the base Blizzard nameplates module can still come in and save the day. This fix feels icky to me since it's a hack on top of a hack but it can't really be fixed otherwise without wider changes to the module system or the two modules themselves. Might be nice to have a system to allow modules to negotiate with each other to see who gets to load. Fixes #1157

### DIFF
--- a/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
+++ b/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
@@ -95,7 +95,6 @@ function TRP3_BlizzardNamePlates:OnModuleInitialize()
 
 	local addons = {
 		"Kui_Nameplates",
-		"Plater",
 		"TidyPlates",
 	};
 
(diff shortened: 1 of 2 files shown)


Meorawr (approved):

Solanya (approved):

Fylwind:
Thank you!

PR #1160, opened 2025-01-18
Add Group and Category TOC directives

Meorawr:
Initial translations for these categories are machine-translated, but they look consistent with Blizzard's own strings as used for realm category names with the exception of ptBR; however the ptBR translation is much longer so it might be the case that Blizzard specifically chose a much shorter string so as to avoid UI issues.

Explicit Groups have been defined on both TOCs to prevent any cases where the auto-grouping might for some reason decide to break our addons apart, eg. if someone made an addon called "tot" and made it depend upon "totalRP3".

image

I'd like some native speakers to confirm these strings before we use them ideally, and for us to remain consistent with (read: edit and replace) [this list here](https://warcraft.wiki.gg/wiki/Addon_Categories#Roleplay) so that we're the authoritative source of all things RP.

### DIFF
--- a/totalRP3/totalRP3.toc
+++ b/totalRP3/totalRP3.toc
@@ -9,8 +9,8 @@
 ## Category-esES: Juego de rol
 ## Category-esMX: Juego de rol
 ## Category-frFR: Jeu de rôle
-## Category-itIT: Gioco di ruolo
-## Category-koKR: 롤플레이
+## Category-itIT: Gioco di Ruolo
+## Category-koKR: 롤플레잉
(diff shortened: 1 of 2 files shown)


Meorawr:
Names should be fine to go - itIT and koKR didn't have community translations applied so I've just stolen Blizzard's globalstrings for them instead.

Yes the casing is inconsistent, but it's inconsistent in-game amongst these locales too. I'm not going to be the one to tell the Italian or Portuguese community how to case their phrases.

Solanya (approved):

Raenore (approved):

PR #1165, opened 2025-02-26
Fix error on opening walkup/proficiency/automation action dropdowns

Solanya:
In dropdowns where we have some options with an attached texture and some without, we decided to pass nil, which doesn't work anymore due to GetAtlasInfo throwing a tantrum with nil.

Since we used that trick a couple times, I'd rather add the safety check at the start of the AttachTexture function than add empty strings around. However, that one call that explicitly had a nil passed and nothing else can go.


Meorawr:
I'd rather just fix up the call sites than add an early return check.

Meorawr:
I'll make you grumble some more.

Go back to AttachTexture and re-add the early return, but precede it with an `assertsafe(icon)` call.

Meorawr (approved):

PR #1173, opened 2025-03-09
Color options (right-click copy color, color name, paste color)

Raenore:
https://github.com/user-attachments/assets/d5443740-cfce-4fbb-8656-da2dc849c6d1

Adds a right-click menu for our colorpickers (Extended follows after).
> [!NOTE]  
> These options work across your player profile, companions, relations and settings.

There are three options:
- Copy color
Copies the color from the current colorpicker frame to an internal clipboard of sorts to be pasted later.
- Copy color name  
Opens a popup where you can quickly copy the color's hex to send it to someone or put it in another colorpicker if necessary.
- Paste color
Pastes the color you had copied earlier with "Copy color" to the current colorpicker frame.

> [!TIP]  
> Colorpicker locations:
> **Player**
> *Profile*
> - Class color swatch
> - Eye color swatch
> 
> *Characteristics*
>  - Class color
>  - Eye color
> 
> **Companion**
> - Information
>  - Name color
> 
> **Relations**
> - Relation color
> 
> **Settings**
> - OOC color
> - Title text color
> - Main text color
> - Secondary text color

> [!IMPORTANT] 
> The following localizations were added and require NEW translations:
> - `REG_PLAYER_COLOR_TT_OPTIONS`
> - `REG_PLAYER_COLOR_TT_COPY`
> - `REG_PLAYER_COLOR_TT_COPYNAME`
> - `REG_PLAYER_COLOR_TT_PASTE`
> 
> These are in-line with how the icon options were made.


Meorawr (changes requested):
I don't like that every single color picker is copy/pasting 90% of the same code when it doesn't need to.

The only difference in the right click handling between them is the paste action, which could  just be implemented as an optional callback stored on the button widget itself.

Also I'm going to complain about the change from OnClick -> OnMouseDown.

Raenore:
> I don't like that every single color picker is copy/pasting 90% of the same code when it doesn't need to.
> 
> The only difference in the right click handling between them is the paste action, which could just be implemented as an optional callback stored on the button widget itself.
> 
> Also I'm going to complain about the change from OnClick -> OnMouseDown.

Made into a mixin and OnClick change made.

totalRP3/UI/Widgets.xml
```diff
@@ -467,6 +456,7 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
 			<PostClick>
 				PlaySound(TRP3_InterfaceSounds.ButtonClick);
 			</PostClick>
+			<OnMouseDown method="OnClick"/>
```
Meorawr, line 459:
```suggestion
			<OnClick method="OnClick"/>
```

totalRP3/UI/Widgets.xml
```diff
@@ -404,7 +404,7 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
 	</Button>
 
 	<!-- Color picker button -->
-	<Button name="TRP3_ColorPickerButton" virtual="true">
+	<Button mixin="TRP3_ColorPickerButtonMixin" name="TRP3_ColorPickerButton" virtual="true">
```
Meorawr, line 407:
```suggestion
	<Button name="TRP3_ColorPickerButton" mixin="TRP3_ColorPickerButtonMixin" virtual="true">
```

Meorawr (changes requested):

totalRP3/UI/Widgets.xml
```diff
@@ -467,6 +456,7 @@ https://raw.githubusercontent.com/Meorawr/wow-ui-schema/main/UI.xsd">
 			<PostClick>
 				PlaySound(TRP3_InterfaceSounds.ButtonClick);
 			</PostClick>
```
Meorawr, line 458: If I were a butthole I'd also want the other scripts relocated to the mixin, but I'm happy leaving these where they are for now.

Raenore: You're not a butthole? I've done this regardless.

totalRP3/UI/Widgets.lua
```diff
@@ -201,3 +203,82 @@ end
...
+		local copyNameButton = rootDescription:CreateButton(L.REG_PLAYER_COLOR_TT_COPYNAME, OnCopyColorClicked);
+		copyNameButton:SetEnabled(hexa ~= nil);
+
+		local pasteButton = rootDescription:CreateButton(L.REG_PLAYER_COLOR_TT_PASTE, OnPasteColorClicked);
+		pasteButton:SetEnabled(TRP3_API.GetLastCopiedColor() ~= nil);
+
+		if hexa then
+			rootDescription:CreateButton("|cnRED_FONT_COLOR:" .. L.REG_PLAYER_COLOR_TT_DISCARD .. "|r", function() self.setColor(nil, nil, nil); end);
+		end
+	end
```
Solanya, line 273: I don't know if anyone cares, but technically this whole block from `local hexa` to here is only relevant if it's a RightButton click

Solanya (dismissed):

totalRP3/UI/Widgets.lua
Meorawr: I'm fine with it as-is - it's not a hot path unless someone is being weirdge and autoclicking the button.

totalRP3/UI/Widgets.lua
```diff
@@ -201,3 +203,82 @@ end
...
+---@return string|table color Contains the last color that was copied.
+function TRP3_API.GetLastCopiedColor()
+	return g_lastCopiedColor;
+end
+
+TRP3_ColorPickerButtonMixin = {};
+
+function TRP3_ColorPickerButtonMixin:OnLoad()
+	TRP3_ColorButtonLoad(self);
+end
```
Meorawr, line 225: Could the body of TRP3_ColorButtonLoad be inlined into this OnLoad function, and TRP3_ColorButtonLoad removed?

Not sure if Extended is using it.

Meorawr (dismissed):

totalRP3/UI/Widgets.lua
Solanya: Doesn't seem like it does, so it should be fine

totalRP3/UI/Widgets.lua
Raenore: It's a fair point, but we're keeping it as this for now (discussed in Discord).
Resolving it for clarity.

Meorawr (approved):

Solanya (approved):

PR #1177, opened 2025-03-11
Fix relative key issue in PetBrowser

Meorawr:
This usage of `relativeKey` is technically invalid. As the key being specified is in fact its parent, resolution of its key won't work until _after_ the parent has been fully parsed.

Conveniently, the result of such behaviour though is that it just anchors to the parent anyway - so it's looked correct, but only by chance.


Raenore (approved):
Technically invalid is the best kind of invalid. I think.. I hope?
This looks fine, really. If it does cause issues in the future, we know who to blame.

PR #1178, opened 2025-03-18
Fix Prat realm removal setting

Solanya:
Realm removal setting wasn't working in either Prat or TRP when using both addons. Dug a little bit, seems it gets added by Prat when giving it `nil` for the additional fields.

If changing the fields to be an empty string instead when the TRP setting is enabled (or player on the same realm), the realm now appears hidden whether the setting is enabled in Prat or in TRP.


Meorawr (approved):

Raenore (approved):
TRP 🤝 Prat

PR #1180, opened 2025-03-21
Thou shalt not disable the update prompt

### LINKED ISSUE #1179: "New update available" pop-up points to non-existent setting to disable it
## Describe the issue
When the local version is outdated, a pop-up eventually (after some synchronisation) appears about the addon version being too old. It says this can be disabled, but the option to disable can not be found:

Image

## Expected behaviour
Do not show the details about how to disable the warning if the ability to disable it had been removed earlier. Or reintroduce the ability to disable it.

## Issue checklist

- ~~The issue occurs with the latest version of the addon.~~
  - This claim is fundamentally impossible to discertain.
- [x] The issue occurs with Total RP 3 as the only enabled addon.

## Steps to reproduce
Give us a list of steps to reproduce the issue with.

1. Have an older version of the add-on installed. I tried with **3.0.5**.
2. Log in while others are also playing on the server.
3. Wait about 5 seconds after opening TRP window for the warning to appear.

## Additional information

It should appear in the latest version as well. As of commit b04c588e43d56c5433db49ef48f57bc8f6735f95, the message and the usage in code is still present:

https://github.com/Total-RP/Total-RP-3/blob/b04c588e43d56c5433db49ef48f57bc8f6735f95/totalRP3/Locales/enUS.lua#L1249

https://github.com/Total-RP/Total-RP-3/blob/b04c588e43d56c5433db49ef48f57bc8f6735f95/totalRP3/Modules/Register/Main/RegisterExchange.lua#L265-L277

Notice how `checkVersion()` or its 2 usage points do not contain any indication of checking whether the user disabled these alerts.

Meorawr:
Fixes #1179.

The setting was disabled and removed in 32425b566b5e1517fea4d2ded4e4fe9c80eb0be8 (about 7 years ago). Removed the loc strings that would have otherwise represented it, and cleaned up the update prompt overall to make it a bit less verbose.

Notably, we no longer format the version number difference into the update message because it's a silly idea to try and describe the difference between a dotted version number pair like "2.7.4" and "3.0.9" with a single numeric value like "13".

image


Meorawr:
Also, I'm tempted to just merge the two strings together and remove the more-than-three-versions check. Consider a case where we discover that we've been incorrectly trimming realm names when sending out requests which would result in profile transfers not working correctly - you'd only see that part of the message if you were horribly out of date to begin with.

Solanya (dismissed):

Raenore (approved):
Looks fine.

Although for consistency,. I would still at least suggest perhaps to use `|cnWARNING_FONT_COLOR` instead of `|cnRED_FONT_COLOR`. Given that is what we've been implementing for the purpose of 'watch out for this' highlighting.

But no skin off my back if you want to stick with RED for now.

Solanya (approved):

PR #1181, opened 2025-03-23
Fix "Open profile" not showing on community menus

Solanya:
For players in the guild or in-game community right-click menus, the "Open profile" option hasn't been showing up recently due to the unit referenced changing. This is adding the new units so the menu option appears again.

I don't have a guild or in-game community but someone else who shall raemain unnamed tested it on her end.


Ghostopheles:
This seems to work as one would expect but while testing it I uncovered something entirely unrelated but altogether far more sinister.

Meorawr (approved):

Raenore (approved):
Tested and seems to work fine in both Guilds and Communities once more.

PR #1182, opened 2025-03-23
Kill the maximize button

Meorawr:
Delivering us from evil and honoring our promises of old, it's time to kill the maximize button given it's had a rather stupid bug for a few months that no one's told us about. We hate the feature, people click it all the damn time accidentally, etc. Yeet.

This will break Extended, but that's Extended's problem. Make your own damn resize and close buttons.


Raenore (dismissed):
This seems fine, leave the loc alone now.

Raenore (approved):
Stop dismissing my reviews.

Tested with latest changes, nothing seems to be out of order (yet).

PR #1186, opened 2025-03-23
Misc. utility function cleanup

Meorawr:
We had some functions that were either unused, used in one location, used only by extended, or were trivially replaceable with default UI equivalents hanging around. Kill them all.


Raenore (approved):
This seems in order, assuming Meow tested this all in his infinite wisdom.

Meorawr:
Imagine thinking I tested this stuff

Solanya (approved):

PR #1190, opened 2025-04-22
Fix missing tutorial templates

Solanya:
HelpPlateBox and its highlight have left the chat without alerting anyone, this feels like a quick hack job and the border isn't the right color but we can look at it again later

### DIFF
--- a/totalRP3/Core/MainStructure.lua
+++ b/totalRP3/Core/MainStructure.lua
@@ -348,8 +348,8 @@ local function showTutorial(tutorialStructure)
 	for frameIndex, frameInfo in pairs(tutorialStructure) do
 		if not BUTTONS[frameIndex] then
 			BUTTONS[frameIndex] = CreateFrame( "Button", nil, TRP3_TutorialFrame, "TRP3_TutorialButton" );
-			BUTTONS[frameIndex].box = CreateFrame( "Frame", nil, TRP3_TutorialFrame, "HelpPlateBox" );
-			BUTTONS[frameIndex].boxHighlight = CreateFrame( "Frame", nil, TRP3_TutorialFrame, "HelpPlateBoxHighlight" );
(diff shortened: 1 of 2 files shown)


Raenore:
I did a bit of checks here and there and noticed why the color is off:
https://github.com/Gethe/wow-ui-source/blob/live/Interface/AddOns/Blizzard_HelpPlate/Blizzard_HelpPlate.lua#L120
```lua
function HelpPlateBoxMixin:OnLoad()
    for index, texture in ipairs(self.Textures) do
        texture:SetVertexColor(1, 0.82, 0);
    end
end
```

I hacked it, temporarily, by putting this in `ShowTutorial`:
```lua
for _, texture in ipairs(box.Textures) do
    texture:SetVertexColor(1, 0.82, 0);
end
```



I'm sure this can put you in the right direction on how you want to tackle this or the like.

Also I wonder if for:
```xml
<Frame name="TRP3_HelpPlateBox" inherits="TRP3_ThinBorderTemplate" mixin="HelpPlateTileMixin" virtual="true" hidden="true">
```

It might not be more appropriate to use:
```xml
<Frame name="TRP3_HelpPlateBox" inherits="TRP3_ThinBorderTemplate" mixin="HelpPlateBoxMixin" virtual="true" hidden="true">
```

So basically `HelpPlateTileMixin` vs `HelpPlateBoxMixin`? Although I don't know how you've hacked it together so perhaps you had a reason to go with one over the other.

Solanya:
I think I was trying to inherit something directly that did not work, not sure how I ended up with that result.
Weirdly, inheriting from HelpPlateBoxMixin does nothing to fix the color issue? Your fix however does work. I have no idea what's happening there.

Raenore (approved):
Tested it once more with latest pull from this branch (and latest Extended for good measure).

No errors reported and the tutorial borders look a mighty fine yellow now.

PR #1192, opened 2025-04-26
Use UnitOwnerGUID for companion owner queries

Meorawr:
This API was added in 11.1.5 and will eventually make its way to Classic versions, allowing us to (eventually) get rid of the C_TooltipInfo and scanning tooltip nonsense here.

### DIFF
--- a/totalRP3/Core/UITools.lua
+++ b/totalRP3/Core/UITools.lua
@@ -579,7 +579,13 @@ local function getCompanionOwner(unitType, targetType)
 	local ownerName;
 	local ownerRealm;
 
-	if C_TooltipInfo then
+	if UnitOwnerGUID then
+		local ownerGUID = UnitOwnerGUID(unitType);
+
+		if ownerGUID ~= nil then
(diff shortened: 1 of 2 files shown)


Solanya (approved):

totalRP3/Core/UITools.lua
```diff
@@ -579,7 +579,13 @@ local function getCompanionOwner(unitType, targetType)
 	local ownerName;
 	local ownerRealm;
 
-	if C_TooltipInfo then
+	if UnitOwnerGUID then
+		local ownerGUID = UnitOwnerGUID(unitType);
+
+		if ownerGUID ~= nil then
+			ownerName, ownerRealm = select(6, GetPlayerInfoByGUID(ownerGUID));
```
Solanya, line 586: Actually, if I wanted to nitpick, that part is present in both ifs, we could get it out of there

Meorawr: We could, but that other branch'll be deleted at some point anyway.

Solanya: Fair

Solanya (approved):

PR #1193, opened 2025-04-28
Fix link conversion on MSP

Solanya:
I received an anonymous hint that icon links were busted on MRP. After investigating both sides, I found out we were sending invalid data over on MRP when converting {link*...} tags into the \[...](...) version MRP is used to.

image

Doing the link tag conversion prior to converting the rest of the tags fixes the issue, as there is no nested brackets anymore.


Raenore (approved):
Checked the regex on it and all seems to be in order.
Plus two separate accounts telling me that it works (one I believe more than the other).

Good job on spotting this anonymous tipster and Sol.

PR #1198, opened 2025-07-07
Fix broken realm name normalization pattern

Meorawr:
Realm name normalization needs to skip period characters, but also our pattern here was broken - it'd only replace _one_ invalid character because the repeat specifier was inside the character set block.


Solanya (approved):

Raenore (approved):
Crazy patterns, it works.

PR #1202, opened 2025-07-18
Restore target frame buttons settings + add settings title for them

Solanya:
During some big modernization pass last summer (#1108), the toolbar and target frame settings registration got moved to be properly registered only when those were enabled, but the target frame one was moved a bit too early and excluded the buttons visibility options.

It is now back in its proper place, and as an added bonus, I split those settings (and the toolbar ones) with a title, didn't like how they were all bunched up with the rest of the settings.

image
image

### DIFF
--- a/totalRP3/Locales/enUS.lua
+++ b/totalRP3/Locales/enUS.lua
@@ -568,7 +568,7 @@ Possible status:
 	CO_TARGETFRAME_USE_2 = "Only when IC",
 	CO_TARGETFRAME_USE_3 = "Never (Disabled)",
 	CO_TARGETFRAME_ICON_SIZE = "Icons size",
-	CO_BARFRAME_BUTTONSVISIBILITY = "Buttons visibility",
+	CO_BARFRAME_BUTTONSVISIBILITY = "Button visibility",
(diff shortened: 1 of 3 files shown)


totalRP3/Locales/enUS.lua
```diff
@@ -568,6 +568,7 @@ Possible status:
 	CO_TARGETFRAME_USE_2 = "Only when IC",
 	CO_TARGETFRAME_USE_3 = "Never (Disabled)",
 	CO_TARGETFRAME_ICON_SIZE = "Icons size",
+	CO_BARFRAME_BUTTONSVISIBILITY = "Buttons visibility",
```
Meorawr, line 571:
```suggestion
	CO_BARFRAME_BUTTONSVISIBILITY = "Button visibility",
```
Drop the suffix.

Meorawr (dismissed):

totalRP3/Locales/enUS.lua
Solanya: the suffix?

Meorawr (approved):

PR #1203, opened 2025-07-25
Somewhat fix companion support in Classic

Meorawr:
Initially, this PR was just to swap to using the more efficient C_PetJournal.GetOwnedPets API to remove a bunch of stupid hacks, however when testing in Classic I discovered that a lot of our assumptions on the pet and mount journals in the various Classic flavors are wrong and broken.

So - I rewrote it all again. The issue is that the C_PetJournal and C_MountJournal namespaces now exist with functions in Classic Era, so a lot of the code paths being taken on those clients assume that there's data behind those APIs - but there's not.

To remedy this, we now have a wrapper around the API in TRP3_CompanionUtil that's forked based upon the client type - specifically, we'll load a variant that uses the collections journals in Mainline, Wrath CN (if we supported it), Cata, and MoP. Classic Era and TBC load a variant that uses our static data.


Solanya (approved):

totalRP3/Core/Vanilla/CompanionUtilOverrides.lua
```diff
@@ -0,0 +1,322 @@
+-- This should be converted to an [AllowLoadGameType ...] directive when supported
+-- on all clients.
+--
+-- While the collections journal wasn't introduced until Mists, it was added
+-- into Cataclysm and then Wrath (CN) as part of Classic.
+
+if LE_EXPANSION_LEVEL_CURRENT >= LE_EXPANSION_WRATH_OF_THE_LICH_KING then
+	return;
+end
```
Meorawr, line 9: At some point I'll get to the TODO in Core.xml to move all our loading to the TOC.

When I do that, I'll remove this logic and make the client skip the file via `[AllowLoadGameType standard, mists, cata, wrath]` entirely.

totalRP3/Core/Vanilla/CompanionUtilOverrides.lua
```diff
@@ -0,0 +1,322 @@
...
+
+-- On clients that don't have a pet or mount journal, we statically embed
+-- a database with just enough data to allow us to build our own fake version
+-- of the journals in-game.
+--
+-- This static data doesn't include localized strings, and needs to be
+-- filtered based on what's actually available in the client. To do this we
+-- need to do some sparse loads for item and spell data prior to registering
+-- the companion.
```
Meorawr, line 160: One thing not documented here - previously we filtered out companions and mounts that didn't exis in a client via C_Item.DoesItemExistByID and C_Spell.DoesSpellExist.

The former of these functions has been "broken" for a while and returned false only if the spell ID were wholly invalid.

Now, we don't use those functions - instead, we trigger a spell/item load request and only register the companion in our static data tables if we get a successful result.

totalRP3/Core/Vanilla/CompanionUtilOverrides.lua
```diff
@@ -0,0 +1,322 @@
...
+-- last successful cast will represent the current battle pet.
+--
+-- Note that we can't tell when a companion pet is dismissed, so our query
+-- data will always contain the data for the last-successful cast even if
+-- that cast dismissed the pet. Realistically this should be fine since
+-- if it's dismissed other players can't see the unit to request the data
+-- anyway.
+--
+-- For persistence across UI reloads we store the summoned pet data in our
+-- saved variables, and reset it upon initial login of a new character.
```
Meorawr, line 279: This logic was relocated from RegisterCompanionsMain, where it was horribly broken - we were never passing the cast GUIDs or isInitialLogin parameters through to the callbacks.

totalRP3/Core/CompanionUtil.lua
```diff
@@ -0,0 +1,133 @@
...
+	end
+
+	-- Is the last mount that we tested still the current summon?
+	if cachedSummonedMountID ~= nil then
+		local _, _, _, isActive = C_MountJournal.GetMountInfoByID(cachedSummonedMountID);
+
+		if not isActive then
+			cachedSummonedMountID = nil;
+		end
+	end
```
Meorawr, line 102: This last-mount optimization was present in Classic handling of companion code previously, but not Mainline.

So - if you were mounted in Mainline, receiving requests for profile data could trigger a full scan of your mount journal. Multiple times.

Now we remember what mount ID you were last active on and optimistically re-query that, saving the cost of the full scan.

totalRP3/Core/Globals.lua
```diff
@@ -72,7 +72,7 @@ TRP3_API.globals = {
 TRP3_ClientFeatures = {
 	ChannelBroadcasts = (WOW_PROJECT_ID == WOW_PROJECT_MAINLINE),
 	WarMode = (LE_EXPANSION_LEVEL_CURRENT >= LE_EXPANSION_BATTLE_FOR_AZEROTH),
-	Transmogrification = (LE_EXPANSION_LEVEL_CURRENT >= LE_EXPANSION_CATACLYSM),
+	Transmogrification = (LE_EXPANSION_LEVEL_CURRENT >= LE_EXPANSION_WRATH_OF_THE_LICH_KING),
```
Meorawr, line 75: Transmog was backported to Wrath CN, so updating this as an extra.

Solanya (approved):

PR #1204, opened 2025-07-27
Add new profile import/export format

Meorawr:
This adds a new profile export format based around the native C_EncodingUtil APIs, aimed at solving a few issues with our exports:

- Profile exports aren't currently compressed, resulting in more complaints than I care to remember about the 20 KB size limit.
- Profile exports are subject to the crappy warning about Word or Google Docs mangling data irrecoverably.

Before going on with the solution, here's how the new export looks:

```
-----BEGIN TRP3 PROFILE-----
Name: Argent Dawn - Cool Profile
Exported: 2025-07-27 02:45:57
AddOn-Version: -dev

PY5PT4NAFMSln8J44uLRpGKbnnn8KcRKN7CkvZHnupVNgCUPxHgj8eyXEL34Kd1q0+vM/Gbm/fJjO1
85q7mzWC6WIi/z+PO+JX1QlUywlvsrl55l09s+vjb2je1pXdns3w/bCt8kfW1FiYSil6S6XonuB1I3
Ytg8IckGNsn6lIeHeASWje4wg9jbX6MolRxkbeqLc0XRnsDiIGs0kLdZ75BIaYKQe0Aa0uwCwsQ/Xo
nP3AQpm7mDBbvcCvBRv/TfwJ0R+N0ELPrbhcjs88ACfms+WH6tOjFBxo0ZHIVf
-----END TRP3 PROFILE-----
```

The new format is basically same as the old `{ addonVersion, profileID, profileData }` table internally, but serialized via CBOR, compressed, and then wrapped into a base64 encoded PEM envelope with lines wrapped at 78 characters each.

The wrapping is to avoid some of the performance pitfalls that editboxes suffer from when dealing with large strings - basically, word wrapping sucks and hard wrapping solves the problem.

Because the profile data is now quite heavily encoded we include some headers before the data to give the user some indication on what the profile actually contains.


totalRP3/Core/Profiles.lua
```diff
@@ -627,7 +624,7 @@ function TRP3_API.profile.init()
 				showConfirmPopup(loc.PR_PROFILEMANAGER_IMPORT_WARNING:format(Utils.str.color("g") .. profiles[profileID].profileName .. "|r"), import);
 			end
 		else
-			Utils.message.displayMessage(loc.DB_IMPORT_ERROR1, 2);
+			Utils.message.displayMessage(string.format(loc.PR_IMPORT_ERROR, errorOrProfileID), 2);
```
Meorawr, line 627: As a fun fact - the DB_IMPORT_ERROR1 loc string doesn't exist in core. It's an extended only thing.

totalRP3/Core/EncodingUtil.lua
```diff
@@ -0,0 +1,177 @@
...
+
+TRP3_EncodingUtil = {};
+
+--
+-- PEM Encoding and Decoding Utilities
+--
+
+function TRP3_EncodingUtil.IsPEMEncodingSupported()
+	return C_EncodingUtil and C_EncodingUtil.EncodeBase64;
+end
```
Meorawr, line 12: Temp function that'll be cleaned up when Classic Era gets its shit together. 

Currently this is used to branch into using the old export format, rather than trying to shoehorn together Lua-based implementations of all these things.

totalRP3/Core/EncodingUtil.lua
```diff
@@ -0,0 +1,177 @@
...
+		local label, data, headers = decoder:Decode();
+
+		if label then
+			return index + 1, label, data, headers;
+		end
+	end
+
+	local index = 0;
+	return DecodeNext, self, index;
+end
```
Meorawr, line 166: This DecodeAll interface isn't used currently, but it exists to support a hypothetical future world of exporting/importing multiple profiles at once by just allowing the individual PEM blocks to be concatenated.

Raenore (approved):
Tested and confirmed working, too bad editboxes are super wonky.

Solanya (approved):

PR #1205, opened 2025-07-27
Hide toolbar and target frame when in pet battles

Meorawr:
When entering pet battles we have this annoying tendency to leave stuff like the toolbar and target frame open. Hook up a small integration into the state driver system that'll allow us to monitor changes to pet battle state, and update visibility of the frames accordingly.


Raenore (approved):
Tested by starting a pet battle, seems to correctly hide and restore everything with no errors.

Solanya (approved):

PR #1206, opened 2025-08-01
Add request indicator to main profile view

Meorawr:
This implements a small overlay on our register profile view at the bottom right corner if you're looking at a profile that we're actively in the process of requesting new data from. This aims to make it a bit clearer that the addon is actually doing work that might change what you're reading, and to avoid the awkward state where we sometimes just hang with an "Unknown" character name if comms are being slow.

https://github.com/user-attachments/assets/de3a6377-bb7f-4adf-8f9d-52b0be8fb3be

The indicator shows generally if you're on a profile page irrespective of what tab you've got selected, and it'll show if there's _any_ outstanding data for the profile. We could technically make it only show on tabs actively being requested, but that might be a bit strange to look at - this is more stable a display at least, and matches how we show indicators on tooltips so has the benefit of being consistent.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -427,6 +427,12 @@ stds.wow = {
 			},
 		},
 
+		ResizeLayoutMixin = {
+			fields = {
+				"OnShow",
+			},
+		},
+
 		ScrollBoxConstants = {
 			fields = {
 				"DiscardScrollPosition",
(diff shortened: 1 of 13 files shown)


Raenore (approved):
Tested this out with a few profiles on live, all seems in order and noticed no errors.

Solanya (approved):

PR #1207, opened 2025-08-02
EditBox reference fixes

Meorawr:
these gon' go boom on patch day, y'hear


Raenore (approved):
The final bomb the Goblins from Undermine planted before we leave their zone.. EditBoxes!

No errors or issues after this fix, good job.

PR #1208, opened 2025-08-03
Addon communication optimizations

Meorawr:
This sets us up for a cataclysmic communications protocol break of our own accord, presently scheduled for patch 12.0.0, 5.5.2, and 1.15.9.

With this break, we'll fix two inefficiencies with unlogged communications for both profile transfers via Battle.net and general Extended object transfers:

- Serialization of data will go through the native CBOR serialization APIs.
- Encoding of compressed data will use addon-channel safe encoding, rather than text-channel safe encoding.

These two changes should substantially reduce the amount of bandwidth we're using - particularly, the text-channel encoding has a near 50% (!) overhead because over half the byte space is being escaped.

Additionally, for logged transfers we now use a new serialization format that avoids the use of loadstring entirely.


totalRP3/Core/Compression.lua
```diff
@@ -10,36 +9,36 @@ local LibDeflate = LibStub:GetLibrary("LibDeflate");
...
-	Ellyb.Assertions.isType(data, "string", "data");
+-- Protocol break for 12.0: Historically, data compression intended for use
+-- in addon messages has been using a codec for regular chat messages. This
+-- incurs a ~53% overhead in transmission size as over half of the range of
+-- bytes needs to be escaped.
+local useAddonChatEncoding = (select(4, GetBuildInfo()) >= 120000);
 
-	local compressedData = LibDeflate:CompressDeflate(data);
+function Compression.compress(data, willBeSentViaAddOnChannel)
+	local compressedData = TRP3_EncodingUtil.CompressString(data);
```
Meorawr, line 19: I got rid of the assertions in these functions as the native API and LibDeflate already validate types anyway - so we're just duplicating the check needlessly.

totalRP3/Core/Compression.lua
```diff
@@ -10,36 +9,36 @@ local LibDeflate = LibStub:GetLibrary("LibDeflate");
...
+			compressedData = assert(LibDeflate:DecodeForWoWChatChannel(compressedData));
 		end
 	end
 
-	local decompressedData, _ = LibDeflate:DecompressDeflate(compressedData);
-	if decompressedData == nil then
-		error(TRP3_API.Colors.Red("[AddOn_TotalRP3.Compression.decompress ERROR]:") .. "\nCould not decompress data \"" .. TRP3_API.Colors.Grey(tostring(compressedData)) .. "\"");
-	end
-
+	local decompressedData = TRP3_EncodingUtil.DecompressString(compressedData);
```
Meorawr, line 41: There's some misc. cleanup here:

- Got rid of the TODO; returning the passed data would probably have just blown up callers, so let's make it an actual assert.
- The native compress and decompress APIs error if something is wrong with the data, so there's no need to do nil checks. Any usage of LibDeflate Decode/Decompress APIs has been wrapped in an assert to mirror this behaviour.

totalRP3/Core/EncodingUtil.lua
```diff
@@ -1,8 +1,41 @@
...
+local ADDON_ENCODE_TABLE = { ["\000"] = "\001\002", ["\001"] = "\001\003" };
+local ADDON_DECODE_TABLE = { ["\001\002"] = "\000", ["\001\003"] = "\001" };
+
+function TRP3_EncodingUtil.EncodeAddOnMessage(data)
+	return (string.gsub(data, "[%z\001]", ADDON_ENCODE_TABLE));
+end
+
+function TRP3_EncodingUtil.DecodeAddOnMessage(data)
+	return (string.gsub(data, "\001[\002\003]", ADDON_DECODE_TABLE));
+end
```
Meorawr, line 37: I'd like to drop LibDeflate as a dependency one day entirely and not have to keep it around just for its addon channel encoding.

totalRP3/Modules/Register/Main/RegisterExchange.lua
```diff
@@ -479,15 +479,9 @@ local function incomingInformationTypeSent(structure, senderID, channel)
 	QueryCooldowns[informationType][senderID] = nil;
 	TRP3_Addon:TriggerEvent("REGISTER_DATA_RECEIVED", senderID, informationType);
 
-	local decodedData = data;
-	-- If the data is a string, we assume that it was compressed.
-	if type(data) == "string" then
-		decodedData = Utils.serial.safeDecompressCodedStructure(decodedData, {});
-	end
```
Meorawr, line 486: I'm 99.9% certain this code path can never be hit - the four information types that this is set up for always serialize data as a table, never as a string, and there's no calls to the utility that compresses data via LibCompress anyway.

I assume this was to support ancient versions of the addon, but it's definitely no longer needed at this point.

totalRP3/Core/Utils.lua
```diff
@@ -735,18 +733,6 @@ local function safeDeserialize(structure, default)
...
-	return deserialize(libCompress:Decompress(libCompressEncoder:Decode(message)));
-end
-
-Utils.serial.safeDecompressCodedStructure = function(message)
-	return safeDeserialize(libCompress:Decompress(libCompressEncoder:Decode(message)));
-end
-
-Utils.serial.encodeCompressStructure = function(structure)
-	return libCompressEncoder:Encode(libCompress:Compress(serialize(structure)));
-end
```
Meorawr, line 748: Two of these were unused, and one only had the single usage in RegisterExchange.lua - so killing all three.

Aside - I want to kill the other Utils.serial functions, but they're currently being used in Extended until that migrates to using the new export/import format.

Solanya (approved):

PR #1212, opened 2025-08-05
The great default profile purge

Solanya:
There has been some weird profile ID manipulation noticed last month which messed with how profiles are shown in the directory. As default profiles are not meant to be editable, we now enforce deletion of any invalid fields for a default profile.

I'm also regenerating the default profile ID every login session in order to clean up lingering identifying data, and clearing up default profiles in the auto-purge (since those aren't displayed in the directory anyway).

Also I'm sanitizing the content of the URL tag. I've had that change waiting for a while and I forgot what exact use prompted it. Might have been some multiline shenanigans.

### DIFF
--- a/totalRP3/Core/Utils.lua
+++ b/totalRP3/Core/Utils.lua
@@ -528,6 +528,8 @@ local function GenerateLinkFormatter(line, defaultLinkColor, includeBraces, isMa
 			text = "[" .. text .. "]";
 		end
 
+		url = Utils.str.sanitize(url, false);
+
 		if linkType then
 			url = TRP3_LinkUtil.CreateLinkString(linkType, url);
 		end
--- a/totalRP3/Modules/Register/Characters/RegisterMain.lua
+++ b/totalRP3/Modules/Register/Characters/RegisterMain.lua
@@ -682,7 +682,7 @@ local function cleanupProfiles()
 	TRP3_API.Log("Protected profiles: " .. CountTable(protectedProfileIDs));
 	local profilesToPurge = {};
 	for profileID, profile in pairs(profiles) do
-		if not protectedProfileIDs[profileID] and (not profile.time or time() - profile.time > getConfigValue("register_auto_purge_mode")) then
+		if not protectedProfileIDs[profileID] and (TRP3_API.profile.isDefaultProfile(profileID) or not profile.time or time() - profile.time > getConfigValue("register_auto_purge_mode")) then
 			tinsert(profilesToPurge, profileID);
 		end
 	end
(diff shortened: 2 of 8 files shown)


Raenore (approved):
Tested this a few ways, and the new default profile ID is generated on every login and reload, which is very neat.

I also tried to manipulate things, but since we remove the previous profile ID on login, it can never be changed.

I also reverted parts of the code and the code entirely to see if I could still manipulate it by other means. However, as long as the receivers have the latest TRP, the default profiles will be cleaned on the receiver's end, even if the user finds a way to make them locally different (which is the best they can still get for them).

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -681,7 +682,7 @@ local function cleanupProfiles()
 	TRP3_API.Log("Protected profiles: " .. CountTable(protectedProfileIDs));
 	local profilesToPurge = {};
 	for profileID, profile in pairs(profiles) do
-		if not protectedProfileIDs[profileID] and (not profile.time or time() - profile.time > getConfigValue("register_auto_purge_mode")) then
+		if not protectedProfileIDs[profileID] and (TRP3_API.profile.isDefaultProfile(profileID) or not profile.time or time() - profile.time > getConfigValue("register_auto_purge_mode")) then
```
Meorawr, line 685: Do we have a protection against default profiles ending up in the protected profile table?

totalRP3/Core/Utils.lua
```diff
@@ -528,6 +528,8 @@ local function GenerateLinkFormatter(line, defaultLinkColor, includeBraces, isMa
 			text = "[" .. text .. "]";
 		end
 
+		url = Utils.str.sanitize(url, false);
```
Meorawr, line 531: I'd like to see the bool assigned to a variable with an actual name, because I can't remember off the top of my head what it's doing.

Something about multiline right? I'm thinking of people using links for hover tooltips in their about page.

Solanya: It's multiline, yeah.

totalRP3/Modules/Register/Characters/RegisterMain.lua
Solanya: We don't. I probably should move the default profile check so it ignores protected profiles.
However thinking about it, relations and notes are now invalid to apply to characters with default profile. I don't want to delay the change because I think the current behaviour is unhealthy, but we might have to re-think how to handle those in the future. (Notes implementation in particular is starting to feel very cursed to me.)

PR #1213, opened 2025-08-05
Flyway Log counting is hard

Raenore:
Saving Total RP 3 once again (I wish).
If from build is +1, it will always be the same as the following build log-wise, so that's silly.

Before:
```lua
TRP3_Flyway = {
  ["currentBuild"] = 22,
  ["log"] = "Patch applied from 22 to 22 on 15/08/24 05:41:37",
}
```

After
```lua
TRP3_Flyway = {
  ["currentBuild"] = 22,
  ["log"] = "Patch applied from 21 to 22 on 05/08/25 05:40:56",
}
```

PS: Extended also needs this fix, given it stol-- uses the same Flyway idea.

### DIFF
--- a/totalRP3/Modules/Flyway/Flyway.lua
+++ b/totalRP3/Modules/Flyway/Flyway.lua
@@ -10,18 +10,18 @@ if not TRP3_Flyway then
 end
 
 local function applyPatches(fromBuild, toBuild)
-	for i = fromBuild + 1, toBuild do
+	for i = fromBuild, toBuild do
 		if type(TRP3_API.flyway.patches[tostring(i)]) == "function" then
 			TRP3_API.Logf("Applying patch %s", i);
 			TRP3_API.flyway.patches[tostring(i)]();
 		end
 	end
-	TRP3_Flyway.log = ("Patch applied from %s to %s on %s"):format(fromBuild, toBuild, date("%d/%m/%y %H:%M:%S"));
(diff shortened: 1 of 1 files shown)


Meorawr (changes requested):
I'm 90% certain this isn't going to just change the logging - we'll end up running old migrations twice, and on new installs we'll error since there's no version zero migration.

totalRP3/Modules/Flyway/Flyway.lua
```diff
@@ -10,7 +10,7 @@ if not TRP3_Flyway then
 end
 
 local function applyPatches(fromBuild, toBuild)
-	for i = fromBuild, toBuild do
+	for i = fromBuild + 1, toBuild do
```
Meorawr, line 13: But then it's not really `fromBuild` now is it, since we've changed this function to be exclusive of its minimum bound which is quite the opposite of what I'd expect.

Why aren't we just leaving the original lines as they are and simply `-1`ing the fromBuild in the actual formatting call a few lines below?

Meorawr (changes requested):

Meorawr (approved):

totalRP3/Modules/Flyway/Flyway.lua
Raenore: Yes dad, should be resolved now.

PR #1214, opened 2025-08-07
Protect profiles from auto-purge again

Solanya:
I moved the default profile before checking protected profiles but moved the parentheses out of it as well which is pretty bad. x_x


Raenore (approved):
Parenthesis logic looks good here.
Only unprotected profiles should be hit now, while protected don't get purged.

PR #1216, opened 2025-08-25
Fix Enter/Escape calls on input popups

Solanya:
Text input popup was throwing an assertsafe error when trying to press Enter/Escape while editing the editbox.
Number input popup was straight up not working (because still using .button1 and .button2).

Instead of using the "OnClick" script, I just call the StaticPopup_OnClick which doesn't go through the assertsafe.
*However*, I could just copy the OnAccept/OnCancel functions (with minot adjustments as self is EditBox). Depends what's preferred.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -606,6 +606,7 @@ stds.wow = {
 		"ShowingCloak",
 		"ShowingHelm",
 		"ShowUIPanel",
+		"StaticPopup_OnClick",
 		"StaticPopup_Show",
 		"StopMusic",
 		"StopSound",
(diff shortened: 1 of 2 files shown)


Meorawr:
What on god's green earth is this code doing and can it just be `GetDialogButton1(self:GetParent()):Click()`?

Meorawr:
Maybe "what" is the wrong word, more "why the fuck is it falling over itself when the task is to just click a button".

But now I actually look at it some more, keep the StaticPopup_OnClick approach.

Meorawr (approved):

PR #1217, opened 2025-08-30
Add dice roll throttle + player hyperlink

Solanya:
As there is spam potential on this command, adding two improvements to it:
- Player hyperlink on the entire chat line, so it's easy to "Right-Click > Ignore" the source of the spam.
- A short cooldown (2 seconds) to limit the impact of spam in the first place, which should not impact standard use of it. The cooldown has not been added to PARTY/RAID broadcasts, as those should be more controlled environments not affected by potential spam.


Meorawr (approved):
Probably good enough for now.

Raenore (approved):
This looks good.

I agree, with Meow, that future-wise we can target some of this for more improvements.
But there being a throttle on non-controlled broadcasts is a good thing.

PR #1218, opened 2025-09-30
Add IsVirtual method to custom data providers

Solanya:
The custom data providers written for the icon/background browsers were missing this new(?) method for 11.2.5 so I added it. CreateDataProvider sets it at false by default, I don't really understand what it's meant for and I'm afraid to ask but it works like this so I trust it's good

### DIFF
--- a/totalRP3/UI/Browsers/BackgroundBrowser.lua
+++ b/totalRP3/UI/Browsers/BackgroundBrowser.lua
@@ -292,7 +292,7 @@ local function CreateImageDataProvider(model)
 	end
 
 	function provider:IsVirtual()
-		return false;
+		return true;
 	end
 
 	provider:GenerateCallbackEvents({ "OnSizeChanged" });


Meorawr (approved):
I've not checked what the purpose of IsVirtual is but as long as it works I'm fine with defaulting it to false.

totalRP3/UI/Browsers/BackgroundBrowser.lua
```diff
@@ -291,6 +291,10 @@ local function CreateImageDataProvider(model)
 		return model:GetImageCount();
 	end
 
+	function provider:IsVirtual()
+		return false;
```
Meorawr, line 295:
```suggestion
		return true;
```

Meorawr (changes requested):

Meorawr (approved):

PR #1220, opened 2025-10-06
Refresh directory with new cols and QoL

### LINKED ISSUE #1029: Review directory info columns
It's been brought to attention that the "Profile type" column doesn't really serve much purpose, especially when one type overwhelms any other. While that column could benefit from getting removed, new columns could be added to take its place.

While not an exhaustive list, potential categories suggested were:
- Guild
- Realm
- Faction

Raenore:
This resolves #1029.

This enhances the directory in a few ways:
- The entire directory line is now clickable, instead of just when you hover over the Name part.
- All sorting on first click is ASC (alphabetical), instead of DESC (inverse).
- Profile type was removed as it didn't serve much/any purpose for players.
- Relations are sorted in ASC by highest (Family) to lowest (Unfriendly), in DESC it is lowest (Unfriendly) to highest (Family). This includes custom relations. None is always at the end.
- Guild(s) is added. It displays the first OOC guild found attached to the characters with that profile. This can be sorted ASC and DESC. If multiple guilds exist for a profile, they are listed on hover.
- Realm(s) is added. Displays the first realm found attached to the characters with that profile. This can be sorted ASC and DESC. If multiple realms exist for a profile, they are listed on hover.
- Updated some vague/older variable names, widgets, buttons.

A small video showcasing these changes, relations were put for testing and this is not an RP profile.

https://github.com/user-attachments/assets/af99f1e0-d552-4af4-bcd7-e67207785ecc


whisperity:
Considering that there is a way to customise the "guild" value shown on the profile and in the tooltip (and, if using a compatible nameplate addon, the nameplate even!), I would think showing the "profile"'s guild as opposed to the "OOC guild" (similarly how the IC name is shown, not the OOC name) would be more appropriate.

And then the tooltip could show the OOC guild(s) for the OOC char(s), and the search field could equally respect both, just like how the "name" search does already.

PR #1232, opened 2025-12-05
Housing scan adjustments

Solanya:
- Map scan in neighborhoods now only displays people in the current neighborhood
- Players inside houses will respond with their plot coordinates (fuzzed slightly so players don't all stack in a single point if there are multiple inside a house)
- Players will not respond to the map scan if the house settings are set to anything but "Anyone" (the API function is a bit delayed to update after a change, sometimes up to a couple minutes? but we don't really want to force request the house info every time there is a change...)

As a side change, turned the personal phased zones into an enum, just to be consistent with the added neighborhoods enum


Raenore (dismissed):
Tested and seems to work fine.

I personally think this is a perfect compromise to respect people's set up housing permissions before displaying inside house pins.

Meorawr (dismissed):

Ghostopheles (approved):

PR #1237, opened 2025-12-30
Prefer native C_ColorUtil functions

Meorawr:
Some absolute chad exposed native APIs for HSV <=> RGB and HSV <=> HSL conversions; update existing code to use those instead. Also resolve an issue where for some reason the relationship presets weren't using the proper color constants.

### DIFF
--- a/totalRP3/Core/Color.lua
+++ b/totalRP3/Core/Color.lua
@@ -1,12 +1,7 @@
 -- Copyright The Total RP 3 Authors
 -- SPDX-License-Identifier: Apache-2.0
 
-local TRP3_API = select(2, ...);
-
-local ConvertHSLToHSV;
 local ConvertHSLToRGB;
-local ConvertHSVToHSL;
-local ConvertHSVToRGB;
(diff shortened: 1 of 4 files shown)


totalRP3/Core/Color.lua
```diff
@@ -418,33 +377,17 @@ function ConvertHWBToRGB(h, w, b)
...
-			h = (b - r) / c + 2;
-		else
-			h = (r - g) / c + 4;
-		end
+function ConvertRGBToHSV(r, g, b)
+	local h, s, v = C_ColorUtil.ConvertRGBToHSV(r, g, b);
 
-		h = h / 6;
+	if h == -1 then
+		h = 0;  -- Achromatic input; clamp to zero hue.
```
Meorawr: RGB to HSV conversion has a documented note that it'll return a -1 value for hue for pure white or black inputs. Our old code here I believe clamped such cases to always use a hue of zero, so retain this helper function that applies that conversion.

Raenore (approved):

PR #1243, opened 2026-01-18
Fix target frame character test

Solanya:
When deleting a profile tied to a character and clicking on them again, if you cannot get their profile, you're stuck with a target frame with 2 buttons, ignore and report to Blizzard (last one erroring since there is no profile to report).

I fixed the test which should check that the target actually has a profile rather than is just known.


Ghostopheles (commented):
I was gonna approve this immediately to be funny but I can't test it, so I'll settle with a comment containing a single, very helpful acronym.

LGTM 👍👍

PR #1249, opened 2026-01-21
Fix error when targeting player without profile

Meorawr:
Unsure if this is the correct fix overall for what we were intending to accomplish with this change, but for stupid reasons hasProfile has the assert that fires unless you manually check if we have a character entry first - so any hasProfile needs a stupid, pointless isUnitIDKnown because wisdom once decided that simply returning false was too obvious.


Raenore (approved):
Our dear testers (Ghost & Keyboard) confirmed this fixed through the PR.

I trust them, and if it's not fixed we'll blame them instead.

Ghostopheles (approved):

PR #1250, opened 2026-01-21
Fix tooltip health formatting

Meorawr:
Unit health is pretty much permanently secret in Midnight and so our fancy tooltip formatting (and direct comparison against `0`) can't work. Disabling this for now until we can think of a better solution if we want to continue supporting this feature.

If we do want to support this going forward, check C_StringUtil for functions that truncate strings if they're a numeric zero - and probably use AbbreviateNumbers to handle the formatting and division. Probably no need to kill the feature outright.


Solanya:
Tooltip health formatting has been fixed but I didn't come up with a way to hide it while full health just yet. I figure it's a decent compromise for now.

PR #1253, opened 2026-01-21
Disable Kui nameplates module

Solanya:
As Kui currently sits in the shadow realm but we check that it's enabled to disable the Blizz customization, the easiest solution at the moment is to kill the module.

We also disabled customization when using TidyPlates, and that addon also seems to be closing shop in Midnight, so turning that off too.


Meorawr (approved):
straight up fire

PR #1264, opened 2026-01-30
Delete missing phrases on all locales when Exporting All

Solanya:
We sometimes reset loc keys from other languages when the original key has been modified, but the script to export all locales does not handle deleting missing keys other than the English ones, leading to Curseforge trying to put the keys back. Feels like we should apply this on all languages, unless I'm missing something and this would delete the English ones too?


Meorawr:
> Feels like we should apply this on all languages, unless I'm missing something and this would delete the English ones too?

I was probably trying to avoid this problem out of caution. Worth a shot though - worst case, we have all the strings.

Meorawr (approved):

PR #1270, opened 2026-02-04
ElvUI tooltip skin update

kodewdle:
we currently not using NineSlice because of restrictions.
image below to fix relation border + it removes hooks which makes the code more optimal.


Raenore (approved):
Image

This fixes the tooltip border colors (relationships) in Midnight for ElvUI.

PR #1271, opened 2026-02-04
ElvUI target and toolbar skin update

kodewdle:
handle skinning of Target and Toolbar frame better.
- remove some useless code and the hook is not required here so it makes the execution once
  - unless a user changes the color settings in ElvUI that is handled by ElvUI code so no hook needed.
- `customBackdropAlpha` is only set for non-Transparent templates.
  - [when `template` is `nil`, which is like `Default` in ElvUI]: it will use whatever the user has for Tooltip Opacity in ElvUI settings.
- points are adjusted a bit for: Header, Title, and GlanceBar to fit a little more similar to Default style without ElvUI
  - this changes their points and expects the parent to be the anchor.

### DIFF
--- a/totalRP3/Modules/Register/Filter/RegisterMatureFilterDictionary.lua
+++ b/totalRP3/Modules/Register/Filter/RegisterMatureFilterDictionary.lua
@@ -1020,7 +1020,7 @@ local currentLocale = DEFAULT_LOCALE;
 
 function TRP3_API.utils.resources.getMatureFilterDictionary()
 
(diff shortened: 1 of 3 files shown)


Raenore (approved):
Per @kodewdle 
> my changes remove hooks and only execute code when the skin process the very first time
and it use our template correctly rather than handle frames like a tooltip

Everything looks fine, and we can now tweak offsets easier for the toolbar and target frame.

Blizz
Image

ElvUI
Image

Solanya (approved):

PR #1296, opened 2026-02-28
Fix rogue icons in Plater

Solanya:
If there's no displayInfo we should still hide the extra icon frame attached to the nameplate


Ghostopheles (approved):
I think you may be onto something with these changes here.

PR #1297, opened 2026-02-28
Set the target type accordingly when creating a mount profile

### LINKED ISSUE #1201: [Companion profiles] Creation of new profile for a mount from toolbar does not correctly set associated entity type to "mount"
## Describe the issue
If a companion profile for a mount is created using the _🔵 [Summon Warhorse]_ button on the character tool-bar, the created profile will not have the associated entity's "type" correctly set to `Mount`.

The similar behavioural pipeline when using the toolbar for a battle pet with the ⛓️ button does not have exercise this issue.

## Expected behaviour
The associated entity type as shown in the _Companions_ list should be set to `Mount` accordingly.

## Issue checklist

> You can replace the "[ ]" with "[x]" for each option below to tick it.

- [x] The issue occurs with the latest version of the addon. `3.0.12`
- [x] The issue occurs with Total RP 3 as the only enabled addon.

## Steps to reproduce
Give us a list of steps to reproduce the issue with.

1. Mount up on a mount that does not have an associated companion profile. (For Dracthyr characters, _:dragon_face: [Soar]_ is also an acceptable alternative here!)
2. Click the _🔵 [Summon Warhorse]_ ability and create a profile.
3. Navigate to the companion list and hover the little ℹ️  icon for the mount's profile
4. Observe that the type is left empty.

5. Click on the _🔵 [Summon Warhorse]_, unlink the profile, and then link profile just created. Now the value is filled.

## Additional notes
I was not able to find any exact implications for this bug. It does not seem to cause any direct issues whatsoever, apart from this visual quirk. The _Directory -> Companions_ list has a `Companion's type` filter, but no matter what I put into this box from a collection of almost 500 companions of other players, I can't seem to get the filter to match anything. I would assume the reason is because this field is not accurately filled by anyone, or this "type" field is not communicated over the network anyway.

Ghostopheles:
Previously, mount profiles would be created/linked from the target frame with no profile/link type. This corrects the injustice and fixes #1201.

### DIFF
--- a/totalRP3/Modules/Register/Companions/RegisterCompanionsProfiles.lua
+++ b/totalRP3/Modules/Register/Companions/RegisterCompanionsProfiles.lua
@@ -386,6 +386,7 @@ local function companionProfileSelectionList(characterID, targetType, buttonClic
 		ownerID = characterID;
 		if ownerID == Globals.player_id then
 			companionID = tostring(getCurrentMountSpellID());
+			targetType = TRP3_Enums.UNIT_TYPE.MOUNT;
 		else
 			companionFullID = TRP3_API.companions.register.getUnitMount(characterID, "target");
 		end


Raenore (approved):
Image

Tested and seems fixed, good job.

PR #1298, opened 2026-02-28
Fixing all the nameplate bugs

Meorawr:
Resolving a bunch of issues with nameplates because apparently y'all broke them while I was busy.

- Blizzard nameplates should no longer display icons, guilds, or titles on incorrect nameplates.
- Blizzard nameplates now correctly refresh when toggling the name only mode setting.
- Blizzard nameplates now propagate the Size setting to the title and guild text fontstrings.
- Adjusted what functions we hook to apply customizations for Blizzard nameplates. Notably, it should no longer be necessary to check if the nameplate or its unit frame is ever forbidden.
- Relocated ignore checking to ShouldCustomizeUnitNamePlate - ignored units shouldn't be considered customization targets to begin with.
- We now populate the display info with OOC guild data for units that don't have RP profiles if guild name displays are enabled.
- We now populate the display info with class color data for units that don't have RP profiles if the "Use class color by default" setting is ticked.

### DIFF
--- a/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
+++ b/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
@@ -355,16 +355,7 @@ function TRP3_BlizzardNamePlates:UpdateNamePlateName(nameplate)
 		local hasHealthBarOverlap = NamePlateSetupOptions.unitNameInsideHealthBar and not TRP3_NamePlatesUtil.IsNameOnlyModeEnabled() or not displayInfo.isPlayerUnit;
 
 		if displayInfo.shouldColorName and not hasHealthBarOverlap then
-			local color = displayInfo.color;
-			if UnitIsUnit(unitToken, "mouseover") then
-				local r = Saturate(color.r * 1.25);
-				local g = Saturate(color.g * 1.25);
-				local b = Saturate(color.b * 1.25);
-
-				overrideColor = TRP3_API.CreateColor(r, g, b);
-			else
-				overrideColor = TRP3_API.CreateColor(color:GetRGB());
-			end
+			overrideColor = displayInfo.color;
 		end
 	end
 


totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
```diff
@@ -27,16 +27,12 @@ end
 
 local function UpdateFontStringWidgetColor(widget)
 	local desiredColor = widget.TRP3_overrideColor or widget.TRP3_originalColor;
-	if desiredColor then -- To-Do: Temporary Midnight Alpha fix for errors on "Show Offscreen Nameplates", missing color data.
-		CallUnhookedMethod(widget, "SetVertexColor", desiredColor:GetRGBA());
-	end
```
Meorawr: I don't think this check is needed anymore - this was probably broken to begin with because we weren't accomodating unit frame pooling.

totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
```diff
@@ -128,38 +124,113 @@ end
...
 		return;
 	end
 
-	local nameplate = unitframe:GetParent();
-	local frameName = nameplate:GetName();
+	-- luacheck: no redefined (shadowing 'nameplate')
+	hooksecurefunc(nameplate, "SetUnit", function(nameplate, unitToken) self:OnNamePlateUnitAssigned(nameplate, unitToken); end);
+	hooksecurefunc(nameplate, "ClearUnit", function(nameplate) self:OnNamePlateUnitCleared(nameplate); end);
+	self:SetNamePlateInitialized(nameplate);
+end
```
Meorawr: Nameplate initialization has to be split into two phases now.

- Nameplate creation is used to set up hooks for when the nameplate is being set up/reset. The SetUnit hook is invoked after a unit frame has been acquired from its internal pool, and ClearUnit right before it's released.
- When the unit is assigned, we then see if we need to run (90% of) our unit frame initialization logic.

totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
```diff
@@ -215,7 +285,7 @@ function TRP3_BlizzardNamePlates:OnUnitFrameSetUp(unitframe)
 		titleWidget:SetVertexColor(TRP3_API.Colors.Grey:GetRGBA());
 		titleWidget:Hide();
 
-		nameplate.TRP3_Title = titleWidget;
+		unitframe.TRP3_Title = titleWidget;
```
Meorawr: Easier to store the custom widgets on the unit frame now that these are pooled.

PR #1300, opened 2026-03-01
Increase default tertiary tooltip font size to 12

Meorawr:
Our default font size for OOC/Currently text in tooltips is a terribly small default. For accessibility, let's increase this to 12 which matches the default size of all tooltips in the game.

We _couuld_ for consistency also reduce the main tooltip font size from 16 to 14 (which would match default tooltip headers) but that's less important than just having readable CU/CO by default.


Raenore (approved):
Frankly the first thing I ever did was put these on 12.
10 is just a horrible small default nowadays, so I endorse this change.

PR #1310, opened 2026-03-17
Block TRP query responses when another MSP addon is enabled

Solanya:
While TRP currently ducks away from using MSP when another MSP addon is enabled, TRP comms are currently still working, leading to weird behaviour where profiles flicker between the TRP version and the MSP version for other TRP users.

In order to fix this, all profile query responses are now disabled when another MSP addon is enabled. This should still allow to receive profiles on TRP (for companion profiles as they're not available on other addons currently) as well as use and respond to broadcast features.

I've also adjusted the wording on the popup, and added a new hoverable icon at the top of the player profile tabs, to make it clearer that having multiple addons is an actual issue.

image
image


Raenore (approved):
On first look, this seems in order.
Less conflicts is always better.

Meorawr (approved):

PR #1313, opened 2026-03-20
Fix error when /trp3 open in instances

Solanya:
This error first came from TRP3 Unit Frames, but technically could still happen with us alone. If `/trp3 open` is called with a unit token (or nothing at all, which uses `"target"` as default), we cannot retrieve the name of the character to open. We just won't do anything when that happens.

```lua
Message: .../totalRP3/Modules/Register/Main/RegisterExchange.lua:604: attempt to index upvalue 'characterToOpen' (a nil value)
Time: Fri Mar 20 21:14:07 2026
Count: 1
Stack:
[C]: ?
[Interface/AddOns/totalRP3/Modules/Register/Main/RegisterExchange.lua]:604: in function 'openProfile'
[Interface/AddOns/totalRP3_UnitFrames/Buttons.lua]:58: in function <Interface/AddOns/totalRP3_UnitFrames/Buttons.lua:57>
```

### DIFF
--- a/totalRP3/Modules/Register/Main/RegisterExchange.lua
+++ b/totalRP3/Modules/Register/Main/RegisterExchange.lua
@@ -601,7 +601,7 @@ function TRP3_API.slash.openProfile(...)
 		end
 
 		-- If no realm has been entered, we use the player's realm automatically
-		if not characterToOpen:find("-") then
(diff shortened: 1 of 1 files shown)


Raenore (approved):

Meorawr (approved):
Rae didn't leave a comment and now I'm sad

PR #1314, opened 2026-03-22
Fix MSP conflicts further

Solanya:
Silly me forgot you could disable the MSP module and therefore not have any msp_RPAddOn. We won't display the warning icon or block comms if that's the case.

However, in the process of figuring out why transfers don't work, I realized my comms fix was incorrect. I don't think there is a way to allow you to query other players profiles while refusing queries yourself without leaving some issues, since the very first comms message includes things like client/version numbers/info fields versions which then get saved and displayed on the tooltip. So no profile comms at all if you misbehave.


Raenore (approved):
Naughty MSP disablers, you can't just do that.

Meorawr (approved):

PR #1315, opened 2026-03-24
Add secret test to GameTooltip anchor

Solanya:
Hopefully this is a temporary issue but for some reason GameTooltip anchors are now secret? So I guess we'll use the default anchor for now and pray...


Raenore (approved):
We all need an anchor in our lives, unless it weighs us down.

PR #1317, opened 2026-03-30
Patch/ignore list tweaks

keyboardturner:
it always bothered me if you left click something in the ignore list it just instantly goes away without any confirmation
this basically handles editing a reason and it also introduces an extra step to remove somebody 

Wow_2026-03-30_11-45-47

### DIFF
--- a/totalRP3/Modules/Register/Main/RegisterList.lua
+++ b/totalRP3/Modules/Register/Main/RegisterList.lua
@@ -272,15 +272,6 @@ local PROFILE_NOTES_ICON = "|TInterface\\AddOns\\totalRP3\\Resources\\UI\\ui-ico
 local WALKUP_ICON = "|TInterface\\AddOns\\totalRP3\\Resources\\UI\\ui-icon-walkup:15:15|t";
 local MATURE_CONTENT_ICON = Utils.str.texture("Interface\\AddOns\\totalRP3\\resources\\18_emoji.tga", 15);
 
-local function onIgnoredActionSelected(value, unitID)
-	if value == "remove" then
-		unignoreID(unitID);
-		refreshList();
-	elseif value == "edit" then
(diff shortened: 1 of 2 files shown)


Solanya (dismissed):

Meorawr:
Doesn't this still have the same one-click-away-from-destruction problem though if you accidentally click remove instead of edit?

keyboardturner:
yeah i just didn't know where the confirmation box is in trp's code

Meorawr:
https://github.com/Total-RP/Total-RP-3/blob/e7c90aa8bf4909215b3f46af72b3fadad90a1964/totalRP3/Core/Popup.lua#L194

keyboardturner:
names are now sorted alphabetically (it was kinda weird this wasn't the case before also) and you get a confirmation box when pressing the remove option
Wow_2026-03-31_11-10-26

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -929,7 +948,12 @@ local function getIgnoredLines()
 	TRP3_RegisterListHeaderGuildTT:Disable();
 	TRP3_RegisterListHeaderRealmTT:Disable();
 
-	return GetKeysArray(getIgnoredList());
+	local ignoredArray = GetKeysArray(getIgnoredList());
+	table.sort(ignoredArray, function(a, b)
+		return a:lower() < b:lower();
```
Meorawr: This needs to use the comparison functions in StringUtil for UTF-8 aware sorting.

Meorawr (dismissed):

Raenore (dismissed):
Good change to properly handle a destructive operation, thank you for that Keyb.

Only note is the loc inconsistency versus all our others.
However Sol mentioned he will tweak it still so should be all good.

Solanya (approved):

PR #1321, opened 2026-04-20
Disable xtensionxtooltip2 swapping to the bottom

Solanya:
Unfortunately, switching channel order causes taint leading to errors as soon as you get into chat lockdown. I will scream into the void hoping this gets resolved eventually but in the meantime, we have to disable this and bear 🐻 with the anger of people complaining their channels are out of order again...


Solanya:
bob is so easily tricked

Meorawr (approved):
Getting in a ticket to un-fuck this would be good.

Raenore (approved):
xtensionxtooltip2 can rest easy for some time, no longer being the bottom.

Fix looks good, given it literally is just disabling the entire thing.

PR #1323, opened 2026-04-20
fix error when trying to use tokens during encounters

Solanya:
If you type a token in chat during unit lockdown, we still try to parse it, causing issues. We'll just use the default value in such cases to prevent errors.


Solanya:
I'm approving myself 🏅

PR #1324, opened 2026-04-20
Request MSP profile even if TT field is missing

Solanya:
We found a MSP profile that couldn't be properly retrieved with TRP despite working fine with MSP addons. Through investigation, we found the TT field was nil, but were unable to determine how that even happened. For now, we'll just assume if the field is nil, we let the process go ahead and save the data anyway, as the issue seems very rare anyway.


Raenore (approved):
What is time really? Just a construct of human perception?

This fix seems fine, if not we'll just blame Sol. That, too, is timeless.

PR #1327, opened 2026-04-26
Fix pet profile check in encounters

Solanya:
I'm kinda surprised nobody reported this yet, but I got the same error starting M+ with a hunter in my party (and me as a warlock) multiple times yesterday. Not sure which pet was causing it, but regardless of it, I've added this check so it doesn't error out anymore.


Raenore (approved):
Good change, Sol.

Thank you for that, you the man. 🫡

Solanya:
Might as well add this to the PR, there was still an error upon starting M+ as we were trying to send a secret pet name through the profile query.

Raenore (approved):
Good change, Sol.

Thank you for that, you the man. 🫡

x2

PR #1329, opened 2026-05-23
Add LibChatFilter for greater inter-addon chat transform compatibilities

Ghostopheles:
[LibChatFilter](https://github.com/Ghostopheles/LibChatFilter) coordinates outgoing message hooks to ensure addons aren't fighting and/or overwriting each other when it comes to mutating messages before they're sent out. It also ensures that the chat history buffer is separated from the outgoing message itself, so users won't see post-transform text in their history.

Will fix the TRP3 chat token parsing when in the presence of these addons (after their respective PRs are merged):
- https://github.com/Ghostopheles/Chattery/pull/2
- https://github.com/keyboardturner/Languages/pull/2

### DIFF
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -40,6 +40,7 @@ externals:
     url: https://www.townlong-yak.com/addons.git/taintless
     commit: default
   totalRP3/Libs/UTF8: https://repos.curseforge.com/wow/utf8/trunk
+  totalRP3/Libs/LibChatFilter: https://github.com/Ghostopheles/LibChatFilter
 
 manual-changelog:
   filename: CHANGELOG.md
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -28,6 +28,7 @@ externals:
   totalRP3/Libs/ChatThrottleLib: https://repos.curseforge.com/wow/chatthrottlelib/trunk
   totalRP3/Libs/Chomp: https://github.com/wow-rp-addons/Chomp
   totalRP3/Libs/LibChatAnims: https://repos.curseforge.com/wow/libchatanims/trunk/LibChatAnims
+  totalRP3/Libs/LibChatFilter: https://github.com/Ghostopheles/LibChatFilter
   totalRP3/Libs/LibDataBroker-1.1: https://github.com/tekkub/libdatabroker-1-1
   totalRP3/Libs/LibDBCompartment: https://github.com/Meorawr/LibDBCompartment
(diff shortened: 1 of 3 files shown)


totalRP3/Modules/ChatFrame/ChatFrame.lua
```diff
@@ -790,39 +791,23 @@ function hooking()
...
+			["%xpf"] = GetTokenSafe(player:GetFirstName(), ""),
+			["%xpl"] = GetTokenSafe(player:GetLastName(), ""),
+			["%xt"] = GetTokenSafe(getUnitRPName("target"), TARGET_TOKEN_NOT_FOUND),
+			["%xf"] = GetTokenSafe(getUnitRPName("focus"), FOCUS_TOKEN_NOT_FOUND),
+			["%xp"] = GetTokenSafe(player:GetFullName(), ""),
+		};
+		text = string.gsub(text, "%%x.[fl]?", tokens);
+		return text;
+	end
+	LibChatFilter.RegisterTransform(ParseTokens);
```
Meorawr: I'd prefer this call to fill out all the optional parameters explicitly. I can't tell at a glance what track we're registering for here unless I go back to the repo and remind myself that it's SEND.

.pkgmeta
```diff
@@ -40,6 +40,7 @@ externals:
     url: https://www.townlong-yak.com/addons.git/taintless
     commit: default
   totalRP3/Libs/UTF8: https://repos.curseforge.com/wow/utf8/trunk
+  totalRP3/Libs/LibChatFilter: https://github.com/Ghostopheles/LibChatFilter
```
Meorawr: I went through the ordeal of alphabetically sorting this list, and here you are ruining my life.

totalRP3/totalRP3.toc
```diff
@@ -49,6 +49,7 @@ Libs\LibMSP\LibMSP.xml
 Libs\LibRPMedia\lib.xml
 Libs\LibWindow-1.1\LibWindow-1.1.lua
 Libs\UTF8\lib.xml
+Libs\LibChatFilter\lib.xml
```
Meorawr: Of course, dependencies in a TOC have a more rigid ordering... but this could be right after LibChatAnims so I don't have to cry at the ordering, y'know?

Meorawr (changes requested):

Solanya (approved):

PR #1331, opened 2026-05-30
Preserve Self-Notes When A Profile Is Imported

Bitwise1057:
Notes written on your own profile are lost when that profile is imported into another slot. This re-keys the self-note so it survives the import, and fixes the same latent bug on the login path that rotates the default profile's ID.

---
Profile notes are stored in `profile.notes`, keyed by the target profile's ID:
- A note about another player is keyed by that player's profile ID.
- A note on your own profile is keyed by your own profile's ID and reads it back the same way.

A profile's ID is tied to its slot and not the data it holds. On import, the destination slot keeps its own ID but the imported self-note still carries the source profile's ID as its key. The lookup misses and the note renders blank.

Notes about other players are keyed by those players' IDs which don't change across the import so they come through fine which is why only the self-note disappears.

`updateDefaultProfile()` has the same flaw: it assigns the default profile a fresh ID on every login without re-keying the self-note.

In both cases the data is lost as it is orphaned under a key nothing looks up.

---

Note: This attempt was unable to recover already orphaned notes as their old key is gone from any lookup and only notes handled after the changes were made are fixed.

### DIFF
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -605,7 +605,7 @@ function TRP3_API.profile.init()
 		if version ~= nil then
 			local import = function()
 				data.profileName = profiles[currentProfileID].profileName;
-				wipe(profiles[currentProfileID]);
+				profiles[currentProfileID] = nil;
 				local destinationProfileID = currentProfileID;
 				if not profiles[errorOrOldProfileID] then
 					destinationProfileID = errorOrOldProfileID;
--- a/totalRP3/Core/Profiles.lua
+++ b/totalRP3/Core/Profiles.lua
@@ -602,7 +602,7 @@ function TRP3_API.profile.init()
 		local code = TRP3_ProfileImport.content.scroll.text:GetText();
 		local version, errorOrOldProfileID, data = TRP3_ProfileUtil.DeserializeProfile(string.trim(code));
 
-		if version ~= nil then
+		if version ~= nil and data ~= nil then
 			local import = function()
 				data.profileName = profiles[currentProfileID].profileName;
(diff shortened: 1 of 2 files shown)


Solanya:
Default profile change is irrelevant, can't access its own notes tab.
As far as the other part, I have additional changes I wanted to make but I can just take care of those.

Bitwise1057:
> Default profile change is irrelevant, can't access its own notes tab. As far as the other part, I have additional changes I wanted to make but I can just take care of those.

Makes sense on the default profile block, I've reverted the changes I made regarding it.

I know this was an extremely low priority issue for you so I wanted to take a crack at getting it handled and out of the way for more important objectives, feel free to close this PR if you'd rather fold your additional changes in as one set elsewhere. Whatever's least friction for you.

totalRP3/Core/Profiles.lua
Ghostopheles, line 1: Should probably move the new part down to be inside the `if data then` block to avoid ghosts, right?

Solanya: No, the if data would need to move

Solanya:
Imported profile will now reuse its old profile ID when possible (if the key isn't already in use) so other players notes can also remain functional whenever possible. Migration is still in place in cases where the old profile ID is unavailable.
Also did some renaming of variables, the existing ones were a bit unclear; and moved the data tab check (although probably unnecessary) in the correct position.

Raenore (dismissed):
It's nice to see this being handled now.

Thank you @Bitwise1057 and @Solanya for taking care of this.

Solanya (approved):

PR #1332, opened 2026-06-16
Rework alpha check in CreateColor

Solanya:
There was an error with secret alpha values in nameplates as Acquire would try to do math on secrets. I just changed so it shouldn't trip on secrets anymore (probably)

image


Raenore (approved):
Good job on catching this, Sol.

The commit history surely creates some color in and by itself. 😉

PR #1335, opened 2026-07-02
Profiles Rework 2026

Raenore:
This was meant to release in like 2024? 2025? You know how it goes.

Definitely will require a lot of checking given how it changes profiles.lua as a whole, probably tweaking too if the two overlords don't like it (this has not changed code-wise since its original inception over a year+ ago).

Note that this includes the note fixes on profile import from #1331.

Below is a video of it in action (video dated March 2025).
Video still has the old import/export code, but it is adapted to work with PEM etc from #1204.

https://github.com/user-attachments/assets/f92712a3-9cc5-4aef-ab8f-209690831120

Some extra notes:
- Import no longer happens on / overwrites an existing profile. It creates a new profile instead.
- There are clear errors telling the user when a profile name is duplicated or other potential issues.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -703,6 +703,7 @@ stds.wow = {
 		"ReloadUI",
 		"RemoveChatWindowChannel",
 		"ResetCursor",
+		"RunNextFrame",
 		"RoundToSignificantDigits",
 		"SafePack",
 		"Saturate",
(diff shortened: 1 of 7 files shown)


Ghostopheles:
finally

Meorawr (changes requested):
ㅤ

Raenore:
> ㅤ

Change deez, but done.

Solanya (approved):

PR #1336, opened 2026-07-02
Relation Reorganization

Bitwise1057:
This PR serves to add a small quality of life change to the Relations tab by allowing users to organize each option by adding a Move Up / Move Down option to the pre-existing hamburger menu.

Changes to organization are reflected in their respective dropdowns as well.

image


Solanya:
We already have plans to modify this UI, including using drag & drop for reordering like we do in other places in the addon.

PR #1337, opened 2026-08-02
Classic Era fixes

Solanya:
- MenuUtil.ShowTooltip/HideTooltip is deprecated on Retail but Classic yeeted it out already so might as well replace them now
- `icons_64x64_important` atlas key is not on Classic Era so used a different icon for it
- Tooltip data handler is present in the Era code but not used, so brought back the old tooltip code

As a result, Classic Era can be brought back to the present and supported again. I'm sure everyone is ecstatic about that.


Raenore (dismissed):
"We're so back."
-Someone, probably.

Code looks fine, but I got no testing chars for Classic so I'll take your word for it actually working. 👀

totalRP3/totalRP3.toc
```diff
@@ -1,4 +1,4 @@
-## Interface: 120007
+## Interface: 120007, 11509
```
Meorawr: TBC should work too, they're the exact same API. Mists is probably also fine.

Solanya: TBC works except for all the missing icons but like, fuck that
I'll try it on MoP later just to check

Solanya:
Brought back the TBC icon replacements and the transparent icon from the addon list, as well as reenabled TBC/MoP. Classic is fully back on the menu.

Meorawr (approved):

PR #1340, opened 2026-08-10
Reorder relations and fix deletion

Solanya:
- Relations can now be reordered by using the existing tech from the characteristic traits reordering. This was done by moving the existing code out of the characteristics page and doing some adaptation so it would also work for the relations page. I'm not super happy about the state of the changes, as the way the order is set in relations is different from the traits. If there are visible improvements that could be made, feel free. If we gotta rework the structure of relations as a whole, I guess we can delay the feature, this was the only way I could see it being made for 12.1 but it's not critical.

- Also removed the deletion condition, as it should have been from the start. We just remove the relation from all assigned profiles when deleting it rather than blocking deletion until all the assignments are manually removed.


Solanya:
Y'know I could have split both changes so we could merge deletion on its own... If we wanna do that I'll figure it out.

Raenore (changes requested):
There are some issues going on with Additional Info and Personality traits reordering.

Additional Info seems to not move the icons around when reordering.
Personality traits seems to have issues in moving entirely.

Only checked for a few minutes, but raising it nonetheless.

Relations reordering works just fine though, so GG on that.

Raenore:
With these changes, additional info & personality traits can reorder again properly.

Also added a fix for relations reordering that I came across.. `updateOrderAfterDrag` now only walks the relations actually shown, as leftover hidden widgets were overwriting the new order, silently reverting the move after deleting a relation.

Raenore (approved):

PR #1341, opened 2026-08-11
High-contrast about page

Ghostopheles:
Adds a config option to increase contrast on about pages by setting the background to the default, and stripping out custom colors.

~~Intentionally does not prevent you from changing the background or text color on your own profile, but important to note that you won't be able to see any changes if this setting is enabled.~~ (Not anymore)

Applies to all three templates, and their child frames, if any.

### Before
image

### After
image


totalRP3/Locales/enUS.lua
```diff
@@ -494,6 +494,8 @@ The description doesn't have to be limited to |cnGREEN_FONT_COLOR:physical descr
 	CO_REGISTER_ABOUT_H3_SIZE_TT = "Size of the text between {h3} tags. Default: %d",
 	CO_REGISTER_ABOUT_P_SIZE = "Paragraph text size",
 	CO_REGISTER_ABOUT_P_SIZE_TT = "Size of the text outside of header tags. Default: %d",
+	CO_REGISTER_ABOUT_DISABLE_CUSTOM_BGS = "Use high contrast backgrounds",
+	CO_REGISTER_ABOUT_DISABLE_CUSTOM_BGS_TT = "Uses only the high contrast background when viewing a profile's about page.\n\nDoes not prevent you from changing your own about page backgrounds.",
```
Ghostopheles: I'm not super into the tooltip string, so please shower me in your suggestions

totalRP3/Modules/Register/Characters/RegisterAbout.lua
```diff
@@ -104,8 +112,30 @@ getDefaultProfile().player.about = {
 
 local draftData;
 
+local function stripColorTagsFromAboutText(text)
+	text = text:gsub("{/?col[^}]*}", "");
```
Meorawr: Given the state of the average profile, I'd be concerned that this matches too much and erase crap like "here's my butterfly {collection}".

```suggestion
	text = text:gsub("{col:[^}]+}", ""):gsub("{/col}", "");
```

Meorawr (commented):
Companion pets have abouts too.

totalRP3/Locales/enUS.lua
```diff
@@ -494,6 +494,8 @@ The description doesn't have to be limited to |cnGREEN_FONT_COLOR:physical descr
 	CO_REGISTER_ABOUT_H3_SIZE_TT = "Size of the text between {h3} tags. Default: %d",
 	CO_REGISTER_ABOUT_P_SIZE = "Paragraph text size",
 	CO_REGISTER_ABOUT_P_SIZE_TT = "Size of the text outside of header tags. Default: %d",
+	CO_REGISTER_ABOUT_HIGH_CONTRAST = "High contrast about page",
+	CO_REGISTER_ABOUT_HIGH_CONTRAST_TT = "Raises the contrast of the about page by removing any non-default backgrounds or text colors.\n\nDoes not prevent you from changing the background or text color in your own profile.",
```
Meorawr: After some thoughts;

- Settings controls should describe the benefit to the user, not the mechanism being applied.
- The tooltip should then move onto the behaviour of the setting in terms of what the user will see.
- As you had before, any "surprising" exceptions or scope to a setting should come at the end after some blanks.
- Apply highlights to at-most two things that the user should be drawn to if they were quickly scanning the tooltip (in this case, the direct effect, and the key exception).
- Avoid any technical language - you didn't really have any, but I was iffy on "contrast" as a word so I scrubbed it. We don't really need that word to describe what the setting does for the user.

```suggestion
	CO_REGISTER_ABOUT_HIGH_CONTRAST = "Make About pages easier to read",
	CO_REGISTER_ABOUT_HIGH_CONTRAST_TT = "Displays About pages with a |cnGREEN_FONT_COLOR:consistent background and text color|r, hiding custom backgrounds and text colors.\n\n|cnGREEN_FONT_COLOR:You can still customize|r the background and text colors of your own profile.",
```

Meorawr (dismissed):
I'm fine with pretending companion pet profiles don't exist for now at least.

Meorawr (approved):

Solanya (approved):

PR #1343, opened 2026-08-11
Fix ScrollingEditBox readOnly flag acting out of line

Ghostopheles:
previous check disregarded a `false` value for `.readOnly`. It now behaves better


Solanya (approved):
Makes sense with current times that we'd pretend false things are true

PR #1345, opened 2026-08-12
Fix 12.1 Nameplate issues

Raenore:
This PR does four things:
1. Fixes name to be white font again when they are within the nameplates' health bar.
2. Blizz now exposes "Only Show Names" from their settings, mirror this CVar when it is set to our settings so it stays in sync (and make the setting query it in `OnShow` as well).
3. Add a flyway so that when people log in on 12.1, their Blizzard CVar is synchronized to what they set on our settings page.
4. The old system remains as a fallback for Classic, as I can't test myself if the CVar there also persists now or not.

### DIFF
--- a/totalRP3/Modules/Flyway/Flyway.lua
+++ b/totalRP3/Modules/Flyway/Flyway.lua
@@ -3,7 +3,7 @@
 
 TRP3_API.flyway = {};
 
-local SCHEMA_VERSION = 24;
+local SCHEMA_VERSION = 25;
 
 if not TRP3_Flyway then
 	TRP3_Flyway = {};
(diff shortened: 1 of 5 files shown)


Meorawr (changes requested):
I'd rather us not store the setting at all in our config if the cvar is now persistent.

Solanya:
Removed the config key outside of its use in the flyway to only rely on the CVar now that it's permanent.

Raenore (commented):
I cannot "approve" my own PR.

However I've tested the latest changes by Sol on 12.1 PTR and this seems to work fine.
Blizzard's setting and our own are now sync'd, without any further fluff.

Solanya (approved):

PR #1347, opened 2026-08-12
Collect issues (luacheck & editorconfig) together before marking job as failed

Raenore:
Note: This will mean that even on failures in luacheck or editorconfig that -that- specific step is marked as 'success' (green) even if there's issues found. The 'fail' (red) check now happens at the "Check lint results" step instead.

The main thing is that the build shows it failed, and that we collect all issues together instead of having multiple failure pushes because one step blocked the other.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,17 +17,28 @@ jobs:
           submodules: recursive
 
       - name: Run Luacheck
+        id: luacheck
         uses: nebularg/actions-luacheck@v1
+        continue-on-error: true
         with:
           args: --no-color -q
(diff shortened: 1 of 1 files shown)


Meorawr (dismissed):
failure

Raenore:
Someone's gotta take the L for the team.

Thanks for taking responsibility @Solanya.

Raenore:
Tested and confirmed working, ready for merge.

As soon as this is merged, all -future- PR jobs will run with these settings.
For current PRs they will probably have to be rebased on `main` so they have the proper job files.

Meorawr (approved):

PR #1351, opened 2026-08-16
Github Actions Cleanup

Meorawr:
My wow-ui-schema repository is going to get archived as Blizzard have long-since fixed all the errors in their own XSD.

I need to update all of our XSD references in this repository to not reference the stale XSD in that repo. While I'm here, I'd also like to have all of our XML files consistently reference the exact same local XSD file that we've had to commit into the repo for xmllint validation.

So, this does a few things;

- Relocates .github/scripts/ui.xsd -> Types/UI.xsd. The XSD isn't a github-specific thing, and Types is a fairly generic directory-of-schema-like-stuff as-is.
- Adds a totalRP3.xsd file to our main addon directory. This, itself, just has an include against ../Types/UI.xsd - but, in the future could be expanded to include intrinsics if for some reason we ever find ourselves using them.
- Updates every XML file to use a relative path reference to totalRP3.xsd.

I also then wanted to make pre-commit and our actions setup a bit... easier to use on Windows.

- Swaps out usage of xmllint for python and its lxml module, which is a bit easier to install.
- Swaps out the Makefile for a Justfile, [which actually supports Windows as a first-class target](https://github.com/casey/just).

And then while I was at it, I noticed half our actions were using inconsistent versions of external things, wasn't validating XML at all, and our setup of the notification webhook wasn't correct in most of the workflows.


Solanya (approved):

Raenore (approved):
Checking over the actions, these seems in order.

And if not, they'll just fail on GitHub and we'll catch it (as this won't break anything for the users).

PR #1352, opened 2026-08-17
Avoid secret explosions from Nameplate color hooks

Meorawr:
Avoid the use of our cached color machinery for colors obtained through secure posthooks in the Blizzard nameplate decorator. These can contain secret values, which error when we generate a cache key for the object.

We don't want to construct our brand of color objects where the components are individually secret as multiple methods on them would, also, implicitly fail in the face of secrets. Instead, pack the color components into a raw table with no methods and update the one place of code that's unpacking them to work with direct field access.

### DIFF
--- a/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
+++ b/totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
@@ -27,12 +27,12 @@ end
 
 local function UpdateFontStringWidgetColor(widget)
 	local desiredColor = widget.TRP3_overrideColor or widget.TRP3_originalColor;
-	CallUnhookedMethod(widget, "SetVertexColor", desiredColor:GetRGBA());
+	CallUnhookedMethod(widget, "SetVertexColor", desiredColor.r, desiredColor.g, desiredColor.b, desiredColor.a);
 end
 
 local function UpdateStatusBarWidgetColor(widget)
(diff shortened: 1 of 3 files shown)


totalRP3/Modules/NamePlates/NamePlates_Blizzard.lua
```diff
@@ -27,12 +27,12 @@ end
 
 local function UpdateFontStringWidgetColor(widget)
 	local desiredColor = widget.TRP3_overrideColor or widget.TRP3_originalColor;
-	CallUnhookedMethod(widget, "SetVertexColor", desiredColor:GetRGBA());
+	CallUnhookedMethod(widget, "SetVertexColor", desiredColor.r, desiredColor.g, desiredColor.b, desiredColor.a);
```
Meorawr, line 30: While `color:GetRGBA()` is portable between both Blizzard's ColorMixin and our own Color metatable, in this case I feel it best that -  given the secrecy issue - we avoid both objects entirely, as not all methods on both of those can be called by us when the color components are secret.

It'd be confusing to have to remember which methods are secret-safe and which ones aren't. No methods? No problem!

Raenore (approved):
Tested while forcing `/console addonMapRestrictionsForced 1` in Stormwind.

With the 2nd commit, I don't see any errors anymore relating to our `Color.lua` file.
Mind you, this was done at like 3.30 AM in EU so there weren't tons of player nameplates.

Solanya (approved):

PR #1360, opened 2026-09-08
Internationalized UI search and sorting improvements

Meorawr:
This integrates the new 12.1.5 `C_Intl` APIs into most search and sorting areas in our UI.

The main benefits of these changes are locale-aware collation and string transforms. For example, the directory now ignores accents when it comes to sorting and places them in natural list positions. We can additionally globally strip all non-letter characters across locales from sort keys, preventing Unicode symbol spam from negatively affecting the directory.

image

When searching, we additionally make use of case-and-accent-insensitive collations - such that a search for a latin input of "resume" will find my stupidly named pet, "Résumé".

image

With this we can largely remove all of our usage of old sort and search code patterns - so, usage of APIs like `strcmputf8i` are now gone, and we no longer (incorrectly) case-fold by converting strings to lowercase.

There's catches to this, however:

1. Classic doesn't have these APIs. Nor does live, so we'll have to deal with the dance of an incompatible update on patch day while rollout occurs across EU.
2. On the PTR, some of our C_Intl API usage - particularly [C_Intl.FindStringMatches](https://warcraft.wiki.gg/wiki/API:C_Intl.FindStringMatches) with single-character inputs, and [C_Intl.Transliterate](https://warcraft.wiki.gg/wiki/API:C_Intl.Transliterate) are slow. Like, stupidly slow. Transliterating a directory of 12k character names takes about 1 second, that level of slow.

Because of this, there's a rather unfortunate and significant set of changes in one of the worst areas of the code - the directory. I basically had to make it perform searching and sorting in an asynchronous task with progress indicators in the UI, because it was _that_ slow.


totalRP3/Core/AdvancedSettings.lua
```diff
@@ -81,7 +81,7 @@ TRP3_API.RegisterCallback(TRP3_Addon, TRP3_Addon.Events.WORKFLOW_ON_LOAD, functi
 	end
 
 	-- Sort locales alphabetically.
-	table.sort(localeTab, function(a,b) return a[1] < b[1] end);
+	table.sort(localeTab, function(a,b) return TRP3_StringUtil.SortCompareStrings(a[1], b[1]) end);
```
Meorawr, line 84: I did a broad pass over `table.sort` usage and - in cases where we're dealing with localized strings - put in a call to a wrapper that'll do a C_Intl based comparison with an appropriate default collation (Primary; so case and accent insensitive).

totalRP3/Core/FunctionUtil.lua
```diff
@@ -50,3 +50,53 @@ function TRP3_FunctionUtil.Throttle(timeout, callback)
...
+---
+---@class TRP3.AdaptiveTimeBudgetOptions
+---@field minimumBudgetSeconds number Minimum slice budget.
+---@field maximumBudgetSeconds number Maximum slice budget.
+---@field catchUpRatio number Fraction of the idle interval to reclaim.
+---@field checkInterval integer Clock-check interval.
+---
+---@param options TRP3.AdaptiveTimeBudgetOptions?
+---@return fun(): boolean
+function TRP3_FunctionUtil.CreateAdaptiveTimeBudgetChecker(options)
```
Meorawr, line 69: The reason this was added is for the Register's async refresh. We need to split the work across multiple frames within a coroutine.

The Icon browser had dumb logic where it'd do a fixed amount of work per frame, with a hack to work around some animation quirks that ensures it never processes >5% of the dataset in a single tick.

I didn't replicate this into the register; the appropriate amount of fixed is hard to estimate - instead, we've got an adaptive solution that tries to schedule work in budgeted time slices.

Broadly, the result here is that with the default budgeting this will allow ~8ms of work to proceed per tick while running at >= 30 FPS. Naturally, above 60 FPS, this will slash the user's framerate to about that amount. However, odds are the user won't really notice this - and, it's less aggressive than the "fixed-amount" of work that the icon browser does.

If the player is below 30 FPS then we'll end up scheduling an amount of work somewhere between an 8ms time slice and a 100ms time slice. We cap at 100ms because we don't want to run foul of Blizzard's strict throttles if, for some reason, the player is filtering their directory in combat.

totalRP3/Core/SortUtil.lua
```diff
@@ -0,0 +1,54 @@
+-- Copyright The Total RP 3 Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+TRP3_SortUtil = {};
+
+---@param array table
+---@param comparator fun(a: any, b: any): boolean
+function TRP3_SortUtil.MergeSort(array, comparator)
```
Meorawr, line 8: Why are we implementing merge sort in Lua? We can't yield coroutines across C calls in Lua 5.1. We want a yieldable equivalent of `table.sort` for the directory - so, we need to do it in Lua.

Now, humorously, this appears to actually be _almost as fast as_ table.sort despite being a Lua implementation. I assume PTR quirk, but I did blink a few times when I'd measured it.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
 end
 
-local SearchableStringCache = setmetatable({}, {
-	__index = function(tbl, str)
-		tbl[str] = GenerateSearchableString(str);
-		return rawget(tbl, str);
-	end,
-});
+-- String Matching
+------------------------------------------------------------------------------
```
Meorawr, line 95: String matching largely got reworked into the concept of a dedicated Matcher object. This stores a query string (both in raw form and prepared form), and maintains a local cache of prepared strings when used for searches or scoring.

This replaces the global caching we were previously doing, and allows for the cache to be evicted when no longer needed - something we now do in the icon browser, for example (which saves us about 4 MB of memory when it's closed).

The downside is that a few operations - like the free functions for FindBestMatch or IsExactOrSubstringMatch - need to create a throwaway matcher. I may revisit this another time and rework it again.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
 	-- Largely just copying what Blizzard does for console autocomplete.
 	-- Lower scores are better.
 
-	local substringStartIndex = string.find(candidateText, searchText, 1, true);
-	local editDistance = CalculateStringEditDistance(searchText, candidateText);
+	local substringStartIndex = FindFirstStringMatch(candidateText, searchText, collation);
 
-	if not substringStartIndex and editDistance == math.max(#searchText, #candidateText) then
-		return math.huge;  -- Supplied text doesn't match at all.
+	if not substringStartIndex then
```
Meorawr, line 125: The removal of edit distance checking here is intentional for reasons I can't remember. I think it related around partial matches of strings that really shouldn't ever match.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
 	end
 
-	if requireExactMatch then
-		return (strcmputf8i(searchText, candidateText) == 0);
+	local candidate = self:GetCandidate(candidateText);
+
+	if self.predicate then
+		return self.predicate(self.preparedQuery, candidate);
+	elseif self.mode == "exact" then
+		return self.preparedQuery == candidate;
```
Meorawr, line 201: Previously, exact match for whatever reason did a case-insensitive comparison unconditionally. That's still technically the default, but it's no longer reflected here - the strcmputf8i call was replaced with raw equality because the defaults of our matcher are such that the prepared query and candidate would be case folded already, unless explicitly opted out.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
+---@private
+---@param text string
+function StringMatcher:GetNormalizedString(text)
+	local normalized = self.normalizedStringCache[text];
+
+	if normalized == nil then
+		normalized = text;
+
+		if self.normalization == "destructive" then
+			normalized = string.gsub(normalized, "[%p%c%s]+", "");
```
Meorawr, line 312: May be worth using a Transliterate transform here one day, but for now I'm inclined to just strip regular Latin punctuation and whitespace.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
+---@class TRP3.SortKeyOptions
+---@field collation Enum.CollationStrength?
+---@field transliterator string?
+---@field projection? fun(value: any): string?
+---@field emptyKeyPosition TRP3.SortKeyEmptyPosition?
+
+---@param value any
+---@param options TRP3.SortKeyOptions?
+---@return string
+function TRP3_StringUtil.GetSortKey(value, options)
```
Meorawr, line 370: Sorting just comes down to this one function. The main thing here is C_Intl.GetSortKey, which takes an input string and returns an opaque binary key that can undergo regular lexicographical comparison - so, call this and then just do a regular `a < b` test.

totalRP3/Core/StringUtil.lua
```diff
@@ -23,116 +33,365 @@ function TRP3_StringUtil.DequoteString(string)
...
+	local prefix = "";
+
+	if options.emptyKeyPosition ~= nil then
+		local isEmpty = (normalizedValue == "");
+		local emptyValuesSortLast = (options.emptyKeyPosition == TRP3_SortKeyEmptyPosition.Last);
+		prefix = (isEmpty == emptyValuesSortLast) and "\255" or "\000";
+	end
+
+	local collation = options.collation or Enum.CollationStrength.Primary;
+	return prefix .. (C_Intl.GetSortKey(normalizedValue, collation) or "");
```
Meorawr, line 396: The `or ""` is because the API for whatever reason returns nil for an input of `("", Enum.CollationStrength.Primary)`. I assume ICU internally wants to return a zero-length string, and the API is saying "let's make that a nil" - but that's wrong and bad.

totalRP3/Modules/Map/MapPoiMixins.lua
```diff
@@ -110,6 +82,31 @@ function GroupedCoalescedMapPinMixin:GetMouseOverPinsByTemplate(pinTemplate)
...
+
+		local orderA = a.sortOrder or 0;
+		local orderB = b.sortOrder or 0;
+
+		if orderA ~= orderB then
+			return orderA < orderB;
+		end
+
+		return (a.sortName or "") < (b.sortName or "");
+	end
```
Meorawr: No changes here now (there were some, but they're lost to time).

totalRP3/Modules/Register/Main/RegisterRequestIndicator.lua
```diff
@@ -21,24 +21,19 @@ function TRP3_RegisterRequestIndicatorMixin:OnLoad()
 	-- indices must be populated or we'll end up with a mostly invisible
 	-- frame. Also, this requires render layer flattening to be enabled or no
 	-- fade will occur.
-	--
-	-- As of writing - Classic Era (1.15.7) doesn't support this tech, so
-	-- it'll have no fading.
-
-	if self.Background.SetAlphaGradient then
```
Meorawr, line 28: Cleanup; I was playing with some small UI tweaks in the register and noticed we had this old check that's no longer relevant.

totalRP3/Modules/Register/Main/RegisterTemplates.xml
```diff
@@ -33,4 +33,73 @@
...
+					<Anchors>
+						<Anchor point="CENTER"/>
+					</Anchors>
+				</Texture>
+			</Layer>
+		</Layers>
+		<BarTexture parentKey="Ring" file="Interface\Garrison\orderhalltalents-timer-fill" alpha="0.8">
+			<RadialProgressBar startOffset="0.5"/>
+		</BarTexture>
+	</StatusBar>
```
Meorawr, line 52: https://github.com/user-attachments/assets/cf39d684-7d1b-40d2-b6f0-665d7b0f9dc6

Circular progress bars? In _my_ UI?

totalRP3/Modules/Register/Main/RegisterTemplates.lua
```diff
@@ -154,3 +154,136 @@ function TRP3_RegisterInfoSwatchLineMixin:SetValueColor(color)
...
+function TRP3_RegisterListRefreshToastMixin:UpdateProgress()
+	local state;
+
+	if self.task ~= nil then
+		state = self.task:GetState();
+	end
+
+	-- SetOnUpdateMode is required to flush hidden status bar values;
+	-- interpolated updates can continue from stale rendered value if a
+	-- SetValue call without interpolation is made while the bar isn't visible.
```
Meorawr, line 278: fuckin' blizzard bugs man

totalRP3/Modules/Register/Main/RegisterUIList.xml
Meorawr, line 1: I made some visual tweaks to the register UI while I was here. Not huge ones - I just removed the weird dead space around the table and made the list/header tightly hug the frame border. Also fixed a small issue where the highlight on rows was bleeding out above/below the background asset, and added a small alternating color pattern to rows.

Image

totalRP3/Resources/UI/Templates.xml
```diff
@@ -5,6 +5,21 @@
...
+
+	<Texture name="TRP3_ProgressBarBorderRightTemplate" file="Interface\AchievementFrame\UI-Achievement-ProgressBar-Border" virtual="true">
+		<Size x="24" y="0"/>
+		<TexCoords left="0.812" right="0.8745" top="0" bottom="0.75"/>
+	</Texture>
+
+	<Texture name="TRP3_ProgressBarBorderCenterTemplate" file="Interface\AchievementFrame\UI-Achievement-ProgressBar-Border" virtual="true">
+		<Size x="24"/>
+		<TexCoords left="0.0625" right="0.812" top="0" bottom="0.75"/>
+	</Texture>
```
Meorawr, line 21: These were used by the icon browser. I was also originally using the same modal input-blocking "Searching..." overlay for the directory, but it was annoying. Keeping these templates ripped out though.

totalRP3/UI/Browsers/IconBrowser.lua
```diff
@@ -48,7 +48,7 @@ function IconBrowserSearchTask:__init(predicate, model)
 	self.total = model:GetIconCount();
 
 	-- On small data sets do only 5% of the set per tick to avoid UI flicker.
-	self.step = math.min(500, math.ceil(model:GetIconCount() / 20));
+	self.step = math.min(100, math.ceil(model:GetIconCount() / 20));
```
Meorawr, line 51: Intentional lowering with how much slower the C_Intl APIs appear to be. May put the adaptive stuff in here one day, but not yet (I'd need to solve the anim flicker problem).

Types/Game.d.lua
Meorawr, line 1: Did a pass over our manual type annotations and pruned the unused ones.

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -44,7 +43,11 @@ local currentlyOpenedProfilePrefix = TRP3_API.register.MENU_LIST_ID_TAB;
 local REGISTER_PAGE = TRP3_API.register.MENU_LIST_ID;
 
 local function openPage(profileID, unitID)
-	local profile = getProfile(profileID);
+	local profile = TRP3_API.register.getProfileOrNil(profileID);
+	if not profile then
+		return;
+	end
```
Meorawr, line 49: You'll see this transform a bunch in places. We don't block user input while the directory is refreshing -  this means we need to guard against users clicking deleted profiles while the UI rebuilds itself.

Not blocking input is a choice; essentially, it's really annoying to have the directory throw up a giant "Searching..." screen each time you change tabs (or, simply open the page for it).

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -353,8 +405,6 @@ end
 local function decorateGenericLine(line)
 	line.Click:RegisterForClicks("LeftButtonUp", "RightButtonUp");
 	line.Click:SetScript("OnClick", onLineClicked);
-	line.Click:SetHighlightTexture("Interface\\FriendsFrame\\UI-FriendsFrame-HighlightBar-Blue");
-	line.Click:SetAlpha(0.75);
```
Meorawr, line 357: Moved to XML.

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -518,68 +572,86 @@ local function decorateCharacterLine(line, elementData)
 		TRP3_API.FormatShortcutWithInstruction("SHIFT-CLICK", loc.CL_TOOLTIP));
 end
 
-local function getCharacterLines()
-	local nameSearch = TRP3_RegisterListFilterCharactName:GetText():lower();
-	local guildSearch = TRP3_RegisterListFilterCharactGuild:GetText():lower();
+local function CreateCharacterLineBuilder()
```
Meorawr, line 575: The changes in here look scarier at first glance. There's not really any aside from:

- Using the matcher.
- Replacing the immediate loop over profiles with a function return; the function is supplied profiles one by one.
- Returning a PROPERLY STRUCTURED TABLE WITH NAMED FIELDS rather than a stupid 8 billion element array with all the stuff in it, holy shit who even writes that sort of code

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -1060,13 +1394,6 @@ end
 TRP3_API.RegisterCallback(TRP3_Addon, TRP3_Addon.Events.WORKFLOW_ON_LOAD, function()
 	createTutorialStructure();
 
-	-- To try, but I'm afraid for performances ...
-	TRP3_API.RegisterCallback(TRP3_Addon, Events.REGISTER_DATA_UPDATED, function(_, unitID, _, dataType)
-		if TRP3_MainFrame:IsShown() and getCurrentPageID() == REGISTER_LIST_PAGEID and unitID ~= Globals.player_id and (not dataType or dataType == "characteristics") then
-			refreshList();
-		end
-	end);
```
Meorawr, line 1068: We no longer update the directory live when register data is received. Doing so all this time was bad anyway - beyond a thousand profiles in the directory, you'd hitch the client on every update to any profile while the directory was open, and odds are you're probably not even looking at the exact entries in the list for people you're receiving data from. It's more problematic to keep this now with the async refresh - so removing it entirely. The user can manually refresh by clicking the apply filters button anyway.

totalRP3/Core/FunctionUtil.lua
SpiritOfSilvermoon: A couple of questions :

First, a rather naive question - how does one test this changeset overall without addons being enabled on the 12.1.5 PTR (or is there some cvar etc. to force addons to be enabled anyway in the current PTR state that I'm (quite possibly) unaware of) ?

Second, potentially 8ms per frame still seems like quite a lot (let alone 100ms which is definitely going to result in very jarring human perceivable impact for just about anybody).  Out of curiosity, how slow do things get from a responsiveness to complete pending work standpoint for typical sized directories, if a fixed realtime wall clock budget of, say, 100-300usec per wakeup work slice is used (perhaps with a sanity backstop of a minimum amount of work progressed before yielding) ?

Especially considering there may well be other addons doing time-sliced per-frame work of their own peculiar nature, and of course those all add up together (with all the game's own per-frame work) if the stars happen to align for multiple addons doing that sort of thing concurrently.

Meorawr: Given a synchronous directory search case that takes 1s, attempting to batch that work into 300us time slices would require `1000ms / 0.3ms` (~3334) steps.

Wake-up granularity available to addons is once per frame. Ignoring the problem that profiling timestamp granularity exposed to addons doesn't expose microseconds, at 60fps, `3334 / 60` would mean our previously-one-second search takes nearly a whole minute to complete.

I would consider that a slightly worse user experience.

Practically, we can't tell how much work other installed addons are going to consume on the current or future frames.

We make use of elapsed timings between frames to counterbalance, but that's added complexity for questionable gain.

Given directory search requires the window effectively be focused and direct user input given, it's reasonable for us to take a majority of a slice to complete their request.

The only requirements we have are to not visibly lock up the client for long periods (200ms+), noting that figure is where script timeout execution limits begin to conditionally kick in. The best balance for responsiveness and completion timing is ~4-12ms chunks.

In the case of a low frame rate client - as in, consistently below 30fps - I'm happy to eschew the "don't lock the client too much" requirement as in that scenario, we don't really have a choice in the matter. That the search completes in a timely manner is somewhat critical to the entire feature.

SpiritOfSilvermoon: Thanks for the reply!  I don't think there's any need for microsecond-precise wakeup times to reduce per-frame work which obviously will never happen on a real machine pumping real frames on any real world scene (i.e., just as the code is already doing, GetTimePreciseSec can be used for microsecond precision accounting of a time budget when checked periodically as work progresses, so this is purely a question about the time budget parameters chosen).

But yes, I agree that if the user is waiting (versus some background task etc.), that there needs to be a careful balance between responsiveness to the user versus framerate impairment, and it's undesirable to keep the user waiting overly long if they are expecting to wait on the results to continue with their workflow.  Do we know if 1 second of cumulative processing time is relatively representative of a typical sorting scenario?

How does one exercise these changes against the 12.1.5 API set at the current time, to try them out?

(For what it's worth, 30fps isn't really a low framerate scenario for my machine.  Of course, I prefer to have settings for visual quality increased.  But that's much more like a typical modern expansion open world scene's complexity ends up being on my system, at my preferred quality levels.  Others may have different machines or different configurations, of course.)

totalRP3/Core/AdvancedSettings.lua
SpiritOfSilvermoon: Shall the end result of this be that filtering in the TRP directory based on user search strings now considers accent and non-accent characters to be identical (e.g. for raw character names)?

(For what it's worth, on NA RP realms, it's very common for people to insert lookalike accent characters into their (typically US English) names to work around limited character namespace availability.  Of course these come with all the attendant issues with it being less easy to type those on a US English configuration.

For that reason, for my own addons, I do try to collapse these for user input search purposes, as I think it ends up being a better user experience in situations where people use these accents to find another way to be named "Joe" (with some accent) as opposed to an actual native language use of those accents.

Historically I've done this in addons I maintain using a giant pattern match that a bright someone else came up with, although it's relatively slow and has to carefully tiptoe around hitting pattern match implementation limits in WoW's Lua dialect, so I'm eagerly awaiting a way to replace it with a much simpler and hopefully more performant facility like this.

Granted, that might be a US English-centric view where those accents are largely used to squeeze extra names out of the namespace than what might be seen in non-English locales.)

In either case, thanks for putting up with my curiosity!

Meorawr: Yes, searches for "moon" would match both "Moonprincess" and "Møōnprincess" alike. In terms of sorting, they also group together and sort naturally.

totalRP3/Core/FunctionUtil.lua
SpiritOfSilvermoon: Here's another idea that might be worth considering, after reviewing `C_Intl` a little more closely.  The general summary is to put a cache in front of `C_Intl.Transliterate` in the hopes that it's sufficient to avoid most of the cost.

(This is more or less what I end up doing for the pattern-replacement mechanism I use for my addons to normalize accents etc..  Warning, code written in text editor only without running it, but hopefully this gets the concept across.)


```lua
local function Normalized(s)
    return C_Intl.Transliterate(s, "...desired transliterate args") -- remove accents (e.g. ICU documentation (though I'm far from an expert on ICU) suggests "NFD; [:Nonspacing Mark:] Remove; NFC."))
end

local SortStrings = {}

function GetSortString(s)
    if not s:match("[\128-\255].") then return s end -- No normalization needed for 7-bit-clean strings.

    local r = SortStrings[s]

    if r then return r end

    r = Normalized(s)
    SortStrings[s] = r
    return r
end
```

I was hoping to get rid of even this translation cache layer in my logic, but perhaps I'll have to keep it (hopefully `C_Intl` doesn't manage to be slower than a cascade of complex string pattern match substitutions, but we'll see...).

In my case, it wasn't sufficiently important to build any complex machinery to remove entries from the `SortString` cache (particularly with purely 7-bit strings sieved out up front).  But you'd have to see for your data sets as to whether that's a problem or not.  For my addons this sped up sorting to the point where it was fast enough to do all in one go without deferring work across frames etc..

Meorawr: We had a global cache for our old implementation. It's not a bad idea - it's got cheap implementation and works well for repeated queries, However, there's a few things we don't like about it.

Firstly, if the cache is a regular table without a special `__mode` then we're committing to paying a potentially large memory cost for the lifetime of the UI session as more and more strings get added into it. The table ultimately pins the strings and prevents their collection.

It's hard to predict up-front how large that table truly gets in an average session. In the worst case, suppose that for example the user has opened the icon browser (~roughly 30k strings for icon names), the music browser (~5k?), and has a decent sized directory (say 5k profiles) then we're looking at a table that's got 40k entries in it. Lua rounds this up to a 65,536 node hash table allocation.

We can potentially have substantially more than 40k distinct strings involved here, because both the keys and normalized results are references to strings. If, for example, a significant fraction of inputs produce distinct normalized outputs, we're retaining tens of thousands of additional GC objects as well.

What's undesirable about this scenario, aside from the table taking up a rather significant 3 MiB in its node array, is that the Lua GC during its mark phase has to synchronously walk all nodes inspecting their keys and values and following any references to GC objects.

The node scan being linear memory access itself isn't too bad; the problem is following references to that many strings. This is considerably less cache-friendly than a linear scan, and as the UI runs synchronously on the main thread of the client, this has risk of introducing client stalls.

It's hard to estimate the cost. The property that concerns me most though is that the cost ultimately has unbounded growth owing to the session-wide lifetime of the cache, and that the work to traverse is likely to occur at times distant to and unrelated from the operation that bulk populated the cache. Cache population would occur when the user interacts with the addon for search and sort functionality, but GC work can occur while the player is in combat.

If, instead, the cache is weak (`__mode = "kv"`) then we avoid the worst case scenario with respect to memory retention. Strings that aren't strongly referenced elsewhere can eventually be collected. However, we lose determinism.

Cache lifetime becomes tied to GC behaviour, which is hard to predict as the GC itself is driven by allocations elsewhere - be it from Blizzard's code, other addons, or our own actions. We can thus end up in a situation where a user performs an action that bulk-fills the cache, then spends a minute or so browsing results, and in that minute passive allocations elsewhere have advanced the collector such that strings were marked eligible and subsequently collected. Then, if the user refines their search query, we need to do all that work again.

A possible middle ground would then be a single global "strong" cache with an explicit eviction interface. However, the problem there is how one synchronizes such a resource with multiple consumers. The icon browser shouldn't have to negotiate with the music browser to know that it's safe to flush the cache.

Where caching is beneficial, the approach we're instead leaning toward is local strong caches with deterministic lifetimes, as used for example by `StringMatcher`. This allows us to associate a persistent cache with, for example, the icon browser. The browser can populate the cache with exactly the strings it normalizes, and can evict that cache fully when the browser is closed.

But it's worth noting that this does not, alone, negate the _cost_ of normalization; only that of repeated effort to do so. If the icon browser has 30k strings, that normalization process must occur at least one time. It remains undesirable for us to lock the client during that first normalization pass, and thus we still need mechanisms to defer work across multiple frames.

Elsewhere, persisting normalized filter or sort keys in saved variables may technically be an option, but it introduces problems such as needing a way to invalidate those strings if normalization rules change. And, still, it's not desirable to lock the client for a one-time-pass over that data on login.

At most, we might look into setting up another local cache within the registry to prevent duplicate re-normalization if we ultimately find that provides measurable benefit on average. We're not likely to swap back towards a global cache model.

Ghostopheles: so true king

Raenore:
I did an initial pass on this on the 12.1.5 PTR.

Notes:
- This looks good, the new coat of paint in the directory feels good overall. Future-wise we should try to give it a left-margin (consistent with all other tabs), but that's for whenever we actually handle this page anew.
- All the searches I tested were diacritic-agnostic now, which is good. Including Profiles, Pet Browser, Mount Browser, etc. A welcome change for special character handling.
- The directory looked fine in its sorting and listing, I saw no immediate issues.

I'm putting this as a normal review right now, as I've not conducted a 100% full test yet.
Just initial findings/musings, if you will.

Also this branch will require rebasing given the flattening quite literally flattened its merge situation.

totalRP3/Core/FunctionUtil.lua
SpiritOfSilvermoon: Thanks for entertaining my thoughts here.

In the case of my own addons, I also (as with the code above) just did a straight pass-through for purely 7-bit strings, which removed a lot of things out from the start-- only stuff that's actually going to have a (potentially) non-identity sort key gets a table entry or has to go through any further processing.  That was done on the theory that the normalization is always going to be an identity transformation for purely 7-bit strings anyway so there's no need to do anything special for them.

If the icon browser is, for example, all built-in filenames (though I realize you're likely just using this as one example), I suspect that case falls out entirely as I don't think I remember seeing much in the way of built-in filenames having more than relatively simple US-ASCII characters.  But, if you buy that the strings that are for things like file names etc. are all going to fall out and avoid taking table entries, then I think you really just have to be concerned with user data entries (character names etc.).  Those might well all be things that need real, transformed sort keys (hopefully not all but certainly some non-trivial fraction).

Not sure if that changes your calculus at all in terms of practical sizing.  Still may be tough to estimate the size of a user's directory or how many oddball strings with accents might arise in some complex TRP profiles, of course.

For what it's worth, my observations were, for a data set of even 17K-ish strings (user strings, some fraction of them being character names with accents etc.), that even rebuilding the table from scratch (i.e. the first time "cold hit" case) wasn't really super noticeable from a user impact perspective; the real performance drain was from all the backtracking references that sorting operations did which were re-re-re-re-(re)translating the same strings over and over which quickly added up to something much more substantial.  If you buy that argument (as well), then even marking the table as weak keyed might not be substantially impactful because there's probably not going to be a GC going on slicing in to the middle of your search operation (or at least if that's happening all the time, there's probably bigger fish to fry as far as performance issues afoot...), and once you've re-warmed the cache it's good enough for any sort operation to finish quickly-ish.

So I had (for reference, just as some back of the napkin numbers):

- About 17K strings (which were primarily a mix of RP realm character and guild names)
- Of these, 1524 had a non-identity sort key translation (so around 10% give or take).  All the rest just short-circuited sort key normalization and could safely use the input string (as they were pure "7-bit clean" strings).
- Building the translation cache "from scratch" for a cold start typically took around 0.0058sec ... 0.006sec or so on a few (not very scientific) runs on my machine.  That's however not using `C_Intl` but a cascade of (arguably clever but messy to maintain) string pattern matches to replace accents with non-accent base characters, and it's within the realm of possibility that `C_Intl` manages to be even slower than that code (I have as of yet no data on that).

In that case I probably could have just made the cache entirely local to the search function and wiped it afterwards, looking back on this data; I only used it when the user typed a string into a search filter type box which wasn't terribly frequent.

Now of course it's always possible to hit the worst case scenario where all of your 5K profiles had a bunch of strings with accents.  My data set was just names (guild and character), and on NA (not EU) RP realms, so it's possible EU realms with heavier concentrations of non-English language users or TRP profiles (in general) might have greater use of accents... but just a few data points to think about.


Defintely agree that it's desirable to avoid needing complex synchronization between different consumers of sort keys to need to agree on when keys are/aren't still referenced, etc. - wouldn't want to go down a road that imposes that sort of non-local complexity on all the consumers of the facility.

In either case, appreciate the technical discussion!

(Edit: (And to clarify- my "build from scratch" number was actually measuring not only the time to build the cache table but also to do all the sorting and building of data structures afterwards all from a first-run from-cold-start situation.  So that's definitely an overestimate as I just had quickly hacked that instrumentation back in to get some basic numbers.  But, your experiences may vary, of course.))

Raenore: For my own curiosity, you mentioned your own addons multiple times and how these changes sped things up.

Might I inquire which addons these are (names/github links/etc)? Just so we can have a look at the implementation and maybe find a thing or two from code that's already proven to work etc.

SpiritOfSilvermoon: Not that I have published anywhere unfortunately ... don't have anything on GH/CurseForge at the moment (we'll see if I ever do decide to publish things but it's mostly stuff for myself and a couple of friends presently, very (very) small userbase).  But the sort key normalization is a derivation of "artyfox"'s Aug 2022 algorithm which could be found here, should you be curious: https://stackoverflow.com/a/73283799 , of which my (relatively minor) contribution is mainly just the cache front-end behavior glued atop it and the early short-circuit if it's a pure 7-bit string.

... though what's outlined in that article requires a small bit of adaptation to make it work under WoW's Lua dialect, mostly to break the big normalization `gsub` pattern up into a cascade of multiple pattern matches, as WoW's Lua stdlib croaks complaining that the pattern complexity exceeds its implementation limits if you try that using the version if the article "as-is" (presumably nobody at the time thought it likely that anyone would be crazy or creative (take your pick...) enough to send such a large pattern through...).

So the core of what I did was basically what was described above, except instead of `C_Intl.Transliterate...`, a series of `stringvalue:gsub()` calls to go through two cascades of pattern match tables (two to not exceed what WoW's `stringvalue:gsub` can handle).  Again assuming that the quick 7-bit-only check says it's worth proceeding further, otherwise early escape exit.  The version in my addon is a bit messier than the above suggestion, as I didn't really encapsulate the cache table maintenance like I should have and dumped that into the (one) caller, and never went back to clean it up once it was performant enough to stop worrying about.

Hopefully, `C_Intl` is actually fast enough to drop the creative pattern match thing as while clever, it's definitely a bit of an "interesting" pattern & of course if any new accents show up, must be manually maintained.

SpiritOfSilvermoon:
> I did an initial pass on this on the 12.1.5 PTR.
> 
> Notes:
> 
> * This looks good, the new coat of paint in the directory feels good overall. Future-wise we should try to give it a left-margin (consistent with all other tabs), but that's for whenever we actually handle this page anew.
> * All the searches I tested were diacritic-agnostic now, which is good. Including Profiles, Pet Browser, Mount Browser, etc. A welcome change for special character handling.
> * The directory looked fine in its sorting and listing, I saw no immediate issues.
> 
> I'm putting this as a normal review right now, as I've not conducted a 100% full test yet. Just initial findings/musings, if you will.
> 
> Also this branch will require rebasing given the flattening quite literally flattened its merge situation.

Might I ask as to what procedure you use to try this on 12.1.5 PTR?

(This is probably a naive question.  But I don't see anything obvious as to how to force an addon to be loaded on early-stage PTR builds, and the current PTR build configuration appears to have the built-in addon support switched off (I might hazard a guess that that might be so that the developers don't get deluged with people complaining about broken addons when things are still heavily in flux).  I'm probably overlooking some documentation somewhere.)

I'd been considering something like a `/run` script to pop up an editbox frame to loadstring() something larger you paste in, but that's all sorts of unpleasant (not to mention things get even more unpleasant very quickly as soon as addons with XML frame declarations or multiple source files or SVs etc. come in), and if there's an easier way to do these things on earlier-stage PTR versions for addon developers, I'd be very curious!

Ghostopheles:
> the current PTR build configuration appears to have the built-in addon support switched off

Addons are not disabled on the 12.1.5 PTR. Make sure you have the "Load out-of-date addons" checkbox ticked.

SpiritOfSilvermoon:
> > the current PTR build configuration appears to have the built-in addon support switched off
> 
> Addons are not disabled on the 12.1.5 PTR. Make sure you have the "Load out-of-date addons" checkbox ticked.

Yep, you're right.

Turns out that this was my own silly fault: I failed to notice that this time around the flavor is `_xptr_` instead of `_ptr`_, so of course, empty `Interface\AddOns` directory (as I'm working out of the wrong directory...), and all the UI for addons gets turned off as the client sees that there aren't any there (which there really aren't).  Sigh.

Thanks for humoring my silly questions.  I'll see if I can poke around with this build on the PTR a bit using my production TRP SVs to see if anything blows up, so that I'm at least returning some useful time back in exchange for your time here :)

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -144,96 +151,96 @@ local sortingMap = {
...
 }
 
-local function switchSorting(key)
-	local pair = sortingMap[key];
-	if not pair then
-		return;
-	end
-	sortingType = (sortingType == pair.asc) and pair.desc or pair.asc;
-	refreshList();
+local function CompareRelationsAscending(a, b)
```
Raenore: Prior ascending relations sort used `>` and descending used `<` with 0 treated as highest, so the two directions were opposites.

This was rewritten in this branch so that it now uses `<`. Which means both directions now sort non-zero relations the same way. Making toggling only move the "None" (order 0) entries between first and last.

We should either flip `CompareRelationsAscending` back to `>`, or make it treat 0 as lowest so it stays opposite to the descending one.

totalRP3/UI/Browsers/IconBrowser.lua
Raenore, line 294:
```suggestion
	self.matcher:SetQuery("");
	self.searchCategories = {};
```

Copied from discord:
repro:
1. Open icon browser
2. type "fire"
3. choose icon and close browser.
4. re-open browser
5. choose a filter category randomly
6. notice it only shows things with "fire" in it, even though search is empty.

From what I read..
`ClearAllFilters()` sets `self.searchQuery = ""`
But it never calls `self.matcher:SetQuery("")`.

The followup` SearchBox:SetText("")` doesn't fix it, because `SetSearchQuery("")` sees the query is already `""` and skips the matcher.

So I guess it can be fixed by... adding `self.matcher:SetQuery("")` in `ClearAllFilters`

I can confirm background browser doesn't have it, given that moves through `SetSEarchQuery`.

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -728,7 +804,7 @@ local deleteCompanionProfile = TRP3_API.companions.register.deleteProfile;
 
 local function decorateCompanionLine(line, elementData)
 	decorateGenericLine(line);
-	local profileID = elementData[1];
+	local profileID = elementData.profileID;
 	local profile = getCompanionProfiles()[profileID];
 	line.id = profileID;
```
Raenore, line 809:
```suggestion
	line.id = profileID;
	
	if not profile then
		return;
	end
```

Repro from discord:
> Basically removing a pet and going back to the companions tab directory throws it.
> Given there's on nil guard in `decorateCompanionLine`

Image

totalRP3/Modules/Register/Main/RegisterList.lua
```diff
@@ -144,96 +151,96 @@ local sortingMap = {
...
 
-local function switchSorting(key)
-	local pair = sortingMap[key];
-	if not pair then
-		return;
-	end
-	sortingType = (sortingType == pair.asc) and pair.desc or pair.asc;
-	refreshList();
+local function CompareRelationsAscending(a, b)
+	return a.relationOrder < b.relationOrder;
```
Raenore, line 155:
```suggestion
	return a.relationOrder > b.relationOrder;
```

Prior ascending relations sort used `>` and descending used `<` with 0 treated as highest, so the two directions were opposites.

This was rewritten in this branch so that it now uses `<`. Which means both directions now sort non-zero relations the same way. Making toggling only move the "None" (order 0) entries between first and last.

We should either flip `CompareRelationsAscending` back to `>`, or make it treat 0 as lowest so it stays opposite to the descending one.

Raenore (changes requested):
Alright, done my bigger review as mentioned last time it wasn't 100% thorough yet.

Found 3 issues and noted them with improvements/suggestions where possible.

Aside from these, everything seems to be in order for now.

totalRP3/Modules/Register/Main/RegisterList.lua
Meorawr: Couldn't repro this myself but the idea is sound (do you need a shitload of companion profiles?).

PR #1361, opened 2026-09-08
Add AGENTS.md guidance

Meorawr:
Adds an AGENTS.md with relevant guidance for model-assisted development tasks.

We can't reasonably police the usage of such tooling when it comes to contributions, nor is there a technical reason to dismiss its usefulness. Most of our concerns toward the technology are along the lines of culture and bad practices surrounding their use.

Practically, we should prepare for these tools becoming more commonplace in development. The guidance here is a rough, largely untested skeleton intended to help model-assisted contributions follow our technical preferences and project expectations.

The presence of this file is not to be construed as anything more than that. We're not opening the gates to vibe-coded changes and filling the addon with slop du jour. Contributions remain subject to maintainer review, and we expect contributors remain responsible for understanding, validating, and explaining their changes.

Also tagging @Ghostopheles because I think he communes with the clank more than most of us.


AGENTS.md
```diff
@@ -0,0 +1,74 @@
...
+
+- For new code, prefer explicit construction with `TRP3_API.AllocateObject` and/or `TRP3_API.SetObjectPrototype`, followed by an explicit `object:__init(...)` call. This keeps initialization signatures visible to LuaLS. Existing `CreateObject` usage does not need to be changed solely for this preference.
+- Use mixins for shared behavior and state, including on objects created through the prototype APIs. Keep initialization dependencies explicit by calling the relevant `__init` methods.
+
+## UI Components
+
+- UI frames and regions cannot use the prototype object model. Use mixins for UI composition and lifecycle behavior.
+- Keep XML declarative. Do not add or expand inline Lua logic in XML script blocks; put behavior in a mixin instead.
+- When changing existing inline Lua logic, suggest moving it to a mixin and ask whether that refactor should be included.
+- In UI mixins, define `OnLoad` first, followed by other script handlers such as `OnShow` and `OnHide`, then the remaining methods.
```
Ghostopheles, line 77: > Contributions remain subject to maintainer review, and we expect contributors remain responsible for understanding, validating, and explaining their changes.

I'd recommend adding a section at the end to emphasize this point.

```suggestion
- In UI mixins, define `OnLoad` first, followed by other script handlers such as `OnShow` and `OnHide`, then the remaining methods.

## Reporting and Responsibility

- An agent cannot run the game or load the addon. Passing validation does not necessarily mean that the change works in-game. Do not describe a change as tested or verified on that basis.
- Treat any change touching UI frames, script handlers, event registration, or runtime API access as unverified regardless of what static checks report.
- Conclude every change with a short report covering:
  - what changed and why;
  - which checks were run, and their results;
  - what requires in-client testing;
  - any assumption not confirmed by reading the relevant code, including which clients the change was and was not reasoned about.
- State uncertainty plainly rather than silently resolving it. If a requirement is ambiguous, ask, or implement one reading and say which. 
- The contributor, not the agent, is responsible for understanding, validating, and explaining the change. The report is material for that review, not a sign-off that replaces it.
```

Raenore (approved):
I think, regardless of opinions regarding the tech, this is a good thing to add.

While it might not be our thing, we can't guarantee nor know if someone had/has used agentic or some form of AI when writing unless they decide to tell us.

The issue is that, unlike text writing, code generally has a few "ways" to write something and it's nearly impossible to know if someone has or has not used tooling for it. The only 'true' tell at times is some conventions/standards being applied wrong or the comments (they're generally a dead giveaway).

We can't reasonably figure this out for everything, as a normal author could've come to the code result without having used any tooling whatsoever. It's then just a matter of "does the author understand the code".

So this set of instructions basically mandates that if someone does use tooling, that it's at least not some hallucinating BS (or less) and somewhat 'guardrailed' to work the least against our project.

It's the best middle ground we can achieve in the current environment, where we have to use a "code of honor" at best, but understandably not everyone will follow it.

PR #1362, opened 2026-09-09
Add realm time and weather conditions to automation

Meorawr:
This adds two new automation conditions based off that one discussion we had weeks back about some other addon doing exactly this. These feel the most RP-centric;

- `[time:...]` activates based on in-game (realm) time. Arguments can either be tokens like `day` or `night`, or a range of hours like `18-23`. Multiple tokens or ranges can be specified via `/` delimiters. Example: `[time:day,21-3]` would activate between 06:00 and 17:59, and 21:00 and 02:59. Input out of the range 0-23 will wrap around, so both `3-0` and `3-24` mean "from 03:00 to 23:59".
- `[weather:...]` activates based on, uhh, weather. This accepts tokens `clear`, `rain`, `snow`, and `sandstorm` - matching the options available to transmog situations. As with time, multiple options can be specified via `/` delimitation - so `[weather:rain/snow]` will activate when it's raining or snowing.


totalRP3/Modules/Automation/Automation_Conditions.lua
```diff
@@ -241,3 +241,126 @@ TRP3_AutomationUtil.RegisterCondition({
...
+	day = TimeRanges.Day,
+	night = TimeRanges.Night,
+	morning = TimeRanges.Morning,
+	afternoon = TimeRanges.Afternoon,
+	evening = TimeRanges.Evening,
+	[C_Intl.FoldCase(L.AUTOMATION_TIME_DAY)] = TimeRanges.Day,
+	[C_Intl.FoldCase(L.AUTOMATION_TIME_NIGHT)] = TimeRanges.Night,
+	[C_Intl.FoldCase(L.AUTOMATION_TIME_MORNING)] = TimeRanges.Morning,
+	[C_Intl.FoldCase(L.AUTOMATION_TIME_AFTERNOON)] = TimeRanges.Afternoon,
+	[C_Intl.FoldCase(L.AUTOMATION_TIME_EVENING)] = TimeRanges.Evening,
```
Meorawr, line 263: We allow use of English or localized tokens, consistent with other automation conditions like class.

totalRP3/Modules/Automation/Automation_Conditions.lua
```diff
@@ -241,3 +241,126 @@ TRP3_AutomationUtil.RegisterCondition({
...
+		return false;
+	end,
+});
+
+local WeatherTypes = {
+	Clear = { type = Enum.WeatherType.Clear, threshold = 0.25 },
+	Rain = { type = Enum.WeatherType.Rain, threshold = 0.25 },
+	Snow = { type = Enum.WeatherType.Snow, threshold = 0.25 },
+	Sandstorm = { type = Enum.WeatherType.Sandstorm, threshold = 0.25 },
+};
```
Meorawr, line 330: Weather has minimum threshold bounds for intensity. Transmog situations are the same, but unfortunately the exact thresholds are opaque. Will need to tune these a bit, but I'm assuming ~0.25 for all (though maybe we should make Clear use no threshold.)

totalRP3/Modules/Automation/AutomationConditions.lua
```diff
@@ -241,3 +241,126 @@ TRP3_AutomationUtil.RegisterCondition({
...
+		return false;
+	end,
+});
+
+local WeatherTypes = {
+	Clear = { type = Enum.WeatherType.Clear, threshold = 0.25 },
+	Rain = { type = Enum.WeatherType.Rain, threshold = 0.25 },
+	Snow = { type = Enum.WeatherType.Snow, threshold = 0.25 },
+	Sandstorm = { type = Enum.WeatherType.Sandstorm, threshold = 0.25 },
+};
```
Raenore: I've tested out `weather` and come to the conclusion that the threshold for Clear should be set to `0`.

Given it seems `C_Weather.GetCurrentWeather()` seems to (always?) note Clear as `intensity=0`.
Which means that all my tests in Stormwind, where it had:
```lua
Dump: value=C_Weather.GetCurrentWeather()
[1]={
  intensity=0,
  type=0
}
```

Made the following consistently fail (reported as `two`):
```
[weather:clear] one; two
```

Raenore (changes requested):
This looks promising.

Tested both.

**`time` works, but I think it should be tweaked.**

Using for example:
```
[time:6-20] one; two
```
When it's `20:51` will make it use `one` (true). Because `-20` seems to go until `20:59`.
I'm pretty sure users would expect it to be the upper bound, so until `19:59` or perhaps `20:00` at best.

**`weather` works, but has an issue with `Clear`.**
I've noted that in a comment to where you were pointing this out already.

Raenore (approved):
Further testing on 12.1.5 PTR.

### Weather

```lua
[weather:clear] true; false
```
Now properly shows `true` in Stormwind when it's.. Clear.


### Time
PTR realm time is 17:42 right now.

```lua
[time:17-17] true; false
```
Goes `true`, which makes sense. As we cover the entire hour I think I read?

```lua
[time:17-18] true; false
```
Is also `true`, perfect.

```lua
[time:16-17] true; false
```
Is `false`, also correct.

The named `day`, `night`, etc. still work too.

Good job Meow.

PR #1363, opened 2026-09-09
Restructure CompanionUtil implementation

Meorawr:
This is a small code cleanup for our CompanionUtil abstraction.

There's no functional change to the API here. All that's been done is relocating the implementations to dedicated data provider tables, and providing one extension point (`CompanionUtil.GetDataProvider()`) which is overridden to select an appropriate backing data source for companion data. This means we're only overriding _one_ function, rather than about 12. I also got rid of the ugly `Vanilla/` directory as we no longer really need it with this setup.

### DIFF
--- a/totalRP3/Core/CompanionProviderJournal.lua
+++ b/totalRP3/Core/CompanionProviderJournal.lua
@@ -1,0 +1,104 @@
+-- Copyright The Total RP 3 Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+local cachedSummonedMountID = nil;
+
+local CompanionProviderJournal = {};
+
+function CompanionProviderJournal:CanRenameCompanionPets()
+	return true;
+end
+
+function CompanionProviderJournal:GetCompanionPetIDs()
+	return C_PetJournal.GetOwnedPetIDs();
+end
+
+function CompanionProviderJournal:GetCompanionPetInfo(petID)
+	local speciesID, customName, _, _, _, _, _, speciesName, icon, _, _, _, description = C_PetJournal.GetPetInfoByPetID(petID);
+
+	if not speciesID then
+		return;
+	end
+
+	return {
+		id = petID,
+		name = (customName ~= "") and customName or speciesName,
+		customName = customName ~= "" and customName or nil,
+		icon = icon,
+		description = description,
+		speciesName = speciesName,
+		isCollected = true,
+	};
+end
+
+function CompanionProviderJournal:IsCompanionPetUnit(unitToken)
+	return UnitIsBattlePetCompanion(unitToken);
+end
+
+function CompanionProviderJournal:GetCompanionPetUnitName(unitToken)
+	return UnitNameUnmodified(unitToken);
+end
+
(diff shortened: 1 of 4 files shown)


totalRP3/Core/CompanionProviderJournal.lua
```diff
@@ -0,0 +1,104 @@
+-- Copyright The Total RP 3 Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+local cachedSummonedMountID = nil;
+
+local CompanionProviderJournal = {};
```
Meorawr, line 6: This is just the old implementation of CompanionUtil.lua moved here and stuck on a table of a different name. There's no changes otherwise.

totalRP3/Core/CompanionProviderStatic.lua
```diff
@@ -38,15 +28,17 @@ local function SetSummonedCompanionPetID(petID)
 	TRP3_Companions.summonedPetID = petID;
 end
 
-function TRP3_CompanionUtil.CanRenameCompanionPets()
+local CompanionProviderStatic = {};
+
+function CompanionProviderStatic:CanRenameCompanionPets()
```
Meorawr, line 33: This file better reflects the transformation that was done in CompanionProviderJournal.lua.

totalRP3/Core/CompanionUtil.lua
```diff
@@ -1,47 +1,34 @@
 -- Copyright The Total RP 3 Authors
 -- SPDX-License-Identifier: Apache-2.0
 
-local cachedSummonedMountID = nil;
-
 TRP3_CompanionUtil = {};
 
+function TRP3_CompanionUtil.GetDataProvider()
+	return nil;
+end
```
Meorawr, line 8: Strictly speaking, we don't support a nil data provider - so it's expected that the other calls would error if we somehow found ourselves loaded in a client that didn't override this function. We could maybe make this an assert to make the error clearer.

totalRP3/Core/Core.xml
```diff
@@ -26,7 +26,8 @@
 	<Include file="RegionUtil.lua"/>
 	<Include file="MarkupUtil.lua"/>
 	<Include file="CompanionUtil.lua"/>
-	<Include file="Vanilla\CompanionUtilOverrides.lua"/>
+	<Include file="CompanionProviderJournal.lua [AllowLoadGameType mainline, mists, wrath]"/>
+	<Include file="CompanionProviderStatic.lua [AllowLoadGameType vanilla, tbc]"/>
```
Meorawr, line 30: Pretending cataclysm doesn't exist. Wish I could have done that years back.

totalRP3/Core/CompanionUtil.lua
Solanya: or just a comment to remember down the line

Solanya (approved):

Raenore (approved):
Did my checks on this.

Seems like it is indeed mostly ported over code, with some `self:` calls added for the Methods.

As you noted, only 'thing' is that the base `GetDataProvider()` returns nil but I don't think we have a client that would load neither so it's fine.

PR #1364, opened 2026-09-09
Clean up deprecated API usage and stale luacheck entries

Meorawr:
Self-explanatory; the deprecations here aren't part of the removed set in 12.1.5, but they'll be cleaned up eventually - so migrating them to the newer API now.

We also had entries in luacheckrc for functions and constants we're not using anywhere, and are themselves deprecated.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -217,6 +217,12 @@ stds.wow = {
 			},
 		},
 
+		C_AutoComplete = {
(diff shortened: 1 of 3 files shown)


Raenore (approved):
Had seen these before when looking at the API changes and what might be deprecated in the future.

Good job Meowster.

PR #1365, opened 2026-09-09
Remove deferred localized text infrastructure

Meorawr:
Removes the `TRP3_LocalizeTextOnAddOnLoaded` function that was being used in a few places to lazily apply translated strings.

This was required back in the dark ages before we put `## LoadSavedVariablesFirst: 1` in the TOC, as we couldn't guarantee that our Locale setting would be respected uniformly if a reference to a translated string were evaluated before saved variables loaded. This is no longer an issue, so our `text="..."` attributes can directly reference translated strings using the `TRP3_L_` prefixing that's handled by Locale.lua when it loads.

Maybe in 2027 the `text` attribute will learn to support nested lookups. That'd be nice, wouldn't it?


Raenore (approved):
Tested on 12.1 with zero errors in my (Bug)sack.

This looks all "oui" to me.

Image

PR #1367, opened 2026-09-19
Add regional unique name support

Meorawr:
What a long, strange, disaster this is.

Forever is implementing a concept of a regionally unique name and making the concept of realms something that's not user-facing. The main part of this concept is that we can no longer address players via addon comms or APIs as a simple "Name" or "Name-Realm", it needs to be "Full Name".

Now, you might think that the APIs in-game behave sensibly. They don't. Allow me to elucidate.

```lua
-- Targeting an NPC:
UnitName("target") =>  "Stormwind Guard", nil
UnitFullName("target") =>  "Stormwind Guard", nil

-- Targeting yourself:
UnitName("target") => "Your Name", nil
UnitFullName("target") => "Your Name", "Classic Beta PvE"

-- Targeting another player:
UnitName("target") => "Their", "Name"
UnitFullName("target") =>"Their", "Name"
```

Yes, despite the game clearly being capable of supporting names with spaces in them via UnitName for 20+ years in the context of NPCs, the API decides "fuck it, let's randomly splay the name across both args and sometimes include a realm name".

Now, in terms of our code this isn't actually too bad - the problem we have is a mountain of technical debt and a decade of assumptions on how names are structured. So, I've done what I do best - invent layers of indirection - and exposed a new `TRP3_NameUtil` namespace with APIs that encompass the simple act of querying names from various sources (units or guids), parsing them (user input), and composing/decomposing them.

Name acquisition falls into four main functions;

- `TRP3_NameUtil.GetDisplayName(unitToken)`
- `TRP3_NameUtil.GetUnmodifiedName(unitToken)`
- `TRP3_NameUtil.GetQualifiedName(unitToken)`
- `TRP3_NameUtil.GetQualifiedNameByGUID(unitGUID)`

Both Display and Unmodified names return what the native API BLOODY WELL SHOULD BE RETURNING; calling them will consistently give us "Your Name" or "Their Name" or "NPC Name" on Forever, and (hopefully) just the regular single-word character name on every other client.

Qualified names are what we use for storage, API calls, and addon comms. The returns there are either "Full Name" for Forever, or "Name-Realm" for other clients.

Display and Unmodified names can be decomposed (`TRP3_NameUtil.DecomposeFullName(fullName)`) or have the "given" (ie. first) and "family" (ie. surname) names extracted. This works independently of the actual client naming mode - so `DecomposeFullName("Bob Dylan") => "Bob", "Dylan"` on every client.

Qualified names can also be decomposed. Their decomposed form is different - it's `name, realm`. The semantics here are tied to client naming mode; when using regionally unique names the `name` is the full name (ie. "Some Player") and `realm` is always nil. On all other clients, it'd be the regular "Name", "NormalizedRealm" form.

With that in mind, most of the changes are just swapping to these utilities and removing hardcoded assumptions that realms need to be appended to names, that `-` is a separator, etc.

I've also put in test cases, because the core naming utilities are standalone and easily verified - and, they really need to be, since this is a rather fundamental and risky change.


.github/workflows/pull_request.yml
```diff
@@ -44,6 +57,9 @@ jobs:
           echo "XML validation: ${{ steps.validate_xml.outcome }}"
           exit 1
 
+      - name: Run Lua tests
+        run: just test
```
Meorawr, line 61: I'll clarify up-front; writing tests is optional. I'd suggest that our low level core utilities would benefit from it, but I'm not expecting (nor intending or even considering) writing tests that encompass our UI. 

I'd rather have some tests in for super fundamental stuff _that can run automatically in CI_ than the situation our old test suite had where you had to manually log in and use a separate addon to run them.

totalRP3/Core/CommunicationProtocolBroadcast.lua
```diff
@@ -239,15 +238,15 @@ end
 local function onChannelJoin(_, _, arg2, _, _, _, _, _, _, arg9)
 	if not canaccessvalue(arg2) then return; end
 	if config_UseBroadcast() and arg2 and arg9 == config_BroadcastChannel() then
-		local unitName = unitIDToInfo(arg2);
+		local unitName = TRP3_NameUtil.DecomposeQualifiedName(arg2);
```
Meorawr, line 241: All `unitIDToInfo` calls now call this new DecomposeQualifiedName function which I'm hoping to god is 1:1 enough.

I think there's a few differences on edge cases ("Unknown" or "" names in particular may now return nil) - so, I expect some Lua errors as we discover stuff that previously silently worked.

totalRP3/Core/Configuration.lua
```diff
@@ -96,11 +96,27 @@ Config.resetValue = resetValue;
...
+			local element = elements[index];
+
+			if not element.showPredicate or element.showPredicate() then
+				return index, element;
+			end
+		end
+	end
+
+	return GetNextElement, elements, 0;
+end
```
Meorawr, line 113: Individual settings can now have a `showPredicate` attached. If it returns false, we never put it in the config UI. Beats the old stupid conditional table.insert system.

This is used for tooltip's Show Realm setting.

totalRP3/Core/Globals.lua
```diff
@@ -45,7 +46,9 @@ TRP3_API.globals = {
 	--@end-non-debug@]===]
 
 
-	player = UnitNameUnmodified("player"),
+	player = PLAYER_NAME,
+	player_given_name = TRP3_NameUtil.ExtractGivenName(PLAYER_NAME),
+	player_family_name = TRP3_NameUtil.ExtractFamilyName(PLAYER_NAME),
```
Meorawr, line 51: On Forever:
- `player` = "Full Name"
- `player_given_name` = "Full"
- `player_family_name` = "Name"

On other clients:
- `player` = "Name"
- `player_given_name` = "Name"
- `player_family_name` = nil

totalRP3/Core/Globals.lua
```diff
@@ -94,10 +97,8 @@ setmetatable(TRP3_API.globals.empty, emptyMeta);
 --*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
 
 TRP3_API.globals.build = function()
-	local fullName = UnitNameUnmodified("player");
-	local realm = GetRealmName():gsub("[%s%-%.]*", "");
-	TRP3_API.globals.player_realm_id = realm;
-	TRP3_API.globals.player_id = fullName .. "-" .. realm;
+	TRP3_API.globals.player_realm_id = TRP3_NameUtil.NormalizeRealmName(GetRealmName());
+	TRP3_API.globals.player_id = TRP3_NameUtil.GetQualifiedName("player");
```
Meorawr, line 101: On Forever, `player_id` would be "Full Name". On other clients, "Character-Realm".

totalRP3/Core/NameUtil.lua
```diff
@@ -0,0 +1,284 @@
...
+local function UnitOrPlayerNameFromGUID(unitGUID)
+	local name, qualifier = UnitNameFromGUID(unitGUID);
+
+	if not name or name == "" then
+		local _className, _classFile, _raceName, _raceFile, _sex, _level;
+		_className, _classFile, _raceName, _raceFile, _sex, name, qualifier, _level = GetPlayerInfoByGUID(unitGUID);
+	end
+
+	return name, qualifier;
+end
```
Meorawr, line 43: Mostly a partial workaround for the fact that GetPlayerInfoByGUID doesn't work correctly in Forever. The leading part should be removed when that's fixed.

totalRP3/Core/NameUtil.lua
```diff
@@ -0,0 +1,284 @@
...
+
+	if givenName then
+		local plain = true;
+		local familyNameOffset = string.find(givenName, SURNAME_SEPARATOR, 1, plain);
+
+		if familyNameOffset then
+			familyName = string.sub(givenName, familyNameOffset + 1);
+			givenName = string.sub(givenName, 1, familyNameOffset - 1);
+		end
+	end
```
Meorawr, line 101: This is to handle the stupid case of Forever APIs returning three different forms of values for names. If a given name is actually a full name, split it up and use that instead.

totalRP3/Core/UITools.lua
```diff
@@ -585,16 +586,11 @@ local function getCompanionOwner(unitType, targetType)
 			ownerName = ScanningTooltip:FindMatchingLine(COMPANION_PET_OWNER_PATTERNS);
 		end
 
+		ownerCharacterID = TRP3_NameUtil.ComposeQualifiedName(ownerName);
```
Meorawr, line 589: May need to test cross-realm owner detection of pets in Classic here... if we can be bothered.

totalRP3/Core/UITools.lua
```diff
@@ -620,9 +616,6 @@ function TRP3_API.ui.misc.getCompanionFullID(unitToken, unitType)
 	if shortID then
 		local owner = getCompanionOwner(unitToken, unitType);
 		if owner ~= nil then
-			if not owner:find("-") then
-				owner = owner .. "-" .. globals.player_realm_id;
-			end
```
Meorawr, line 625: `getCompanionOwner` returns a normalized result that includes the realm if needed, so this isn't required.

totalRP3/Modules/ChatFrame/ChatFrame.lua
```diff
@@ -666,26 +664,18 @@ function Utils.customGetColoredName(event, _, _, unitID, _, _, _, _, _, _, _, _,
...
+	characterID = TRP3_NameUtil.GetQualifiedNameByGUID(GUID);
 	---@type Player
-	local player = AddOn_TotalRP3.Player.static.CreateFromNameAndRealm(character, realm);
+	local player = AddOn_TotalRP3.Player.static.CreateFromCharacterID(characterID);
 
 	-- Character name is without the server name is they are from the same realm or if the option to remove realm info is enabled
-	if realm == Globals.player_realm_id or getConfigValue(CONFIG_REMOVE_REALM) then
-		characterName = character;
-	end
+	local characterName = Ambiguate(characterID, getConfigValue(CONFIG_REMOVE_REALM) and "short" or "none");
```
Meorawr, line 673: Ambiguate retains realm semantics even in Forever, so an input of "Full Name" applies no change irrespective of the context parameter. In other clients, this should strip "-Realm" suffixes equivalent to the player's own realm, or all of them if the remove realm setting is configured.

totalRP3/Modules/ChatFrame/Prat.lua
Meorawr, line 1: Prat changes untested.

totalRP3/Modules/Register/Characters/RegisterCharacteristics.lua
```diff
@@ -156,12 +157,13 @@ local function getCompleteName(characteristicsTab, name, hideTitle)
...
+	if not hideTitle and characteristicsTab.TI and characteristicsTab.TI ~= "" then
 		text = strconcat(characteristicsTab.TI, " ");
 	end
-	text = strconcat(text, characteristicsTab.FN or name);
-	if characteristicsTab.LN then
-		text = strconcat(text, " ", characteristicsTab.LN);
+	if characteristicsTab.FN and characteristicsTab.FN ~= "" then
+		text = strconcat(text, TRP3_NameUtil.ComposeFullName(characteristicsTab.FN, characteristicsTab.LN));
+	else
+		text = strconcat(text, name);
```
Meorawr, line 166: It struck me as a bit odd that we append LN even if there's no FN. I think that'd be subtly incorrect in the new world ("Player Name ExtraSurname"), so being stricter here and only using LN if there's an FN.

totalRP3/Modules/Register/Characters/RegisterMain.lua
```diff
@@ -334,7 +334,6 @@ end
 
 --- Raises error if KNOWN unitID
 function TRP3_API.register.addCharacter(unitID)
-	assert(unitID and unitID:find('-'), "Malformed unitID");
```
Meorawr, line 337: There's not really a fixed format for character IDs now, given they can be  "Name-Realm" or "Full Name".

totalRP3/Modules/UnitPopups/UnitPopups.lua
```diff
@@ -231,12 +246,12 @@ local function CreateOpenCharacterProfileButton(menuDescription, contextData)
 	local function OnClick(contextData)  -- luacheck: no redefined
 		local unit = contextData.unit;
 		local name = contextData.name;
-		local server = contextData.server;
-		local fullName = string.join("-", name or UNKNOWNOBJECT, server or GetNormalizedRealmName());
+		local server = contextData.server or contextData.surname;
```
Meorawr, line 249: They decided to rename this field in Forever. In all client code. So, `surname =  "RealmName"` on all but one client. 👏

.luacheckrc
```diff
@@ -14,6 +14,11 @@ ignore = {
 
 	-- Ignore unused self. This would popup for Mixins and Objects
 	"212/self",
+
+	-- Unused variables beginning with "_" are ignored.
+	"211/^_",
+	"212/^_",
+	"231/^_",
```
Meorawr, line 21: It was getting a bit sad seeing lots of unnamed "_"  and having to count them rather than name them.

totalRP3/Core/NameUtil.lua
```diff
@@ -0,0 +1,284 @@
...
+---@return string? qualifiedName Qualified character name for the current naming mode.
+function TRP3_NameUtil.GetQualifiedNameFromString(qualifiedName)
+	if not qualifiedName or qualifiedName == "" then
+		return nil;
+	end
+
+	return Impl.GetQualifiedNameFromString(qualifiedName);
+end
+
+--- Compose a qualified character name from a name and qualifier.
```
Raenore, line 228:
```suggestion

--- Normalize a name and qualifier for the current naming mode.
---@param name string? Name, e.g. "Zugzug".
---@param qualifier string? Realm or family name, e.g. "AzjolNerub".
---@return string? name Name, e.g. "Zugzug".
---@return string? qualifier Realm, e.g. "AzjolNerub".
function TRP3_NameUtil.NormalizeUnitName(name, qualifier)
	return Impl.NormalizeUnitName(name, qualifier);
end

--- Compose a qualified character name from a name and qualifier.
```
Callers outside `NameUtil` had no way to join a surname onto a name for the regional mode.
As `ComposeQualifiedName` ignores its second argument there.

totalRP3/Modules/UnitPopups/UnitPopups.lua
```diff
@@ -60,47 +60,60 @@ local function GetUnitCompanionProfileInfo(unitToken)
...
-local function ShouldShowOpenBattleNetProfile(contextData)
-	if not ShouldShowOpenProfile(contextData) then
-		return false;
-	end
+TRP3_UnitPopupUtil = {};
+
+function TRP3_UnitPopupUtil.GetCharacterIDFromName(contextData)
+	local name = contextData.name;
+	local qualifier = contextData.server or contextData.surname;
+	return TRP3_NameUtil.ComposeQualifiedName(name or UNKNOWNOBJECT, qualifier);
```
Raenore, line 86:
```suggestion
	local name, qualifier = TRP3_NameUtil.NormalizeUnitName(contextData.name or UNKNOWNOBJECT, contextData.server or contextData.surname);
	return TRP3_NameUtil.ComposeQualifiedName(name, qualifier);
```
`GetCharacterIDFromName` now runs the name and surname through `NormalizeUnitName` before composing the ID.

It used to build (for example) `"Heals"` instead of `"Heals Found"`, which causes a "`Unknown character ID"` assert.

Raenore (commented):
Adding two suggestions that should handle, hopefully https://github.com/Total-RP/Total-RP-3-Extended/pull/317.

This is untested, as Forever Beta was restarting but it should be quoted "morally correct".

Raenore:
Can confirm no more TRP (Extended) errors with those suggestions.
Of course if you want them handled differently then by all means.

totalRP3/Core/NameUtil.lua
SpiritOfSilvermoon: Yes, the situation with co-opting the realm name (sometimes, but not always) for UnitName* is a mess.  Not clear why they did that but it's certainly going to make things... interesting if they ever decided to bring forward that support to standard/mainline.

As far as I could tell (a few days ago anyway, unless they've changed it tonight), the qualifier returned by GetPlayerInfoByGUID is the internal realm name, not the surname (unclear to me whether or not this is what you meant by "doesn't work correctly").  In either case, unless Blizzard has indicated an intent to change that, it doesn't seem like using the qualifier (what used to be realm name) from GetPlayerInfoByGUID is advisable.  I have no data on whether Blizzard intends to change that or not, to be clear.

By deduction based upon externally observable behavior, so far, it seemed to me like UnitNameFromGUID and GetPlayerInfoByGUID probably used the same internal machiney under the hood, so I haven't yet encountered a situation where such a fallback provides useful value.

One edge case that you may wish to be aware of is that sometimes, presumably until enough information has been requested from the network, UnitNameFromGUID won't return a surname and will return UNKNOWNOBJECT for the primary name (unlike GetPlayerInfoByGUID).  Making matters worse, there are now (obviously this doesn't apply to Forever) toys available in MN S2 that will cause UNKNOWNOBJECT to be a valid character name for UnitName to return even in the case where the name information *is* available (though not UnitNameUnmodified, fortunately, so this code needn't be confused by that conundrum).  In the case where UNKNOWNOBJECT is returned by UnitNameUnmodified, I've yet to see a case where GetPlayerInfoByGUID returned more data, and it's probably best for simplicity to treat this as "name not available" because as far as I can tell you can't really usefully address a player over the network with only the primary name, anyway, in Forever.

totalRP3/Modules/UnitPopups/UnitPopups.lua
Meorawr: I ended up just making ComposeQualifiedName work consistently with `"Name", "Realm"`, `"Full Name", nil`, and `"First", "Last"` inputs which seems to solve this a bit better.

PR #1371, opened 2026-09-19
Create Profile Enhancements - Enter in name to "Create" + Add flow to default profile create

Raenore:
Enhanced the create profile flow a bit more.

1. When on the name field (which is default filled & focused), you can just press enter to create the profile (if the button is enabled). Makes creating a new profile quicker.
2. The "You don't have a profile, create one to edit info" flow now uses the new profile creation flow as well.

Video below showcases both improvements:

https://github.com/user-attachments/assets/68025051-ae60-4d81-8ff9-f3fee8c2ee46

Optional improvements, but up to the two bozos:

### Make enter to "Create Profile" work regardless if the Name field is focused or not.
I wanted to hold off on that for now as it means we'll eat an enter key always when that frame is open and not sure if we want that.

### Make creating a new profile always switch the user to the new profile. - IMPLEMENTED!
The create one to edit flow always changes you to the new profile, because you started from default.

If wanted, we can extend this behavior to all profile creation so that when a new profile is created (or imported), we always switch to the newly created one.

I'm mixed on this, as it'll be nice QoL and create/import profile is a choice with intent (you don't do that if you don't want to use the new profile for example).. But also, it'll be different from before, so still a behavior change.


Raenore:
> Make creating a new profile always switch the user to the new profile

This is now implemented.

Meorawr (approved):

Solanya (approved):

PR #1372, opened 2026-09-19
Test the package label for conditional packaging

Raenore:
You'll see this run a few times, perchance.

First time I'm doing it without the label to see ~~if it successfully skips~~ it does.

### Run without `package` label
It skips the packager steps as required.

### Run with `package` label
It successfully triggers the packager steps as required.


Raenore:
Ready for review, it seems to work and I've updated the main PR post with the findings.

I'm assuming direct control and re-adding the PR lines to pull `.github\workflows\pull_request.yml` as otherwise this won't really do anything.

Let me know if we require these steps or conditionals in other places, as I'll be adding it to our other projects as well (including the label).

.github/workflows/pull_request.yml
```diff
@@ -44,6 +49,33 @@ jobs:
...
+        run: |
+          if gh api "repos/$GITHUB_REPOSITORY/issues/$PR/labels" --paginate --jq '.[].name' | grep -qx package; then
+            echo "package=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create Package
+        if: steps.labels.outputs.package == 'true'
+        uses: BigWigsMods/packager@v2
+        with:
+          args: -dz
```
Meorawr, line 68: nuts

Meorawr (approved):

PR #1373, opened 2026-09-19
Update client feature tests for Forever

Meorawr:
Quick pass over our project ID tests for things that were incorrect in Forever.

- Transmog is a thing in Forever. Our expansion test doesn't work - however, mercifully, there's a game rule we can use to test if transmog is enabled. This returns false on Classic Era, and true on Forever (and, presumably, Mainline).
- War Mode _could_ be updated to use the dedicated feature test API, _if_ it worked on Forever (it returns true, despite War Mode not existing.)
- The other client features for our Water Elemental workaround, tooltip APIs, and channel broadcasts continue to elude better checks - so I just expanded the project ID comparisons.
- Our tutorial system actually seems to work fine in Classic Era. I assume this changed at some point, so it no longer needs gating.
- Our project ID switch on the Residence open-in-map functionality can just feature check on a modern API (`C_Map.OpenWorldMap`) instead.
  - Also adding a combat lockdown test; we can't show UI panels in combat, so we shouldn't be doing anything.
- The stabled pet browser has more than art dependencies at this point, as it relies on modern C_StableInfo APIs that aren't present in Classic. Gating on those instead of art assets.


Solanya (approved):

Raenore (approved):
1. World map stuff seems fine.
2. Tutorials are neat and seem to work.
3. Pet browser change is good too.
4. War mode well.. We'll see how rulesets react to our changes.

Honestly all of this just sounds like "Forever did an oopsie with certain checks" and we're trying to put out that fire.

